PR #70, opened 2020-11-30
Rails Components: Attribute Schema

AndrwM:
## ☃️ 

Validate Rails Component attributes like we do for React Props

#### Description
I’d like to make this its own Sage bump, if the crew doesn’t mind, because this touches every Rails component. The intent is to precision-add only the argument checking to each SageRails component without updating the attributes themselves to prevent any diff in Kajabi-Products. I defaulted to setting each as an optional attribute unless I was familiar with the Products usage and audited that component on the other side (in Kajabi-Products). This leaves some space for improvement to make these checks even more strict down the road.
One issue that this PR’s diff does highlight is an inconsistent use of symbols and strings for arguments with a defined set of values. (e.g. gap may only be one of [:xs, :sm, :md, :lg]). I think it would be wise for us choose one approach for predictability; however, that falls outside the scope of this PR.

#### This PR currently includes:
- Method (`set_attribute_schema({})`) to build a schema and `attr_accessor`s
- `SageSchemaHelper` module to store reusable schema values (a temporary solution until we are able to use a design tokens file)
- Usage of [classy_hash](https://github.com/deseretbook/classy_hash) which will give us powerful tools to deeply check the hash passed to `sage_component`.
  - Check whether an attr value is _one of_ many different explicit values or types
  - Ability to define custom procs for validation
  - Sets us up to refactor the React component `configs.js` pattern allowed props pattern into a global implementation that these Rails components can also consume.
  - See the [classy_hash docs for examples ](https://github.com/deseretbook/classy_hash)

------------

#### Documentation:
image

#### Preview of error message:
image

### DIFF
--- a/docs/app/views/examples/shared/_code.html.erb
+++ b/docs/app/views/examples/shared/_code.html.erb
@@ -1,19 +1,38 @@
-<%= sage_component SagePanelHeader, {} do %>
-  <h2 id="<%= title %>-example-code">SageRails Code</h2>
-<% end %>
-<div class="example__code">
-  <button class="example__expand-btn" aria-expanded="false" aria-controls="example-rails-code-<%= title %>">Expand this code snippet</button>
-  <div id="example-rails-code-<%= title %>" class="sage-code-snippet" tabindex="-1">
-    <pre class="prettyprint"><code><%= File.read(File.join(Rails.root, "app/views/examples/", type.pluralize, title, "_preview.html.erb")) %></code></pre>
-  </div>
-</div>
+<% component_classname = "sage_#{title}".camelize %>
 
 <%= sage_component SagePanelHeader, {} do %>
-  <h2>HTML Code</h2>
+  <h2 id="<%= title %>-example-code">SageRails Component</h2>
+  <code><%= component_classname %></code>
 <% end %>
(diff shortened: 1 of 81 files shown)


docs/app/views/examples/shared/_code.html.erb
```diff
@@ -1,19 +1,38 @@
...
+<%= sage_component SageCard, {} do %>
+  <%= sage_component SageCardHeader, { title: "Preview Generated HTML Code"} %>
+  <div class="example__code">
+    <button class="example__expand-btn" aria-expanded="false" aria-controls="example-html-code-<%= title %>">Expand this code snippet</button>
+    <div id="example-html-code-<%= title %>" class="sage-code-snippet" tabindex="-1">
+      <pre class="prettyprint"><code><%= CGI.unescapeHTML(CGI.escapeHTML(render("examples/#{type.pluralize}/#{title}/preview"))) %></code></pre>
+    </div>
   </div>
-</div>
+<% end %>
```
AndrwM, line 38: This is for schema preview purposes in the Docs site

philschanely (approved):
This is going to be so helpful! Immense thanks for this work, @AndrwM !

PR #84, opened 2020-12-04
Sage Label React

AndrwM:
## Description
- Refactors secondary button outside of label so that handlers can be attached
- Remove magic color combo hex definitions in favor of sass color funcs
- React Label

We'll need to make some adjustments to the Rails Labels (along with heavy-handed arg checking) later on so that they better match the React arg API and to ensure an implementer uses the appropriate args together.

## Test notes
- Test label storybook labels to ensure it behaves as expected
- Test Rails flavored labels to ensure they still behave as expected
- Do a little dance! 🕺

### DIFF
--- a/packages/sage-assets/lib/stylesheets/tokens/_color_combos.scss
+++ b/packages/sage-assets/lib/stylesheets/tokens/_color_combos.scss
@@ -4,6 +4,14 @@
 /// @group sage
 ////
 
+
+// Color Overrides
+// NOTE: Stopgap until colors below are merged system color definitions
+$-label-primary-icon-background-accent: #3e92ee;
+$-label-warning-background-accent: #f9df8b;
+$-label-danger-background-accent: #f3988c;
+$-label-danger-icon-background-accent: #da665e;
+
 ///
 /// Sage color combinations token
 ///
@@ -14,7 +22,7 @@ $sage-color-combos: (
       foreground-accent: sage-color(charcoal, 500),
       background: sage-color(grey, 300),
       background-accent: sage-color(grey, 400),
-      inner-button-hover: sage-color(grey, 500),
+      icon-background-accent: sage-color(grey, 500),
     ),
     subtle: (
       foreground: sage-color(charcoal, 200),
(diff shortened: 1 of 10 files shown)


packages/sage-assets/lib/stylesheets/tokens/_color_combos.scss
```diff
@@ -4,13 +4,6 @@
 /// @group sage
 ////
 
-
-// Color overrides
-$-label-primary-icon-background-accent: #3e92ee;
```
philschanely, line 9: @AndrwM is the deletion in favor of the darken values below informed by design team members? Last I knew we were going to eventually graft these hex values into the official color spec as lik `n50` values rather than locally here or as darken values. 🤔

packages/sage-react/lib/Label/configs.js
```diff
@@ -6,8 +6,14 @@ export const LABEL_COLORS = {
   DANGER: 'danger',
```
philschanely, line 6: Do we need to update the color mappings? It seems the styles now allow for a `published` instead of `success` but perhaps `success` is a better system name for that color?

philschanely (commented):
Looking good, but a few questions/points of clarification before 👍

packages/sage-assets/lib/stylesheets/tokens/_color_combos.scss
AndrwM: No, it isn't. Its an attempt to make sure these orphaned colors don't diverge from the spec. I can swap it back.

AndrwM: I didn't realize this was a stopgap until we got updated the colors from design

philschanely: Its ok! We could perhaps have left a comment to clarify 😊 Thanks for the interest in keeping things tight!

packages/sage-react/lib/Label/configs.js
AndrwM: Good catch, thanks!

packages/sage-assets/lib/stylesheets/tokens/_color_combos.scss
AndrwM: Reverted :)

QuintonJason (approved):
Niiiiiice!

PR #85, opened 2020-12-04
Adjust popover to allow default styles and customization

philschanely:
- Restores default styling for content in popovers but adds a `sage-popover__content--custom` modifier that turns off default styling.
- Adds `custom_content_class` and logic in Rails for an open, customizable content panel and updates documentation of this setting.
- Adds `customContentClassName` and logic in React for the same.


pixelflips (approved):
👍  LGTM! Thanks for adjusting and keeping it flexible! Much appreciated!

PR #91, opened 2020-12-07
Update uses of Label to match new props

philschanely:
## Description

Seems some implementations of Label were using older props leading to server errors on Objects page.


AndrwM (approved):
My apologies @philschanely thanks!

QuintonJason (approved):

PR #95, opened 2020-12-08
Labels in Dropdowns

philschanely:
## Description

Upcoming work on People list page and the Panel Controls' Bulk Actions requires adding heading items to dropdown lists.

### Screenshots

|  Spec  |  Browser  |
|--------|--------|
| Screen Shot 2020-12-07 at 11 01 20 PM | Screen Shot 2020-12-07 at 11 01 52 PM (Scrolls) |

### Steps for testing
1. See the Dropdown Object in the Documentation site and the new entry for Dropdown menu with Headings.
1. See the Dropdown component in Storybook and the new entry for Dropdown menu with Headings.


AndrwM (commented):
I have concerns but I won't block this, however, I'd think we should look into restructuring asap.

A few notes..
- With this implementation headings will also be filtered in Search.
- With this implementation my preference would be to use `type` rather than `isHeading` to allow for future extensibility.
- I recommend we split `DropdownItem` into multiple subcomponents. Rather than having different render permutations based on vars being set. At this point `DropdownItem` can render 4 different things depends on props passed.

I think it'd be worthwhile to take this back up to Design and ask for a pattern on usage of headings and lines... My suggestion here would be that we have a DropdownSection pattern that can be searched within, can contain a heading, and can have lines (either top or bottom).

philschanely:
Great points, @AndrwM. Hoping we can move forward for now. In response to our concerns:

- This variation ought not to be used with a search dropdown but a standard one. Perhaps we can look into clarifying/preventing the mixing of types in a follow up.
- Funny you said that as I started of with `type` there! 😂  I went with `isHeading` for the sake of simplicity and known cases now. Seems we could have done this earlier with the checkbox and label alternatives. But rather than tackle that refactor now, lets punt to a follow for that too.
- To the prior two points, your third is absolutely valid and I agree a refactor is overdue. But not a priority right now. 😢 

In regard to talking with Design, my short take is that the line designates larger groups while the headings provide subgroups. It was an attempt to avoid having submenus flying everywhere 🤪 . Great topic for a future Sage Squad meeting.

QuintonJason (approved):
Approved. Agreed with the above comments. Once we have the bandwidth, we'll get it refactored

kimjaej (approved):

PR #101, opened 2020-12-10
Sage Typeahead

AndrwM:
## Description
- Typeahead component
- New Sage `useFocusTrap` hook

### Screenshots
screencast 2020-12-09 20-10-12

## Test notes
:)

## Related
TBD
-


packages/sage-assets/lib/stylesheets/patterns/objects/_typeahead.scss
```diff
@@ -0,0 +1,88 @@
...
+
+  &:hover,
+  &:focus-within {
+    background-color: sage-color(grey, 100);
+  }
+}
+
+.sage-typeahead__item-trigger {
+  @include sage-button-style-reset;
+  @include sage-focus-outline($outline-offset-block: -6px, $outline-offset-inline: -6px);
```
AndrwM, line 50: `sage-focus-outline` won't accept rem values based on the math happening internally.

packages/sage-react/lib/Typeahead/TypeaheadItem.jsx
```diff
@@ -0,0 +1,55 @@
...
+      >
+        {title}
+      </div>
+      <div
+        className="t-sage-body-small"
+        style={{gridArea: 'subTitle'}}
+      >
+        {subTitle}
+      </div>
+    </button>
```
AndrwM, line 35: Excited about this pattern. This are boilerplate elements with an additional grid area style that positions the element

docs/lib/sage_rails/app/views/sage_components/_sage_typeahead.html.erb
```diff
@@ -0,0 +1,28 @@
...
+      </button>
+      <div class="sage-typeahead__item-actions">
+        <% item[:actions].each do |action| %>
+          <%= action.html_safe %>
+        <% end %>
+      </div>
+    </li>
+    <% end %>
+  </ul>
+</div>
```
AndrwM, line 28: Unconcerned with a11y support here, this component is for preview only

packages/sage-react/lib/hooks/useFocusTrap.jsx
```diff
@@ -0,0 +1,53 @@
...
+      setup();
+    } else {
+      cleanup();
+    };
+
+    return () => cleanup();
+  }, [active, containerRef.current]);
+};
+
+export default useFocusTrap;
```
AndrwM, line 53: 😈 We now have a flexible focus trapping solution for any react component. This additionally adds `esc` and click outside support.

docs/lib/sage_rails/app/views/sage_components/_sage_typeahead.html.erb
```diff
@@ -0,0 +1,28 @@
...
+    value: nil,
+    contained: true,
+    label_text: "Search"
+  } %>
+  <ul class="sage-typeahead__panel">
+    <% component.items.each do |item| %>
+      <li class="sage-typeahead__item">
+      <button class="sage-typeahead__item-trigger">
+        <i class="sage-icon-<%= item[:icon] %>" style="grid-area: icon;"></i>
+        <div class="t-sage-heading-6" style="grid-area: title;">
```
philschanely, line 13: would `p`, `b`, and/or `span` be ok here rather than `div`? I prefer to avoid `div` when I have text content as direct children and instead look for the best typographical element for the context.

packages/sage-react/lib/Typeahead/Typeahead.jsx
```diff
@@ -0,0 +1,77 @@
...
+            role="listbox"
+            id={A11Y_ID}
+          />
+      }
+    </div>
+  );
+};
+
+Typeahead.defaultProps = {
+};
```
philschanely, line 71: if `items` aren't required consider providing a default value here. Empty array perhaps?

packages/sage-react/lib/Typeahead/TypeaheadItem.jsx
philschanely: 😍

packages/sage-react/lib/hooks/useFocusTrap.jsx
philschanely: ❤️ ❤️ ❤️ 🙌

docs/lib/sage_rails/app/sage_components/sage_typeahead.rb
```diff
@@ -0,0 +1,3 @@
+class SageTypeahead < SageComponent
```
philschanely, line 1: What are your thoughts on creating this component but not having accompanying system JS? is that a future development plan beyond this PR? If not, I wonder how we acknowledge a component as React-only and then don't even build the Rails piece that tempts it to be used without full support. 🤔

philschanely (approved):
Looking really great, @AndrwM . A few comments, questions but nothing blocking.

packages/sage-react/lib/Typeahead/Typeahead.jsx
```diff
@@ -0,0 +1,77 @@
...
+  }, [searchValue]);
+
+  return (
+    <div
+      className="sage-typeahead"
+      ref={containerRef}
+      aria-expanded={open}
+      aria-haspopup="listbox"
+      aria-owns={A11Y_ID}
+      {...rest}
```
joedakroub, line 44: Nice use of ARIA here; solid! ❤️

packages/sage-react/lib/Typeahead/TypeaheadItem.jsx
joedakroub: Very clever; I've not seen anyone use a `grid-area` in an inline style before, smart! 🥇

joedakroub (approved):
Really nice work here Andrew! Thanks! 👍

packages/sage-react/lib/Typeahead/Typeahead.jsx
```diff
@@ -0,0 +1,77 @@
...
+          setSearchValue(e.target.value);
+          e.target.value.length ? setOpen(true) : setOpen(false);
+        }}
+        onClear={() => {
+          setSearchValue('');
+          setOpen(false);
+        }}
+        value={searchValue}
+      />
+      {open
```
JulaB, line 58: Nit: We probably can use only `searchValue` to determine if it's open or not, instead of `open` state.

`const open = searchValue.length > 0;`

JulaB (approved):
👍 This looks great @AndrwM! Just left a couple of suggestions

QuintonJason (approved):

PR #102, opened 2020-12-10
Label Positioning and Structure for Affixed Inputs

pixelflips:
## Description
Labels applied to affixed form inputs are missing styling and are positioned incorrectly.
These changes update the structure of the label and adjust the styling for better positioning.

### Screenshots
|  before  |  after  |
|--------|--------|
|Screen Shot 2020-12-09 at 5 55 36 PM|Screen Shot 2020-12-09 at 5 56 34 PM|

### DIFF
--- a/packages/sage-assets/lib/stylesheets/patterns/elements/_form_input.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/elements/_form_input.scss
@@ -49,7 +49,8 @@ $-input-transition: border 0.2s ease-in-out, box-shadow 0.2s ease-in-out;
   top: sage-spacing(xs);
   z-index: sage-z-index(default, 1);
 
-  &.sage-label {
+  .sage-input--suffixed &,
+  .sage-input--prefixed & {
     position: absolute;
   }
 
@@ -89,6 +90,7 @@ $-input-transition: border 0.2s ease-in-out, box-shadow 0.2s ease-in-out;
   .sage-input--prefixed & {
     transform: translateY(calc(#{$-input-text-height} * -1)) translateX(30px);
     z-index: sage-z-index(default, 2);
+    position: absolute;
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/sage-assets/lib/stylesheets/patterns/elements/_form_input.scss
```diff
@@ -49,6 +49,10 @@ $-input-transition: border 0.2s ease-in-out, box-shadow 0.2s ease-in-out;
   top: sage-spacing(xs);
   z-index: sage-z-index(default, 1);
 
+  &.sage-label {
```
monkeypox8, line 52: An alternative here would be using the related parent selectors. Same specificity, but arguably less of a hassle to override/unravel:
```
  .sage-input__prefixed &,
  .sage-input__suffixed & {
    position: absolute;
  }
```

monkeypox8 (approved):
👍🏼 🔥 LGTM! Thanks for jumping on this fix.

packages/sage-system/lib/inputaffixes.js
```diff
@@ -44,22 +45,27 @@ Sage.inputaffixes = (() => {
...
 
     return elLabel;
   };
 
   const handleAffixClick = (ev) => {
-    if (ev.target.classList.contains(affixClass)) {
+    if (ev.target.classList.contains(labelValueClass)) {
       // Find neighboring input and focus on it
-      ev.target.parentNode.querySelector(`.${fieldClass}`).focus();
+      ev.target.parentNode.parentNode.querySelector(`.${fieldClass}`).focus();
```
joedakroub, line 68: Another option is to use the `closest` function like so:

```ev.target.closest(`.${fieldClass}`).focus();```

https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

joedakroub (approved):
Changes look good Phillip! Thank you! 👍

packages/sage-system/lib/inputaffixes.js
pixelflips: @joedakroub I had initially tried this but keep getting an error and couldn't quite work it out:

`Uncaught TypeError: Cannot read property 'focus' of null`

Any thoughts?

joedakroub: Hmm, sounds like `closest` is missing it. Where is the event target in relation to the element you are searching for? What you have may be best.

pixelflips: @joedakroub I went ahead and merged this one in order to address a small bug in products, but I will put up another fix once I work it out. Thanks for having a look though.

PR #106, opened 2020-12-11
Table tweaks

philschanely:
## Description

This PR makes a few necessary adjustments to Tables and OptionsDropdown for Manage team work in Custom Domains and Contacts.

### DIFF
--- a/packages/sage-assets/lib/stylesheets/patterns/elements/_table.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/elements/_table.scss
@@ -289,6 +289,14 @@ $-table-sort-indicator-direction: rem(7px);
   text-overflow: ellipsis;
 }
 
+.sage-table-cell--checkbox {
+  width: rem(64px);
+}
+
+.sage-table-cell--avatar {
+  width: rem(52px);
+}
+
 .sage-table-cell--align-right {
   text-align: right;
 }
--- a/packages/sage-assets/lib/stylesheets/patterns/elements/_table.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/elements/_table.scss
@@ -202,12 +202,12 @@ $-table-sort-indicator-direction: rem(7px);
 .sage-table-wrapper {
   position: relative;
 
-  .sage-panel > & {
+  .sage-panel & {
     margin-left: -1 * $-table-cell-padding-panel;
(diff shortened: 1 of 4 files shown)


packages/sage-react/lib/Table/TableRow.jsx
```diff
@@ -70,8 +70,16 @@ const TableRow = ({
...
+          'sage-table-cell',
+          className,
+          {
+            [`sage-table-cell--${dataType}`]: dataType,
+          }
+        )
+
         return (
-          <td key={uuid()} className={className} style={style} {...attributes}>
+          <td key={uuid()} className={cellClassnames} style={style} {...attributes}>
```
monkeypox8, line 82: Just out of curiosity, what's the `key` attribute used for here?

packages/sage-assets/lib/stylesheets/patterns/elements/_table.scss
```diff
@@ -202,12 +202,12 @@ $-table-sort-indicator-direction: rem(7px);
 .sage-table-wrapper {
   position: relative;
 
-  .sage-panel > & {
+  .sage-panel & {
     margin-left: -1 * $-table-cell-padding-panel;
```
monkeypox8, line 206: Thoughts around solidifying this (negative offset) pattern as a mixin? I think we're using it in a couple of other places as well.

monkeypox8 (approved):
👍🏼 LGTM!

packages/sage-assets/lib/stylesheets/patterns/elements/_table.scss
philschanely: In favor of it. Follow up ok?

packages/sage-react/lib/Table/TableRow.jsx
philschanely: `key` is used by React to provide a unique identifier for items that are generated through a loop as the table cells are here. So rather than assuming every cell will have an id or even a field name from the api, I just use the uuid() to generate a hash. Is that what you were looking for?

monkeypox8: Ahh, I see. I saw that an `id` was getting passed into the parent row as a data attribute and wasn't certain why this was using a different method, so that makes sense!

joedakroub (approved):
Changes look good Phil! No concerns. 👍

PR #107, opened 2020-12-11
Loader - updated properties references

QuintonJason:
## Description

- [x] - updated references from `shape` to `type` to remove console errors

### Steps for testing
1. Visit panel react page: http://localhost:6006/?path=/story/sage-panel--default
2. Click the `loading` button and verify that the spinner appears


JulaB (approved):
👍 Looking good

PR #113, opened 2020-12-15
Outline Item - resolved misalignment issue with loader

QuintonJason:
## Description

* resolve misalignment with the loader in the outlineitem component

### Screenshots

|  before  |  after  |
|--------|--------|
|Screen Shot 2020-12-15 at 9 59 20 AM|Screen Shot 2020-12-15 at 9 58 57 AM|

### Steps for testing
1. With the bridge on visit an outlines page within the app, url something like http://www.kajabi.test:3000/admin/products/1


philschanely (approved):
👍  Just curious; does `center` make any sense here? If not, no concerns, but I wondered whether `center` would fit the case better?

QuintonJason:
> 👍 Just curious; does `center` make any sense here? If not, no concerns, but I wondered whether `center` would fit the case better?

@philschanely Weirdly the parent, `.sage-outline-item__actions-primary`'s align-items: center` is centering it, with the loader having `position: relative` and `transform` the `flex-start` resets it back to center of the parent; if that makes sense

PR #115, opened 2020-12-15
refactor(rails searchjs): delete search query string param if user clears search

AndrwM:
## Description
Rails searchjs update:
Previously when search was cleared we would clear the query string param value but not the query string param key, now we clear both.

@pixelflips and I paired on this to resolve an issue with searching in the Market domain. 

### Screenshots
n/a

## Test notes
When search is cleared ensure the query string key is also deleted

## Related
n/a

### DIFF
--- a/packages/sage-system/lib/search.js
+++ b/packages/sage-system/lib/search.js
@@ -48,11 +48,17 @@ Sage.search = (function() {
   }
 
   function search(value) {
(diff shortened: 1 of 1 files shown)


joedakroub (approved):
No concern here. 👍

pixelflips (approved):
👍🏼  LGTM!

PR #122, opened 2020-12-16
Update Sage icon font

joedakroub:
## Description
This PR makes changes to the Sage icon font:

### Added:
- icon-ban
- icon-lab
- icon-success
- icon-super-admin

### Removed:
- icon-add-image
- icon-folder-group
- icon-sort

### Steps for testing
1. Visit `/pages/design/icon` in the Sage Design System
2. Ensure all icons match their name


philschanely (approved):
Assuming the deletions are intentional and will be patched where needed 👍

monkeypox8 (approved):
👍🏼 LGTM

re: our earlier discussion around the unicode entries changing on each icomoon export, hopefully we can figure out a way to avoid that. It does seem to be error-prone with so many values shifting.

PR #125, opened 2020-12-17
Sage Docs mobile header fixes

monkeypox8:
### Description
Links in the site header/assistant were breaking the layout on mobile devices and preventing the nav from displaying, so this is just a quick patch until we have a chance for a permanent fix. Header links are hidden from view, and the Release Notes and Changelog links moved to the sidebar. Also, as we're already limited on real estate in the mobile sidebar, the version numbers have been left out for the time being.
 

### Screenshots
|  Before (nav menu unreachable)   |  After  |
|--------|--------|
|before|after|

### DIFF
--- a/docs/app/views/application/_sidebar.html.erb
+++ b/docs/app/views/application/_sidebar.html.erb
@@ -203,6 +203,24 @@ layout_pages = [
       <ul class="sage-nav__list">
         <li><%= link_to "Sandbox", pages_sandbox_path, class: "sage-nav__link #{"sage-nav__link--active" if current_page?(pages_sandbox_path)}" %></li>
         <li><%= link_to "Code Guidelines", "https://github.com/Kajabi/sage/wiki", class: "sage-nav__link", target: "_blank", rel: "noopener noreferrer" %></li>
+        <li class="sage-col--sm-show">
+          <%= link_to "Release Notes",
+            $SAGE_RELEASE_URL,
+            class: "sage-nav__link",
+            target: "_blank",
+            rel: "noopener noreferrer",
+            title: "Github release notes"
+          %>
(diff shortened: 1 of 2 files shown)


docs/app/views/application/_sidebar.html.erb
```diff
@@ -203,6 +203,24 @@ layout_pages = [
...
+            $SAGE_RELEASE_URL,
+            class: "sage-nav__link",
+            target: "_blank",
+            rel: "noopener noreferrer",
+            title: "Github release notes"
+          %>
+        </li>
+        <li class="sage-col--sm-show">
+            <%= link_to "Changelog",
+              "https://github.com/Kajabi/sage/releases",
```
pixelflips, line 217: Should these point to sage-lib?

pixelflips (approved):
LGTM! 👍🏼 - left a comment about updating a link to point to sage-lib  (there's a few more) but definitely not a blocker here. Nice work!

docs/app/views/application/_sidebar.html.erb
monkeypox8: Oh, great catch. Yeah, there's quite a few references scattered in the docs pointing to the old repo. Might just open up an issue as a fast follow to clean those up

pixelflips: Sounds great to me!

philschanely (approved):

QuintonJason (approved):

docs/app/views/application/_sidebar.html.erb
monkeypox8: Cool, I've created #128 as a follow-up

PR #127, opened 2020-12-17
refactor(usefocustrap): update arg names in useFocusTrap to be more indicative of their internal use

AndrwM:
## Description
Tiny PR:
refactor(usefocustrap): update arg names in useFocusTrap to be more indicative of their internal use

I just landed this yesterday, on a second look `onDeactivateCallback` was a bad name because it implies that this will fire on a deactivating event rather than this function _being the function that handles deactivation_.

## Test notes
n/a

## Related
n/a


joedakroub (approved):
Looking good; no concerns here. 👍

philschanely (approved):

PR #135, opened 2020-12-18
Remove Sage-Outline-Item

AndrwM:
## Description
Removing Sage-OutlineItem from Sage.

Because...
- We're using a 3rd party lib to handle the multi-level drag & drop which requires css classes to be closely married to the implementation
- We have the functionality for outline items split across 5+ react components for render optimization
- This is a unicorn component (and was originally planned to be one)
- This doesn't exist in the Design Sys docs
- This implementation cannot be QA'ed within a Sage env because of the depth of interactions within the actual Kjb implementation.

@QuintonJason and I talked about this at length and think this is the appropriate pattern here.

See https://github.com/Kajabi/kajabi-products/pull/17100 for the migration.

### Screenshots
n/a

## Test notes
n/a

## Related
n/a


QuintonJason (approved):
Since this is a one-off component in which the implementation is specific to the build page, I believe this component should live in the app as well. I'd love to hear a case for keeping it in sage aside from it being a component.

philschanely (approved):
✅

PR #143, opened 2020-12-23
Tabs - updates to react choice tabs

QuintonJason:
## Description

* added `sage-tabs--choice` class to react to align with rails implementation

### Steps for testing
1. http://localhost:6006/?path=/docs/sage-tabs--default


AndrwM (approved):
Wow, I'm surprised that we didn't have this +1. Thanks for catching

PR #146, opened 2021-01-04
React Tooltip Alignment

AndrwM:
## Description
Fixes tooltip alignment issue when view is scrolled. Previously the React Tooltip component wasn't taking into account the `window.pageYOffset` like the non-React implementation was. This resolves that issue and also moves the distance setting to a `const`.

### Screenshots
n/a

## Test notes
Enter into a scrolled state within storybook and ensure the tooltip is properly aligned.

## Related
https://github.com/Kajabi/kajabi-products/pull/17164


philschanely (approved):
Good call.

PR #147, opened 2021-01-04
Docs Rails Description

philschanely:
## Description

Uses Rails component for Description and removes older version. No visual change or impact on implementations.


monkeypox8 (approved):
👍🏼 LGTM! I've added issue #150 to address the file output when using the Rails generators

pixelflips (approved):
LGTM! 👍🏼

PR #151, opened 2021-01-04
Icon card components

philschanely:
## Description

This PR continues efforts to tie up loose ends in the system. It adds Rails and React component versions of the IconCard element, updates documentation, and implements these components where they are used in UploadCard.

No visual changes or impact in `products`.


docs/app/views/examples/elements/icon_card/_preview.html.erb
```diff
@@ -1,7 +1,7 @@
 <%
 colors = [
```
QuintonJason, line 2: Future enhancement: is it possible to update the object to contain a label for instances in which the label will be passed to `aria-label`?

docs/lib/sage_rails/app/views/sage_components/_sage_icon_card.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  class="
+    sage-icon-card
+    <%= "sage-icon-card--#{color}" %><%= "-#{component.style}" if component.style.present? %>
+    <%= component.css_classes if component.css_classes.present? %>
+  "
+>
+  <i
+    class="sage-icon-<%= component.icon %>-<%= size %>"
+    <% if component.label.present? %>
+    aria-label="<%= component.label %>"
```
QuintonJason, line 15: Reference to `aria-label` to showcase this conditional

docs/app/views/examples/elements/icon_card/_preview.html.erb
QuintonJason: This comment maybe mute since I see an example in uploadcard

philschanely: Its a good call to show potential variations, and in this case, its probably good to show a label provided as a suggested default.

QuintonJason:
in `sage-react/lib/Icon/Icon.jsx` https://github.com/Kajabi/sage-lib/blob/master/packages/sage-react/lib/Icon/Icon.jsx#L25, `aria-hidden` isn't firing when `null`, can we update the conditional to be something like `if (label === '' || label === null)`?

QuintonJason (approved):
💥

pixelflips (approved):
👍🏼  LGTM! Nice work!

PR #153, opened 2021-01-04
Catalog item content spacing

monkeypox8:
## Description
[Catalog items](https://sage-design-system.kajabi.com/pages/object/catalog_item) can include any combination of sub-items within the `sage-catalog-item__content` area. Because spacing between child elements is being applied specifically to buttons at present, this PR extends the margin to all child elements (see below examples for a mix of `label` and `span`).

### Screenshots

|  before  |  after  |
|--------|--------|
|before|after|

### Related
- SELL-578

### DIFF
--- a/packages/sage-assets/lib/stylesheets/patterns/objects/_catalog_item.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/objects/_catalog_item.scss
@@ -90,8 +90,8 @@ $-image-height-mobile: rem(120px);
   align-self: flex-start;
   flex-flow: row wrap;
 
-  > *:not(:last-child) {
-    margin-right: sage-spacing();
+  > * + * {
+    margin-left: sage-spacing();
   }
 }
 
--- a/packages/sage-assets/lib/stylesheets/patterns/objects/_catalog_item.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/objects/_catalog_item.scss
@@ -88,7 +88,11 @@ $-image-height-mobile: rem(120px);
   display: flex;
   grid-area: content;
   align-self: flex-start;
-  flex-wrap: wrap;
+  flex-flow: row wrap;
+
+  > *:not(:last-child) {
+    margin-right: sage-spacing();
+  }
 }
 
 .sage-catalog-item__aside {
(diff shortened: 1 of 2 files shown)


packages/sage-assets/lib/stylesheets/patterns/objects/_catalog_item.scss
```diff
@@ -88,7 +88,11 @@ $-image-height-mobile: rem(120px);
   display: flex;
   grid-area: content;
   align-self: flex-start;
-  flex-wrap: wrap;
+  flex-flow: row wrap;
+
+  > *:not(:last-child) {
```
monkeypox8, line 93: Couldn't decide between this selector and the lobotomized owl (`* + *`) with `margin-left`

QuintonJason: Solely based on the level of specificity I would be in favor of the lobotomized owl as it has one class vs the class and pseudo-class increased specificity of the `*:not(:last-child)`. It only beats it by one so it doesn't jump high enough to worry me currently.

QuintonJason (approved):

philschanely (approved):

packages/sage-assets/lib/stylesheets/patterns/objects/_catalog_item.scss
```diff
@@ -88,7 +88,11 @@ $-image-height-mobile: rem(120px);
   display: flex;
   grid-area: content;
   align-self: flex-start;
-  flex-wrap: wrap;
+  flex-flow: row wrap;
+
+  > * + * {
```
philschanely, line 93: Just curious, is `:not(:first-child)` any clearer to others?

monkeypox8: `:not(:first-child)` (and `:not(:last-child)` is a little easier to parse, but I think it comes down to readability vs the small advantage in lowered specificity as @QuintonJason mentioned above. With the case here of an unknown makeup of child elements, I'd lean towards the less intrusive method.

philschanely: oh haha I did not see @QuintonJason's comment before! 🤦

PR #154, opened 2021-01-05
Adjust panel controls subcomponents

philschanely:
## Description

This PR makes a few non-visual improvements to the Panel Controls to help with implementing them in context. 

- Selection releases the `id` rather than a checkbox value, which solves a data type conflict (checked values always emit strings that don't compare the same to `id` values for other purposes.
- Added subcomponents for tighter composition and revised `children` wrapper.
- Made a few other tweaks to utilities and the story for easier copy-and-paste of the example there.

### DIFF
--- a/packages/sage-assets/lib/stylesheets/patterns/elements/_search.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/elements/_search.scss
@@ -54,25 +54,12 @@ $-search-icon: "::before";
   &.sage-search--contained {
     position: relative;
     z-index: sage-z-index(default);
+    border-radius: sage-border(radius);
     box-shadow: map-get($sage-toolbar-button-borders, default);
     border: 0;
 
     &::after {
       display: none;
-    }
-
-    .sage-panel-controls__toolbar &:first-child {
-      border-top-left-radius: sage-border(radius);
-      border-bottom-left-radius: sage-border(radius);
-    }
-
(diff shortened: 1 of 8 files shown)


packages/sage-react/lib/PanelControls/PanelControls.story.jsx
```diff
@@ -13,7 +13,7 @@ import PanelControls from './PanelControls';
 const PanelControlsWithData = () => {
   const [selfData, setSelfData] = useState({
     // Set locally
-    articles: [],
+    items: [],
```
joedakroub, line 16: Good call. 👍

joedakroub (approved):
Nice cleanup and re-org on the styles and variable names. No concerns here. Thanks Phil! 👍

packages/sage-assets/lib/stylesheets/patterns/elements/_search.scss
```diff
@@ -84,6 +71,16 @@ $-search-icon: "::before";
       z-index: sage-z-index(default, 2);
       box-shadow: map-get($sage-toolbar-button-borders, focus);
     }
+
+    .sage-panel-controls__toolbar &:first-child:not(:only-child) {
```
joedakroub, line 75: Good catch on re-organizing these styles.

QuintonJason (approved):

PR #155, opened 2021-01-05
Adjustments to toast functionality

pixelflips:
## Description
Updates to toast display to be triggered by clicking on an element (button) within docs site.
In addition, allows toast to be displayed automatically by triggering an event.

### Screenshots
Current visuals do not change with these updates.

### Steps for testing
1. Visit http://localhost:4000/pages/breakout/element/toast
2. Click on the buttons and verify toast appears.


philschanely:
@pixelflips Nice start here. This idea is much needed to clean up the docs site. A few concerns with the current solution:

- Toast is simple enough that passing content this way should be ok, but I wonder about long-term scalability in passing all the HTML content through attrs like this.
- I'm not excited about the disconnect this causes by using JS to add the element; seems like a potential for things to get out of sync if we're not careful.

What do you think about a solution that toggles an element on and off like we do with modals or Quinton's recent work on Popovers? In fact, the React component has some modality built into it for triggering the element on demand. This could be set to fire automatically or to wait for a click from a paired trigger? 

Regardless thanks for taking this on! LMK if you'd like to talk more.

packages/sage-system/lib/toast.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  function unbind(el) {
+    el.removeEventListener("click", removeToast);
+  }
+
+  // toast template
+  function buildToast(evt) {
+
+    let toast = document.createElement("dialog");
```
QuintonJason, line 24: @pixelflips Can this parent element be `output` instead of `dialog`? Is will allow provide the two aria attributes , `aria-live="assetive"` and `aria-atomic="true"`; causing the toast to announce it's text to screen readers and other assistive technologies upon insertion.

pixelflips:
> @pixelflips Nice start here. This idea is much needed to clean up the docs site. A few concerns with the current solution:
> 
> * Toast is simple enough that passing content this way should be ok, but I wonder about long-term scalability in passing all the HTML content through attrs like this.
> * I'm not excited about the disconnect this causes by using JS to add the element; seems like a potential for things to get out of sync if we're not careful.
> 
> What do you think about a solution that toggles an element on and off like we do with modals or Quinton's recent work on Popovers? In fact, the React component has some modality built into it for triggering the element on demand. This could be set to fire automatically or to wait for a click from a paired trigger?
> 
> Regardless thanks for taking this on! LMK if you'd like to talk more.

@philschanely - since this was just an initial stab at some cleanup, your suggestions definitely make sense. Let me check out the modal and popover and I might have some questions.... stay tuned and thanks again!

packages/sage-system/lib/toast.js
```diff
@@ -0,0 +1,69 @@
...
+    el.querySelector(SELECTOR_TOAST_CLOSE).addEventListener("click", handleCloseToast);
+  }
+
+  function unbind(el) {
+    el.querySelector(SELECTOR_TOAST_CLOSE).removeEventListener("click", handleCloseToast);
+  }
+
+  function trigger(config) {
+    let toast = document.createElement("dialog");
+    let toastMessage = `<p class="sage-toast__value">${config.message}</p>`;
```
philschanely, line 24: I still wish we could get around without needing to do this, but at least its only a little content. Would you consider this so the whole can be seen a little better?

```js
toast.innerHTML = `
  <p class="sage-toast__value">${config.message}</p>
  <button
    type="button"
    data-js-toast-close
    class="sage-toast__close sage-btn sage-btn--subtle sage-btn--secondary sage-btn--icon-only-remove"
  >
    <span class="visually-hidden">Close</span>
  </button>
  <i class="sage-toast__icon sage-icon-${config.icon}"></i>
`;
```

philschanely (approved):
One suggestion for the JS markup

packages/sage-system/lib/toast.js
pixelflips: Great suggestion... thanks @philschanely! will adjust right away!

packages/sage-system/lib/toast.js
```diff
@@ -0,0 +1,69 @@
...
+      }, 4000);
+    }
+  }
+
+  // removes toast when timer is up
+  function removeToast(toast) {
+    window.requestAnimationFrame(function () {
+      document.body.removeChild(document.querySelector(`.${TOAST_CLASS}`));
+    });
+  }
```
AndrwM, line 55: The `toast` argument isn't being referenced here. From the code it seems like this may close all toasts when the timer is up.

packages/sage-system/lib/toast.js
```diff
@@ -0,0 +1,69 @@
...
+
+  function init(el) {
+    el.querySelector(SELECTOR_TOAST_CLOSE).addEventListener("click", handleCloseToast);
+  }
+
+  function unbind(el) {
+    el.querySelector(SELECTOR_TOAST_CLOSE).removeEventListener("click", handleCloseToast);
+  }
+
+  function trigger(config) {
```
AndrwM, line 22: For the `config` could you set up some defaults so that `message` is the only required key/value? This will make it easy for devs to use in Products and also matches the current implementation of `App.flash`

packages/sage-system/lib/toast.js
```diff
@@ -0,0 +1,69 @@
...
+      document.body.removeChild(document.querySelector(`.${TOAST_CLASS}`));
+    });
+  }
+
+  // handles click on toast close btn
+  function handleCloseToast(evt) {
+    window.requestAnimationFrame(function () {
+      document.body.removeChild(evt.target.parentElement);
+    });
+  }
```
AndrwM, line 62: When the close is clicked we probably want to stop the timer as well. It may not be a big deal performance-wise, so long as an error isn't being fired when the timer finishes but the toast has already been removed.

AndrwM (commented):
This is great 👍 left a few comments.

AndrwM (approved):
This lgtm 🥇

PR #156, opened 2021-01-06
Conditionally render Sortable item subtitles (Rails)

monkeypox8:
## Description
Corrects an issue when using the Rails version of the Sortable item component where `H2` elements for the subtitle are output, regardless of whether a value is being provided. The benefit of this change is primarily for [accessibility](https://dequeuniversity.com/rules/axe/3.5/empty-heading).

Note: the React version of this component is not affected, since it already performs this check before being rendered.

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_sortable_item.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_sortable_item.html.erb
@@ -8,7 +8,9 @@
 >
   <div class="sage-sortable__item-content">
     <h1 class="sage-sortable__item-title"><%= component.title %></h1>
-    <h2 class="sage-sortable__item-subtitle"><%= component.subtitle %></h2>
+    <% if component.subtitle %>
+      <h2 class="sage-sortable__item-subtitle"><%= component.subtitle %></h2>
+    <% end %>
   </div>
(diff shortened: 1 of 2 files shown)


pixelflips (approved):
LGTM! 👍🏼

docs/lib/sage_rails/app/views/sage_components/_sage_sortable_item.html.erb
```diff
@@ -8,7 +8,9 @@
 >
   <div class="sage-sortable__item-content">
     <h1 class="sage-sortable__item-title"><%= component.title %></h1>
-    <h2 class="sage-sortable__item-subtitle"><%= component.subtitle %></h2>
+    <% if component.subtitle %>
```
philschanely, line 11: Non-blocking but I've been doing like `if component.subtitle.present?` for what I understand to be a little safer of a conditional for property presence in case it's of interest to you.

philschanely (approved):
One non-blocking comment.

docs/lib/sage_rails/app/views/sage_components/_sage_sortable_item.html.erb
monkeypox8: Great call, I'll make the change.

PR #163, opened 2021-01-07
Rails tooltip new window

philschanely:
## Description

Adds `target="_blank"` to popovers Rails component in line with needs in app.

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_popover.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_popover.html.erb
@@ -62,7 +62,8 @@
             name: "launch"
           },
           attributes: {
-            href: component.link[:href]
+            href: component.link[:href],
+            target: "_blank",
           }
         } %>
       </div>
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_popover.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_popover.html.erb
@@ -64,6 +64,7 @@
           attributes: {
             href: component.link[:href],
             target: "_blank",
+            rel: "noreferrer noopener"
           }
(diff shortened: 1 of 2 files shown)


docs/lib/sage_rails/app/views/sage_components/_sage_popover.html.erb
```diff
@@ -62,7 +62,8 @@
             name: "launch"
           },
           attributes: {
-            href: component.link[:href]
+            href: component.link[:href],
+            target: "_blank",
```
pixelflips, line 66: with the `target: "blank"` should we also include `rel="noopener noreferrer"`? I see this throughout the app.

pixelflips (approved):
👍🏼  LGTM! One minor question, but not a show-stopper what so ever!

docs/lib/sage_rails/app/views/sage_components/_sage_popover.html.erb
monkeypox8: Agreed, and I'm reasonably sure that only one of the two attributes should be enough. I'd lean towards `noopener` since that allows us the possibility for tracking these links.

monkeypox8 (approved):
👍🏼 LGTM! Echoing Phillip's statement around the `target="blank"` links.

docs/lib/sage_rails/app/views/sage_components/_sage_popover.html.erb
QuintonJason: Yeah, I thought we needed both attrs `noopener` and `noreferrer`

QuintonJason (approved):

philschanely:
Thanks, Gents! I've adjusted the `rel` attribute on both React and Rails components.

PR #164, opened 2021-01-07
React toast default state

philschanely:
## Description

This small patch ensures React Toast is synchronized with the value passed to `isActive` right from the start. This solves a blinking Toast issue in some browsers.


pixelflips (approved):
👍🏼  LGTM!

QuintonJason (approved):

PR #165, opened 2021-01-08
Corrections to Rails generator file creation paths

### LINKED ISSUE #150: [BUG] Incorrect paths in Rails generators
## Description
Using the `rails generate` scripts for Elements, Objects, etc. throw errors due to outdated stylesheet output paths (see screenshot below).

We'll also want to update the [wiki](https://github.com/Kajabi/sage/wiki/Contributing) to include the working directory (`<root>/docs`) for running these scripts.

```
    style_file = "lib/sage-frontend/stylesheets/system/patterns/elements_#{file_name}.scss"
    style_include_file = "lib/sage-frontend/stylesheets/system/index.scss"
```

## Screenshots
Screen Shot 2021-01-04 at 11 58 12 AM

monkeypox8:
## Description
The URLS for the Rails element/object/token file generators are out-of-date and throw errors when run. This update fixes the paths and fills out some missing documentation.

### Steps for testing
1. From the command-line, navigate to `docs` from the root directory
2. Run one of the [Sage generators from the wiki](https://github.com/Kajabi/sage-lib/wiki/Contributing) and confirm that the associated files are correctly output
3. Repeat step 2 with a different generator

## Related

- closes #150

### DIFF
--- a/docs/lib/generators/sage_token/sage_token_generator.rb
+++ b/docs/lib/generators/sage_token/sage_token_generator.rb
@@ -5,7 +5,7 @@ class SageTokenGenerator < Rails::Generators::NamedBase
     # Style Variables
     style_file = "../packages/sage-assets/lib/stylesheets/tokens/_#{file_name}.scss"
     style_include_file = "../packages/sage-assets/lib/stylesheets/tokens/_index.scss"
-    style_include_line = "// Variable Tokens"\
+    style_include_line = "// Variable Tokens"
     # Create Style File
     template "style.scss", style_file
     # Include Style File
--- a/docs/lib/generators/sage_token/sage_token_generator.rb
+++ b/docs/lib/generators/sage_token/sage_token_generator.rb
@@ -3,14 +3,14 @@ class SageTokenGenerator < Rails::Generators::NamedBase
   def create_sage_token
 
     # Style Variables
-    style_file = "lib/sage-frontend/stylesheets/system/tokens/_#{file_name}.scss"
-    style_include_file = "lib/sage-frontend/stylesheets/system/index.scss"
-    style_include_line = "// Variable Tokens"
+    style_file = "../packages/sage-assets/lib/stylesheets/tokens/_#{file_name}.scss"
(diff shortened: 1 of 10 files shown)


packages/sage-assets/lib/stylesheets/tokens/_index.scss
```diff
@@ -1,3 +1,4 @@
+// Variable Tokens
```
monkeypox8, line 1: comment necessary to use as an insertion point for the token generator

docs/lib/generators/sage_token/sage_token_generator.rb
```diff
@@ -3,14 +3,14 @@ class SageTokenGenerator < Rails::Generators::NamedBase
...
     style_include_line = "// Variable Tokens"
     # Create Style File
     template "style.scss", style_file
     # Include Style File
     gsub_file style_include_file, /(#{Regexp.escape(style_include_line)})/mi do |match|
-      "#{match}\n@import \"tokens/#{file_name}\";"
+      "#{match}\n@import \"#{file_name}\";"
     end
 
     # Markup Variables
```
monkeypox8, line 16: As a follow-up item, we may want to remove the lines below, since new tokens cannot be inserted in the helper until defined.

philschanely (commented):
Do we want this to also generate entries in the rails component area? the `.rb` and `.html.erb` files?

monkeypox8:
> Do we want this to also generate entries in the rails component area? the `.rb` and `.html.erb` files?

Hmm, it should already be generating those files correctly, unless I missed something?

philschanely:
@monkeypox8 perhaps its a follow-up but I wondered if we are or if not, should be, populating entries for a new thing under `docs/lib/sage_rails/app/...` etc. for the corresponding rails component?

monkeypox8:
> @monkeypox8 perhaps its a follow-up but I wondered if we are or if not, should be, populating entries for a new thing under `docs/lib/sage_rails/app/...` etc. for the corresponding rails component?

Ah, I see what you mean now. Good question, my vote would be for adding those file templates for the elements/objects. Okay for a follow up?

philschanely (approved):

PR #195, opened 2021-01-08
Fix broken import on upload card

philschanely:
## Description
Hot fix for a mistake that causes some tests to fail in `products`


dangkevin (approved):
👍🏻

PR #201, opened 2021-01-12
Feature Toggle - conditionally render aria attributes

QuintonJason:
## Description

- [x] - conditionally render aria attributes for image
- [x] - updated sageButton `attribute` to accept `nil`

### Screenshots

|  before  |  after  |
|--------|--------|
|Screen_Shot_2021-01-14_at_11_50_50_AM|Screen_Shot_2021-01-14_at_11_48_15_AM|

### Steps for testing
1. Visit rail Feature Toggle page: http://localhost:4000/pages/object/feature_toggle

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_feature_toggle.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_feature_toggle.html.erb
@@ -20,7 +20,7 @@
       <%= content_for :sage_feature_toggle_input %>
     </div>
   <% end %>
-  <a class="sage-feature-toggle__image-link">
-    <img class="sage-feature-toggle__image" src=<%= component.image %> alt="<%= component.alt_text %>"/>
+  <a class="sage-feature-toggle__image-link" <%= aria-label={component.title} if component.alt_text.blank?%> >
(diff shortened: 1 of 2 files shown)


philschanely (approved):
Love it! 🙌

docs/lib/sage_rails/app/views/sage_components/_sage_feature_toggle.html.erb
```diff
@@ -20,7 +20,7 @@
       <%= content_for :sage_feature_toggle_input %>
     </div>
   <% end %>
-  <a class="sage-feature-toggle__image-link">
-    <img class="sage-feature-toggle__image" src=<%= component.image %> alt="<%= component.alt_text %>"/>
-  </a>
+  <div class="sage-feature-toggle__image-link" >
+    <img class="sage-feature-toggle__image" src=<%= component.image %> alt="<%= component.alt_text %>" <%= "aria-hidden=true" if component.alt_text.blank? %> />
```
monkeypox8, line 24: I'm wondering about the use case of the added `aria` property in this example. If the image has a blank alt attribute /  is intended to be decorative, wouldn't it already be ignored by a screen reader?

QuintonJason: @monkeypox8  Yes, when I ran an a11y audit it kept triggering an error, so I put it in as a safe guard from false negatives

PR #208, opened 2021-01-13
Hotfix: temporarily loosen strictness to allow additional props in components

philschanely:
Hotfix: temporarily loosen strictness to allow additional props in components


AndrwM (approved):
👍 👍 👍

PR #209, opened 2021-01-13
Add choice to tabs schema

philschanely:
## Description

Since Choice component can be a subcomponent generated by `Tabs` `items` prop, the Choice fields need to be added to the validation schema. In addition, a `data` prop is added for providing additional, looser local data.


philschanely:
I'm going to move this update to my other PR related to `choice`.

PR #215, opened 2021-01-14
Patch tabs choice schema

philschanely:
## Description

This PR patches the schemas for tabs and choices to ensure they're flexible with each other. In the end, the `type` property is not necessary for choices.


pixelflips (approved):
❤️  thanks, @philschanely - LGTM!

PR #216, opened 2021-01-15
Checkbox attribute additions

monkeypox8:
## Description
Applies attributes to checkboxes when needed. The most frequent use for this is data-attributes to affect other form controls. Also adjusts logic to conditionally output `name` and `value` properties when not specified.


pixelflips (approved):
LGTM! Nice work 👍🏼

joedakroub (approved):
Nice additions Kevin. Looking good. Thanks! 👍

Issue #203, opened 2021-01-12
FOUNDATIONS: Typography


philschanely:

philschanely:
- [ ] Investigate type size in interactive controls and buttons

PR #222, opened 2021-01-17
Base font size revisions

monkeypox8:
## Description
Decreases base font size globally on larger screens to `14px`:
- Mobile font size remains unchanged (font size: `16px`)
- Font size mappings `sage-font-size()` have been modified to reference css custom properties rather than hard-coded values. This allows us to keep the current Sass mixin and variable references in elements and objects intact
- Consolidates font-face declarations and removes unused entries and files
- Updated documentation

### Design questions
- [x] ~~Which elements will need further size adjustments to match the smaller scale?~~
  - Apply scaling from previous Figma mockup (https://type-scale.com)
  - Reference: https://polaris.shopify.com/design/typography
- [x] Is `992px` the preferred breakpoint for "desktop" / the decreased base font size?

### UXD questions
- [ ] Appetite for applying custom props to other design tokens or Sass variables
- [ ] Future: should we move to a type scale multiplier instead of defining sizes? [see reference here as an example](https://www.rawkblog.com/2018/05/modular-scale-typography-with-css-variables-and-sass/). Doing so may get complicated—we make extensive use of the `rem()` function for sizing, which can't currently be altered to support a different relative base 

### Screenshots

|  before  |  after  |
|--------|--------|
|||

## Related

- #203


packages/sage-assets/lib/stylesheets/tokens/_font_size.scss
```diff
@@ -4,19 +4,41 @@
 /// @group sage
 ////
 
+:root {
```
philschanely, line 7: This tasteful use of custom props is 🥇

philschanely (approved):
Very nice. Looking forward to seeing the asset size smaller now too.

philschanely:
@monkeypox8 per your UXD questions above for posterity, here's my take:

- So if Sass provides variables and CSS provides custom properties, when would we use one or the other? This PR is a great example of tasteful use of custom props in a Sass context. I'd argue alongside this PR that using Sass variables as the default choice is best, but to leverage CSS Custom Properties for cases like this where the variability of the variables is so helpful. So here the same set of variables (font sizes) needs to be applied in a responsive fashion. While Sass variables have override abilities, the Custom Properties seem to accomplish this more elegantly IMO.
- I would not be opposed to a ratio scale for the font sizes as long as we keep the line height snapping to the baseline grid (which may or may not defy the point of a scale ratio depending on your perspective). I believe that keeping at least the baseline to rounded values helps ensure tighter adherence to design specs and better reinforces the baseline grid and thus vertical rhythm.

joedakroub (approved):
Very nice work here Kevin. Love the elegance of the custom props solution. Great cleanup of fonts too. 🥇 👍🏼

pixelflips (approved):
👍🏼  +1 - No concerns here. Awesome work and thanks for the cleanup 🧹

PR #223, opened 2021-01-19
Add React linter

philschanely:
## Description

This PR adds eslint with settings comparable to those in `kajabi-products` and lints the sage-react package. Along the way a few other things were tidied:

- removed all `default` exports of components so as to make importing them consistent within the repo
- started converting a few stories to the new format (Avatar, Button, CopyButton).


monkeypox8 (approved):
👍🏼 No concerns here, linting appears to be working as expected. Had 7 warnings, but no errors!

voodooGQ (approved):
:+1: 

https://app.circleci.com/pipelines/github/Kajabi/sage-lib/1312/workflows/b49b9e60-9471-4ffb-be1b-929072c225c9/jobs/1446

Based on the output from the job above the `yarn lint` step is completing successfully. This is the same check that runs prior to a release, so as long as it's passing here it should pass the rest of the CI checks.

PR #229, opened 2021-01-19
Modal - update docs - rails component

QuintonJason:
## Description

- [x] - add sage rails demo code and updated example 
- [x] - updated props

### Steps for testing
1. Visit rails modal page: http://localhost:4000/pages/object/modal


docs/app/views/examples/objects/modal/_preview.html.erb
```diff
@@ -1,6 +1,48 @@
 <%= sage_component SagePanelBlock, {} do %>
+  <%# Standard Modal %>
   <button class="sage-btn sage-btn--primary sage-btn--icon-right-launch" data-js-modaltrigger="cool-modal">Trigger Modal</button>
-  <%= render "examples/objects/modal/markup" %>
+  <%= sage_component SageModal, { id: "cool-modal" } do %>
```
philschanely, line 4: What do you think about having some subcomponents like `SageModalHeader`, `SageModalContent`, and `SageModalFooter` that packages up a bit more of this content? Something like this:

```erb
<%= sage_component SageModal, { id: "cool-modal" } do %>
  <%= sage_component SageModalHeader, {
    title: "Example Sage Modal",
    aside: sage_component(SageButton, ...)
  } %>
  <%= sage_component SageModalContent, {} do %>
    <!-- modal content composed here -->
  <% end %>
  <%= sage_component SageModalFooter, {} do %>
    <%= sage_component SageButton, ... %> // primary
    <%= sage_component SageButton, ... %> // others such as `aside`
  <% end %>
<% end %>
```

philschanely (commented):
Looking good overall from the cleanup angle. A comment below in regard to making this more robust.

docs/app/views/examples/objects/modal/_preview.html.erb
QuintonJason: @philschanely There are some active modals using the existing structure so I initially didn't want to introduce a potential break. Ideally the solution your proposed is what we ultimately need. My approach was purely getting a working demo to pull from. If you see no issues, I'll give it a go here

philschanely (approved):

docs/app/views/examples/objects/modal/_props.html.erb
```diff
@@ -1,3 +1,9 @@
+<tr>
```
monkeypox8, line 1: Thoughts on also documenting the `spacer`, `active`, and `disable_close` attributes to round things out?

monkeypox8 (approved):
👍🏼 Looks great! Left a comment about documenting the other available attributes, can also be a follow-up as we continue building out these docs.

docs/app/views/examples/objects/modal/_props.html.erb
QuintonJason: updated 👍

PR #236, opened 2021-01-20
Patch default avatar color

philschanely:
## Description

Quick fix ensures DEFAULT (primary blue) is default for Avatar.


pixelflips (approved):
💥  +1

PR #241, opened 2021-01-21
Fix tooltip map

philschanely:
## Description

While linting, I misunderstood where the parenthesis needed to go in this map function and unnecessarily renamed the `children` variable.


AndrwM (approved):
LGTM! +1

joedakroub (approved):

PR #244, opened 2021-01-22
Form Input - added rails component

QuintonJason:
## Description

- [x] - add rails component
- [x] - update props

## Test notes

No visual change

### Steps for testing
1. Visit form input rails page: http://localhost:4000/pages/element/form_input
2. Verify demo and docs


philschanely (approved):
🥇 This is so great! While the form components are set up in simple_form in products its is SOOOOO nice to now have this available as a pure Rails component too! Thank you thank you thank you!

PR #245, opened 2021-01-22
Form Select - added rails component and updated docs

QuintonJason:
## Description

- [x] - added rails component
- [x] - updated docs and props

## Test notes

Visit 

### Steps for testing
1. Visit rails form select page: http://localhost:4000/pages/element/form_select
2. Verify demo and docs


philschanely (approved):
Looks great and works as expected 🙌

PR #247, opened 2021-01-22
Toast adjustments to fix bootstrap styling conflicts

pixelflips:
## Description
The toast component when used within `kajabi-products` is clashing ⚔️   with some bootstrap styles related to the `output` element.

These adjustments will provide the additional properties (font-size and padding) needed to override the bootstrap defaults and prevent this issue.

### Screenshots

|  before  |  after  |
|--------|--------|
|Screen Shot 2021-01-22 at 9 00 54 AM |Screen Shot 2021-01-22 at 9 01 19 AM|

## Test notes
There will be no visual changes within Sage but see screenshots above for how the appearance will change in `kajabi-products`.

### DIFF
--- a/packages/sage-assets/lib/stylesheets/patterns/elements/_toast.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/elements/_toast.scss
@@ -74,6 +74,8 @@ $-toast-viewport-spacing-bottom: sage-spacing();
 
 .sage-toast__value {
   margin-right: sage-spacing(xs);
+  padding: 0;
   color: sage-color(white);
   line-height: sage-line-height(2xs);
+  font-size: sage-font-size(md);
 }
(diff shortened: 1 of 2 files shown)


monkeypox8 (approved):
👍🏼 LGTM, seems to work as a temporary patch. I'm not well versed with the `output` element so I'm hoping @QuintonJason can weigh in

pixelflips:
@QuintonJason when you have a couple of minutes would you mind weighing in on this one if you have any concerns?

packages/sage-assets/lib/stylesheets/patterns/elements/_toast.scss
```diff
@@ -74,6 +74,8 @@ $-toast-viewport-spacing-bottom: sage-spacing();
 
 .sage-toast__value {
   margin-right: sage-spacing(xs);
+  padding: 0;
   color: sage-color(white);
   line-height: sage-line-height(2xs);
+  font-size: sage-font-size(md);
```
philschanely, line 80: For good measure could we perhaps extend the appropriate type spec here?

pixelflips: @philschanely yessir. thanks for pointing that out! 👍🏼

PR #253, opened 2021-01-22
Update dropdown size position

philschanely:
## Description

This PR adds ability to set a `max-width` for dropdown panels and fixes a positioning bug for when dropdowns and pinned and aligned right in React.

### Screenshots

Screen Shot 2021-01-25 at 10 52 58 AM

## Test notes

See storybook and `Panel` where an `OptionsDropdown` is used in context, pinned, and positioned right now with the new `max-width`.


monkeypox8 (approved):
👍🏼 Nice! LGTM

QuintonJason (approved):
💥

PR #264, opened 2021-01-26
Dropdown - fixed string interpolation issue when multiple words present

QuintonJason:
## Description

- [x] - resolve string interpolation issue when multiple words present

### Screenshots

|  before  |  after  |
|--------|--------|
|Screen_Shot_2021-01-26_at_9_00_32_AM|Screen_Shot_2021-01-26_at_8_59_19_AM|

### Steps for testing
1. Visit dropdown sage page: http://localhost:4000/pages/object/dropdown
2. In the `SageLabel Button, with SageLabel Values, and aligned right example`, Inspect the `Delete Category` option
3. Verify that the multi word value is 2 words surrounded by quotes and not only the first word being wrapped in quotes

### DIFF
--- a/docs/app/views/examples/objects/dropdown/_preview.html.erb
+++ b/docs/app/views/examples/objects/dropdown/_preview.html.erb
@@ -97,7 +97,10 @@
       value: "Delete Category",
       icon: "trash",
       style: "danger",
-      attributes: { "href": "#" },
+      attributes: { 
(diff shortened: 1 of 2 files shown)


philschanely (approved):
Thanks for this!

joedakroub (approved):
Great catch! 👍🏼

PR #270, opened 2021-01-27
Adjust deselect functionality

philschanely:
## Description

This PR allows for `SelectDropdown`s to use the `allowDeselect` property to control whether an item within a dropdown can be deselected by clicking it a second time. This is not default behavior standard HTML `select` either, but it can now be enacted when desired in this component rather than forced as a default. 

## Test notes

No changes should occur in Rails. React uses of the `SelectDropdown` component must now set `allowDeselect` to `true` in order to retain the prior setup. No existing implementations of this component in `products` need the prior setup as a default.


joedakroub (approved):
Nice addition! Looking good. 👍🏼

PR #272, opened 2021-01-29
Panel controls select all and sorting

philschanely:
## Description

This PR updates Panel Controls React component as follows:

- Adds a Sort dropdown
- Adjusts how "select all" functions so that selections persist across page changes and matches existing table bulk selections behavior in app.

### Screenshots

Screen Shot 2021-01-29 at 2 15 02 PM

## Test notes

In `sage-lib` see Storybook > PanelControls where this dropdown is visible and actions lead to a console log.

In `products` this enables in-development features that are under feature flags. Existing builds should not be impacted by this, but new features are now enabled for handling "select all" and adding sorting connections.

### DIFF
--- a/packages/sage-react/lib/Dropdown/SelectDropdown.jsx
+++ b/packages/sage-react/lib/Dropdown/SelectDropdown.jsx
@@ -9,6 +9,7 @@ import { DropdownTriggerSelect } from './DropdownTriggerSelect';
 import { DROPDOWN_PANEL_SIZES } from './configs';
 
 export const SelectDropdown = ({
+  align,
   allowDeselect,
   allowMultiselect,
   className,
@@ -174,6 +175,7 @@ export const SelectDropdown = ({
 
   return (
     <Dropdown
+      align={align}
       className={configs.classNames}
       closePanelOnExit={closePanelOnExit}
       contained={contained}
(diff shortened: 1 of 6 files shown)


pixelflips (approved):
👍🏼  LGTM - verified locally in Storybook that dropdown is visible and actions are leading to a `console.log` as expected. Great work!

dangkevin:
The component and code look good on storybook, just some quick questions or thoughts that I just want to make sure regarding,

If I test this branch in conjunction with the master branch on `products`, should the Select All have some sort of limited functionality?

I'm not sure if this simply because we still need to wire up some additional logic or if this is a small bug. If it's the former, then I can most definitely take a look when it gets merged in. 

I'm also seeing this error coming on master. 

Screen Shot 2021-01-29 at 12 00 51 PM

philschanely:
@dangkevin yes we likely have a few tweaks to make before this will work with the setup in the people list. Happy to look at it with you once this comes over or we can prep the fixes using the bridge now if you like. Just let me know.

dangkevin:
@philschanely Sounds good, let's take a look when it comes over!

dangkevin (approved):
LGTM!

kimjaej (approved):
LGTM 👍

PR #273, opened 2021-01-29
Dropdown - conditionally render button in dropdown item

QuintonJason:
## Description

Render a `<button>` in the dropdown item if its `attributes` don't contain a `href`

### Screenshots

|  before  |  after  |
|--------|--------|
|Screen_Shot_2021-01-29_at_2_07_26_PM|Screen_Shot_2021-01-29_at_10_43_40_AM|

### Steps for testing
1. Visit the dropdown page: http://localhost:4000/pages/object/dropdown
2. If the calling code Dropdown item's `item:attributes` exists but does not contain an `href` the rendered element is a `<button>`. If the Dropdown item's `item:attributes` contain an `href` or are not present, the rendered element is a `<a>`.

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_dropdown.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_dropdown.html.erb
@@ -72,12 +72,13 @@
                 <% end if item[:attributes]&.is_a?(Hash) %>
                 role="menuitem"
                 tabindex="0"
+                type="button"
                 class="
                   sage-dropdown__item-control
                   <%= "sage-dropdown__item-control--#{item[:style]}" if item[:style] %>
                   <%= "sage-dropdown__item-control--icon-#{item[:icon]}" if item[:icon] %>
                 ">
-                hey
+                <%= item[:value].html_safe %>
               </button>
             <% end %>
           </li>
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_dropdown.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_dropdown.html.erb
@@ -50,20 +50,36 @@
             "
             role="none"
           >
-            <a
-              <% item[:attributes].each do |key, value| %>
-                <%= key %>="<%= value %>"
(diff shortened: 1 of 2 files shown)


pixelflips (approved):
LGTM! Nice work! 👍🏼

docs/lib/sage_rails/app/views/sage_components/_sage_dropdown.html.erb
```diff
@@ -50,20 +50,36 @@
...
+                  <%= key %>="<%= value %>"
+                <% end if item[:attributes]&.is_a?(Hash) %>
+                role="menuitem"
+                tabindex="0"
+                class="
+                  sage-dropdown__item-control
+                  <%= "sage-dropdown__item-control--#{item[:style]}" if item[:style] %>
+                  <%= "sage-dropdown__item-control--icon-#{item[:icon]}" if item[:icon] %>
+                ">
+                hey
```
monkeypox8, line 80: 😁 Probably should be `<%= item[:value].html_safe %>`

docs/lib/sage_rails/app/views/sage_components/_sage_dropdown.html.erb
```diff
@@ -50,20 +50,36 @@
...
+                "
+              >
+                <%= item[:value].html_safe %>
+              </a>
+            <% else %>
+              <button 
+                <% item[:attributes].each do |key, value| %>
+                  <%= key %>="<%= value %>"
+                <% end if item[:attributes]&.is_a?(Hash) %>
+                role="menuitem"
```
monkeypox8, line 73: Could we add `type="button"` here as a default?

monkeypox8 (commented):
Looking great, makes a huge difference for semantic elements. Couple of items to check out below:

docs/lib/sage_rails/app/views/sage_components/_sage_dropdown.html.erb
QuintonJason: 🤦 ,  the ole filler text strikes back!

PR #276, opened 2021-02-01
Avatar center and sizing

philschanely:
## Description

This PR adds ability to center an Avatar in its layout space and to provide a custom size for the Avatar.

Bonus: Fixes a schema validation error on `PanelRow` rails component.

### Screenshots

Screen Shot 2021-02-01 at 1 10 31 PM

## Test notes

See Objects > Avatar

### Steps for testing in app

Avatar is not used in any production contexts to validate and these changes do not affect default implementations.

Bonus: SagePanelRow examples can be checked in the following instances and they should all behave as they have to date:
- [ ] Dashboard "Here's what's happening" cards
- [ ] Product catalog > row below search bar showing number of products and sort options


pixelflips (approved):
LGTM 👍🏼  - Nice work!

PR #278, opened 2021-02-02
Sortable Component: image support

AndrwM:
BUILD-717

## Description
Adds an optional image to the Sortable component. The image sets its height and min-width to be 45px per the spec, the image width flexes to fill the available space on desktop displays.

### Screenshots

|  mobile  |  wide  |
|--------|--------|
|image|image|

## Test notes
Component location in docs: http://localhost:4000/pages/object/sortable

## Related
BUILD-717

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_sortable_item.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_sortable_item.html.erb
@@ -6,6 +6,14 @@
   data-js-sortable-update-url="<%= component.url_update %>"
   <%= "id=#{component.id}" if component.id %>
 >
+  <% if component.image.present? %>
+    <div class="sage-sortable__item-image">
+      <img
+        src="<%= component.image %>"
+        alt="Cover for <%= component.title %>"
+      />
+    </div>
+  <% end %>
   <div class="sage-sortable__item-content">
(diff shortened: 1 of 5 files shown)


packages/sage-react/lib/Sortable/SortableItem.jsx
```diff
@@ -22,6 +23,11 @@ export const SortableItem = ({
       {...rest}
       className={className}
     >
+      {image && (
+        <div className="sage-sortable__item-image">
+          <img src={image} />
```
philschanely, line 28: I think we need an empty `alt` attribute here for a11y.

docs/lib/sage_rails/app/views/sage_components/_sage_sortable_item.html.erb
```diff
@@ -6,6 +6,11 @@
   data-js-sortable-update-url="<%= component.url_update %>"
   <%= "id=#{component.id}" if component.id %>
 >
+  <% if component.image.present? %>
+    <div class="sage-sortable__item-image">
+      <img src="<%= component.image %>" />
```
philschanely, line 11: I think we need an empty `alt` attribute here for a11y.

philschanely (approved):
Two small adjustments and then 👍  from me!

AndrwM:
@philschanely updated! thanks

PR #279, opened 2021-02-02
Modal content spacing and cleanup

philschanely:
## Description

This PR does some work with Modals in Rails including:

- Adds a spacing parameter on `ModalContent` to allow for `panel` and `card` grid spacing.
- Implements `header_aside` section in `ModalContent`
- Updates documentation for modals to match

## Test notes

See Objects > Modal

### Steps for testing
Existing modals should be unaffected. Check the following to confirm:

- [ ] Product catalogue (Rails)
- [ ] Product outline (React)
- [ ] Contact import mapping step (React)

### DIFF
--- a/docs/app/views/examples/objects/modal/_props.html.erb
+++ b/docs/app/views/examples/objects/modal/_props.html.erb
@@ -40,6 +40,12 @@
   <td><%= md("`nil`") %></td>
 </tr>
 <tr>
+  <td><%= md("Section: `header_aside`") %></td>
+  <td><%= md("Populates the `sage-modal__header-aside`.") %></td>
+  <td><%= md("String") %></td>
+  <td><%= md("`nil`") %></td>
+</tr>
+<tr>
   <td><%= md("Section: `footer`") %></td>
   <td><%= md("Populates the `sage-modal__footer`.") %></td>
   <td><%= md("String") %></td>
--- a/docs/app/views/examples/objects/modal/_props.html.erb
+++ b/docs/app/views/examples/objects/modal/_props.html.erb
@@ -1,26 +1,53 @@
 <tr>
-  <td>active</td>
-  <td>Enabling this property will return the JS early to not initialize any handlers.</td>
-  <td>Boolean</td>
-  <td>null</td>
+  <td><%= md("`active`") %></td>
+  <td><%= md("Enabling this property will return the JS early to not initialize any handlers.") %></td>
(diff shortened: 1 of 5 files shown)


pixelflips (approved):
LGTM 👍🏼

packages/sage-assets/lib/stylesheets/patterns/objects/_modal.scss
```diff
@@ -84,6 +84,24 @@ $-modal-padding-y: sage-spacing(md);
   @include clearfix;
 }
 
+.sage-modal__content--spacing-panel {
+  @include sage-grid-panel();
+
+  // Undo clearfix
+  &::after {
```
monkeypox8, line 91: I can't seem to recall why the clearfix is even needed in the default `.sage-modal__content`. Was it to fix an existing issue?

docs/app/views/examples/objects/modal/_props.html.erb
```diff
@@ -1,26 +1,59 @@
...
+  <td><%= md("Boolean") %></td>
+  <td><%= md("`nil`") %></td>
 </tr>
 <tr>
-  <td>disable_close</td>
-  <td>Enabling this property will return the JS early to not initialize any handlers.</td>
-  <td>Boolean</td>
-  <td>null</td>
+  <td><%= md("`disable_close`") %></td>
+  <td><%= md("Enabling this property will return the JS early to not initialize any handlers.") %></td>
```
monkeypox8, line 9: Should this be something more along the lines of "Enabling this property will prevent the 'close' button from being rendered"?

monkeypox8 (approved):
👍🏼 LGTM, looks to be a useful addition to the modal variations. May be of particular interest to @kajabijamell since I believe he had been working on a similar layout.

docs/app/views/examples/objects/modal/_props.html.erb
philschanely: Good point. I did not inspect this but just tidied up the formatting. Will do in a follow up. i think it is actually accurate but I see the property could be misleading; one has to manually add the close button ATM and that might be nice to automate.

packages/sage-assets/lib/stylesheets/patterns/objects/_modal.scss
philschanely: i tried not to look at the blame... not sue why we even _have_ the clearfix these days 😂

PR #282, opened 2021-02-03
Sage Feature Toggle - Add `links` prop with optional arguments

joedakroub:
## Description
This PR adds the ability to pass `links` to the Sage Feature Toggle.

The `links` prop can be specified as such:

```
links: [
  {
    icon: "launch",
    location: "/",
    target: "_blank",
    text: "Learn more",
  },
  {
    icon: "comment",
    location: "/feedback",
    text: "Feedback",
  }
]
```

This work is being requested as a part of [COMM-671](https://github.com/Kajabi/kajabi-products/pull/17612).

### Screenshots
|Before|After|
|------|-----|
|Before|After|

### Steps for testing
1. Please see https://github.com/Kajabi/kajabi-products/pull/17612 for testing information.

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_feature_toggle.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_feature_toggle.html.erb
@@ -4,16 +4,28 @@
   <<%= feature_toggle_title_tag %> class="sage-feature-toggle__title"><%= component.title %></<%= feature_toggle_title_tag %>>
   <div class="sage-feature-toggle__content">
     <p><%= component.description %></p>
-    <%= sage_component SageButton, {
-      value: component.link_text,
-      style: "primary",
-      subtle: true,
-      icon: { name: "launch", style: "right" },
-      css_classes: "sage-feature-toggle__content-button",
-      attributes: component.link_location && {
-        href:  component.link_location
-      }
-    } %>
+    <div class="sage-grid-template-it">
+      <%= sage_component SageButton, {
+        value: component.link_text,
+        style: "primary",
+        subtle: true,
+        icon: { name: "launch", style: "right" },
+        css_classes: "sage-feature-toggle__content-button",
+        attributes: component.link_location && {
+          href:  component.link_location
+        }
+      } if component.link_text.present? && component.link_location.present? %>
+      <%= sage_component SageButton, {
+        value: component.secondary_link_text,
+        style: "primary",
+        subtle: true,
+        icon: { name: "comment", style: "right" },
+        css_classes: "sage-feature-toggle__content-button",
+        attributes: component.secondary_link_location && {
+          href:  component.secondary_link_location
+        }
+      } if component.secondary_link_location.present? && component.secondary_link_text.present? %>
+    </div>
   </div>
   <% if content_for?(:sage_feature_toggle_input) %>
     <div class="sage-feature-toggle__aside">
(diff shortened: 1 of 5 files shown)


docs/lib/sage_rails/app/views/sage_components/_sage_feature_toggle.html.erb
```diff
@@ -4,16 +4,28 @@
...
-      value: component.link_text,
-      style: "primary",
-      subtle: true,
-      icon: { name: "launch", style: "right" },
-      css_classes: "sage-feature-toggle__content-button",
-      attributes: component.link_location && {
-        href:  component.link_location
-      }
-    } %>
+    <div class="sage-grid-template-it">
```
philschanely, line 7: This seems to work to line these items up, but technically it never sets the container to `display: grid`. Consider a quick `sage-feature-toggle__links` rule to add that and possibly to just use flexbox?

docs/lib/sage_rails/app/views/sage_components/_sage_feature_toggle.html.erb
```diff
@@ -4,16 +4,28 @@
...
+        css_classes: "sage-feature-toggle__content-button",
+        attributes: component.link_location && {
+          href:  component.link_location
+        }
+      } if component.link_text.present? && component.link_location.present? %>
+      <%= sage_component SageButton, {
+        value: component.secondary_link_text,
+        style: "primary",
+        subtle: true,
+        icon: { name: "comment", style: "right" },
```
philschanely, line 22: Will this always be a comment? Any value to providing `secondary_link_icon` an override that defaults to `comment`? 

More broadly I was wondering about switching to a `links` array that allows for you to pass in one or more objects that match the schema for `SageButton` or at least the necessary subset, value, href, etc. I'm hesitant that `secondary_link` could imply that this is styled as a secondary button.

docs/lib/sage_rails/app/views/sage_components/_sage_feature_toggle.html.erb
philschanely: (this class is ideally meant to be used on like SageCardRow or SagePanelRow which each lay the foundation for grid and spacing needed)

philschanely (commented):
A few notes/suggestions here before approving. Super close though.

philschanely:
@joedakroub it would be great to update the `props` file to match your settings and possibly also show another example in `preview` of the additional option you've added.

docs/lib/sage_rails/app/views/sage_components/_sage_feature_toggle.html.erb
joedakroub: @philschanely All great questions. In usage of this component, the "comment" icon is use in a Feedback link (to provided feedback on the beta feature). These links are all optional as far as using this component to display beta features.

As far as a `links` array, I am certainly happy to switch to that approach. I struggled a bit with this component in general because it seems a bit unicorn, but after talk to @monkeypox8 about this last week, we did see where this component could be beneficial for Upsells within Offers and possibly other places.

I too am hesitant about adding a `secondary_link`, so I will adjust this work to use a `links` array instead. Great call!

philschanely: Yeah good call. 

IMO it either needs to afford true customization as we've discussed above or be more locked down to the intended unicorn situation, such as fixing in the labels "Feedback" and "Learn more" for the links and only allowing the particular URLs to be passed in like `feedback_url` and `info_url`. If you're pressed for time that may be a shorter solution ATM. Just thinking out loud.

monkeypox8 (commented):
Looking great! Left a note to confirm that we're bypassing buttons in the link section. If so, we're good to go; adding the `target` attribute isn't a blocker at this time.

monkeypox8 (approved):
👍🏼  Beautifully done!

philschanely (approved):
😍  🏆

PR #283, opened 2021-02-03
Fix table unselect one from all

philschanely:
## Description

This PR fixes an issue with handlining selections: When you select all and then deslect one item the selection should switch to all on the current page except the one you just deselected.

## Test notes
N/A


joedakroub (approved):
Nice solution to this issue Phil. No concerns with this approach. 👍🏼

dangkevin (approved):
Good work Phil, looks good on local!

PR #284, opened 2021-02-03
Dropdown - trigger label will contain recently selected option text onload

QuintonJason:
## Description

- [x] - add `selected` property to the `items` array
- [x] - updated trigger label to contain `selected` text on pageload

### Screenshots

|  before  |  after  |
|--------|--------|
|Screen Shot 2021-02-03 at 3 08 53 PM|Screen_Shot_2021-02-03_at_3_05_17_PM|

### Steps for testing
1. Pending

## Related

-

### DIFF
--- a/packages/sage-system/lib/dropdown.js
+++ b/packages/sage-system/lib/dropdown.js
@@ -27,18 +27,16 @@ Sage.dropdown = (function() {
   // Several variations exist for triggers in selection mode; this is a grouped selector to grab any of them
   const triggerSelectClasses = '.sage-dropdown__trigger--select-labeled, .sage-dropdown__trigger--select';
 
-  const dropdownItemAriaSelectedClass = '.sage-dropdown__panel .sage-dropdown__item[aria-selected="true"]';
-  const dropdownItemControlClass = '.sage-dropdown__item-control';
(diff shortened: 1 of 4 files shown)


packages/sage-system/lib/dropdown.js
```diff
@@ -138,6 +144,16 @@ Sage.dropdown = (function() {
     el.setAttribute('aria-expanded', false);
   }
 
+  function checkSelected(el) {
+    let selectedItem = el.querySelector(dropdownItemAriaSelectedClass); 
+    
+    if (selectedItem != null) {
+      let selectedItemValue = selectedItem.querySelector(dropdownItemControlClass).innerHTML;
```
joedakroub, line 151: Question: when the dropdown is used as a `select`, are there cases where we use HTML in the value or is it always just text? If so, `textContent` may be a better option here than using `innerHTML` due to the parsing that it has to do. Minor quip and may not be relevant in this case, but worth mentioning.

Good write-up about it here: https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innerhtml

joedakroub (approved):
This is great and much-needed. I added one minor comment, but not a blocker. 👍🏼

PR #285, opened 2021-02-03
Webpack config changes for Sage CSS/Fonts

joedakroub:
## Description
This PR is an attempt at moving the Sage fonts out of the stylesheet and what impact that might have. This came out of discussions that were taking place in `dev-js` about how to reduce Sage CSS.

There are probably better/alternative ways to do this, but this seemed the simplest to get things started.

### ! UPDATE !
@voodooGQ was able to update the webpack config to fix the issue with font exporting. Please pull down this branch and test with `kajabi-products` AND Sage Docs.

### Docs
https://webpack.js.org/loaders/url-loader/
https://webpack.js.org/loaders/file-loader/

### Screenshots
I've attached some shots to see the before/after of the stylesheet/font sizes within Kajabi Products:

| |Before|After|
|-|------|------|
|**Stylesheeet**|Before - Stylesheet|After - Stylesheet|
|**Fonts**|Before - Fonts|After - Fonts|

### Testing
1. Checkout `jed_sage_assets_webpack_config`
2. Run `yarn build`
3. Once the assets are compiled, run Kajabi Products AND Sage docs to ensure that the fonts are appearing as they should
4. Check the `Network` tab in your browser and ensure that you are seeing reduced payloads outlined in the After screenshots

### DIFF
--- a/docs/config/webpack/environment.js
+++ b/docs/config/webpack/environment.js
@@ -1,3 +1,7 @@
 const { environment } = require('@rails/webpacker')
 
+environment.loaders.get("sass").use.splice(-1, 0, {
+  loader: "resolve-url-loader"
+});
+
 module.exports = environment
--- a/packages/sage-assets/lib/stylesheets/core/_fonts.scss
+++ b/packages/sage-assets/lib/stylesheets/core/_fonts.scss
@@ -4,7 +4,6 @@
 /// @group sage
 ////
 
-
 $-body-font-path: "../fonts/inter"; // pathname of font directory
 $-body-font-version: 3.15;
 
--- a/packages/sage-assets/lib/stylesheets/core/_variables.scss
+++ b/packages/sage-assets/lib/stylesheets/core/_variables.scss
@@ -4,7 +4,6 @@
 /// @group sage
 ////
 
-
 ///
 /// Active underline configuration
 ///
--- a/packages/sage-assets/webpack/webpack.prod.js
+++ b/packages/sage-assets/webpack/webpack.prod.js
@@ -35,10 +35,7 @@ module.exports = {
       },
       {
         test: /\.(woff|woff2|eot|ttf)$/,
-        loader: "url-loader",
-        options: {
-          limit: 0,
-        },
+        loader: "file-loader",
       },
       {
         test: /\.(png|svg|jpg|jpeg|gif)$/,
--- a/packages/sage-assets/webpack/webpack.prod.js
+++ b/packages/sage-assets/webpack/webpack.prod.js
@@ -25,7 +25,7 @@ module.exports = {
             loader: MiniCssExtractPlugin.loader,
             options: {
               sourceMap: true,
-              sourceMapContents: false,
+              sourceMapContents: false
             },
           },
           "css-loader",
@@ -36,6 +36,10 @@ module.exports = {
       {
         test: /\.(woff|woff2|eot|ttf)$/,
         loader: "file-loader",
+        options: {
+          outputPath: './fonts',
+          publicPath: './fonts'
+        }
       },
       {
         test: /\.(png|svg|jpg|jpeg|gif)$/,
--- a/packages/sage-assets/webpack/webpack.prod.js
+++ b/packages/sage-assets/webpack/webpack.prod.js
@@ -37,7 +37,7 @@ module.exports = {
         test: /\.(woff|woff2|eot|ttf)$/,
         loader: "url-loader",
         options: {
-          limit: Infinity,
+          limit: 0,
         },
       },
       {


monkeypox8:
Due to my limited expertise with webpack, I'll defer to those with greater knowledge of the subject, but this does seem to have removed the fonts from being inlined in the CSS. One small drawback that I'm able to see is a slightly delayed (~500ms) load time, but I think it's more than acceptable as a tradeoff for file size. Otherwise, this appears to be working as expected!

voodooGQ:
I'm by no means an expert here either, however, from what I'm reading on the url-loader documentation when the file size is equal or above the `limit` setting it defaults to `file-loader` instead of `url-loader`. If that's the case we probably should just be using `file-loader` as the default since setting the `limit` to `0` will presumably just default back.

I remember there being some problems with paths and things of the like when I initially tried to setup things with `file-loader`. I found this article which goes into what I think the problem was in depth and how to address it: https://chriscourses.com/blog/loading-fonts-webpack.

The other thing I'd mention is that if we use `file-loader` there's probably going to have to be some adjustments to how the fonts are loaded via the `packs` module. That's just a guess at this point but it seems likely.

philschanely (approved):
🙌

cjbuchmann:
Quick Q, the stylesheet before and after looks dope! I can't see the fonts before very well. Or maybe I'm missing something?

cjbuchmann (approved):
👍  seems like a great start, nice job! Would be nice to test this on a branch with a before/after so that we could show any improvements in build time!

joedakroub:
@cjbuchmann Thanks so much for taking a look. I wasn't able to capture the `before` size of the fonts very well because they are inside the stylesheet in this case.

> 👍 seems like a great start, nice job! Would be nice to test this on a branch with a before/after so that we could show any improvements in build time!

That would be great, how would we go about doing this?

joedakroub:
@voodooGQ I changed the config to use `file-loader`. While this seems to work locally from what I've tested and I believe @monkeypox8 did as well, you mentioned adjustments for font loading via the `packs` module. Do you have any insight into what changes would be needed?

voodooGQ:
@joedakroub I can't remember the exact problem but I believe it had to do with the font path being different for when you import the .scss files directly vs. when you call to them from the built .css file. 

Mainly I'd make sure that it works in both scenarios, when you pull in the *.scss file for variables and such and when you're pulling in the .css file.

Let me know if you want to jump on a call to talk it out a bit.

cjbuchmann:
> That would be great, how would we go about doing this?

@joedakroub so what I might do is create two branches `master_test` and `sage_optimized_test`, where master is just the current version of master, but that way you can essentially freeze it so that only the changes your comparing are your own.

In `sage_optimized_test`, just update the gem dependency to point to this branch and bundle it up. Then just deploy it to any qa box and see how the `master_test` timing compares to the `sage_optimized_test`. OR something akin to that.

cjbuchmann:
Hey all, wanted to get a pulse on this and see how it's going?

joedakroub:
I've been away from this for a moment, but should be able to circle back on this tomorrow.

joedakroub:
@voodooGQ  Circling back on this now. I spent some time trying to address the SCSS issue that you mentioned and was not able to figure it out based on the article you referenced. I see that it is a path issue with how the packs work, but am clearly missing something. Would you have some time next week that we could pair on this?

voodooGQ:
@joedakroub sure thing, I'll reach out on Slack

docs/config/webpack/environment.js
```diff
@@ -1,3 +1,7 @@
 const { environment } = require('@rails/webpacker')
 
+environment.loaders.get("sass").use.splice(-1, 0, {
```
voodooGQ, line 3: https://github.com/rails/webpacker/issues/1476

monkeypox8 (approved):
👍🏼  🔥  LGTM, great work and thanks to all involved! Compiled CSS size has been dramatically reduced from what I can see. Looks to be around a 70-75% drop uncompressed. Fonts are an additional 300-400k, but my understanding is that browsers are able to aggressively cache fonts after the initial download:

|  Before   |  After  |
|--------|--------|
|prod-sage-css|local-sage-css|

kimjaej (approved):
This looks good to me. 👍

cjbuchmann:
Can we do this with a version bump so that we can test the before/after timing on some environment (qa or staging) and record the difference?

joedakroub:
@philschanely When you are able, would you be able to check these changes out to see if they are good on your end? If so, then we can get this into the next version bump per @cjbuchmann comment above.

philschanely:
confirming all looks good and the bridge works fine for me!

philschanely (approved):
🙌

PR #290, opened 2021-02-05
fix(react textarea): fix cursor jumping issue within React <Textarea />

AndrwM:
## Description
The Sage React Textarea was managing its own internal state `onChange` rather than wholesale accepting
the `onChange` prop being passed in from its parent component. This was causing the cursor to to jump to the end of the content when attempting to edit. This now matches the implementation for Sage React Input. IIRC at one point we had this same issue with the React Inputs, the Textarea's were likely overlooked when the bug was fixed.

See the QA video in https://kajabi.atlassian.net/browse/BUILD-627 for more info

### Screenshots

image

## Test notes
When the Textarea does not have a self-managed state, ensure that the Textarea cursor does not jump to the end when editing the value.

This issue was originally discovered here: `https://www.kajabi.test:3000/admin/products/:product_id`,
...with the `:sage_product_admin_ui` feature flag flipped,
...and the Category edit modal open

## Related
BUILD-627

### DIFF
--- a/packages/sage-react/lib/Textarea/Textarea.jsx
+++ b/packages/sage-react/lib/Textarea/Textarea.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
@@ -12,7 +12,7 @@ export const Textarea = ({
   value,
   ...rest
 }) => {
-  const [fieldValue, updateFieldValue] = useState(value);
+  const [fieldValue, updateFieldValue] = useState(null);
   const classNames = classnames(
     'sage-textarea',
     className,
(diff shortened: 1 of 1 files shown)


packages/sage-react/lib/Textarea/Textarea.jsx
```diff
@@ -29,18 +29,15 @@ export const Textarea = ({
     }
   };
 
-  useEffect(() => {
-    updateFieldValue(value);
-  }, [value]);
-
   return (
     <div className={classNames}>
+    yooo
```
philschanely, line 34: meeeee :-D

philschanely (approved):
Thanks for this! 🙌  one small fix below ;)

packages/sage-react/lib/Textarea/Textarea.jsx
AndrwM: hahaha thanks

PR #292, opened 2021-02-08
Storybook Deployment adjustments

voodooGQ:
Adds a `heroku-prebuild` script to the `sage-react:storybook` build. This will install react/react-dom/react-router-dom in lieu of adding to the `devDependencies`. When adding to the `devDependencies` an error occurs due to multiple versions of React. This should address the failing deploy without impacting other areas of the application.

## Testing

I'm not as familiar with the ins-outs of working through Sage on a daily basis, so I'm hoping the reviewer can test this for me. Ideally you should be able to download, run the local instance of Sage without error, as well as create a bridge with Kajabi Products without error.

## Notes

Due to the way Heroku handles installation of engines `yarn` is not available during the `heroku-prebuild` script. Using `npm` instead to install the dependencies.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
       - node/with-cache:
           steps:
             - run: yarn docs:deploy
-            #- run: yarn storybook:deploy
+            - run: yarn storybook:deploy
             - run: yarn sassdocs:deploy
 workflows:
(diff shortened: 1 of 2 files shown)


philschanely (approved):
👍  Tested with the bridge and all is well! 🙌

PR #294, opened 2021-02-09
BUILD-730 - form element error states with parsley

QuintonJason:
## Description

[BUILD-730](https://kajabi.atlassian.net/browse/BUILD-730)
- [x] - update parsley errors to mimic the design patterns that currently exist in form elements

### Screenshots

|  before  |  after  |
|--------|--------|
|Screen Shot 2021-02-09 at 4 16 40 PM|Screen Shot 2021-02-09 at 4 16 09 PM|

#### With the link on...
1. Create a new podcast in the app, visit here: http://www.kajabi.test:3000/admin/sites/1/podcasts/new
2. Without populating any of the form fields, click the `Save` button
3. Verify that parsley's css error state is overwritten.

## Related

- [BUILD-730](https://kajabi.atlassian.net/browse/BUILD-730)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/index.scss
+++ b/packages/sage-assets/lib/stylesheets/index.scss
@@ -89,3 +89,4 @@
 
 // Vendor Modifications
 @import "vendor/tinymcev4";
+@import "vendor/parsley";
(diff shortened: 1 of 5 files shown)


QuintonJason:
One issue after parsley is fired is that the first `<option>` which is `disabled` still gets selected making the `<select>` undesired
Screen Shot 2021-02-09 at 5 23 32 PM

philschanely (approved):
All good and make sense. What do you think about consolidating these all under the `stylesheets/vendor` in some fashion? I can see pros and cons so no strong feelings either but just a thought.

QuintonJason:
> All good and make sense. What do you think about consolidating these all under the `stylesheets/vendor` in some fashion? I can see pros and cons so no strong feelings either but just a thought.

Good callout. I can swap the location

PR #297, opened 2021-02-10
[MAN-1065] Expandable Card Component Updates

pixelflips:
## Description
https://kajabi.atlassian.net/browse/MAN-1065

Updates and refactors Expandable Card to meet updated design spec.

- Moves trigger outside of the component body.
- Adds option for the body to include a background and border
- Adds option for the body to NOT include background and border.

### Screenshots

|  before  |  after  |
|--------|--------|
|Screen Shot 2021-02-10 at 10 41 58 AM|Screen Shot 2021-02-10 at 10 45 30 AM|

## Test notes

(BREAKING) Changes to ExpandableCard require a patch in kajabi-products to adjust the Cloudflare directions within the custom domain flow. PR [#17780](https://github.com/Kajabi/kajabi-products/pull/17780) contains fixes.

### Steps for testing

- Visit http://localhost:4000/pages/breakout/object/expandable_card
- Verify UI and docs

--

- Visit http://localhost:6006/?path=/story/sage-expandablecard--default
- Verify UI and props
- Verify knobs work as expected to add border/background

## Related

- NA

### DIFF
--- a/packages/sage-assets/lib/stylesheets/patterns/objects/_expandable_card.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/objects/_expandable_card.scss
@@ -54,7 +54,7 @@ $-expandable-card-padding-xs: sage-spacing(xs);
     justify-content: center;
     margin-right: rem(4px);
     font-size: rem(4px);
-    transition: all 0.2s linear;
+    transition: transform 0.2s linear;
   }
   &[aria-expanded="true"]::before {
     transform: rotateZ(90deg);
--- a/packages/sage-assets/lib/stylesheets/patterns/objects/_expandable_card.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/objects/_expandable_card.scss
@@ -11,7 +11,7 @@ $-expandable-card-padding: sage-spacing(sm);
 $-expandable-card-padding-xs: sage-spacing(xs);
 
 .sage-expandable-card__body,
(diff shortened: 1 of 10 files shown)


philschanely (approved):
Looking good here! 🙌

packages/sage-assets/lib/stylesheets/patterns/objects/_expandable_card.scss
```diff
@@ -4,113 +4,59 @@
...
-    transform-origin: center;
-    margin: 0 0 0 auto;
-    color: sage-color(charcoal, 100);
-    transition: $-expandable-card-transition transform;
+.sage-expandable-card__trigger {
+  &::before {
+    justify-content: center;
+    margin-right: rem(4px);
+    font-size: rem(4px);
+    transition: all 0.2s linear;
```
monkeypox8, line 57: I think it's okay here, but consider using the specific property here (`transform`) rather than `transition: all`, as it will cause performance issues once additional properties get applied

monkeypox8 (commented):
This is looking really great! React version's 🔥, though I did find that the Rails component doesn't appear to toggle using keyboard controls. Not sure if it's a requirement for the current use case; could be applied in a follow-up.

monkeypox8 (approved):
👍🏼 LGTM, I think we should be good to fix the keyboard control in a fast follow. Great work!

PR #301, opened 2021-02-11
Typeahead enhancements

philschanely:
## Description

This PR adds a few small enhancements to Typeahead:

- Allows for a searchFn callback to override local items for the case where a type ahead's items should come from an API
- Allows for `contained` setting to be passed through to the inner Search (`true`) remains default.

### Screenshots

No visual changes

## Test notes

No changes to functionality to date.

### Steps for testing
1. [ ] (Low) Adds features to Typeahead. Only existing use of this component is in the search header on Product Outline manager. Ensure this still works as expected.

### DIFF
--- a/packages/sage-react/lib/Typeahead/Typeahead.jsx
+++ b/packages/sage-react/lib/Typeahead/Typeahead.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useRef } from 'react';
+import React, { useState, useRef, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import uuid from 'react-uuid';
 import { Search } from '../Search';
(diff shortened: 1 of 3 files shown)


packages/sage-react/lib/Typeahead/Typeahead.jsx
```diff
@@ -27,13 +29,17 @@ export const Typeahead = ({
...
+    const _value = searchValue.toUpperCase();
+    let _results = [];
+    if (searchFn) {
+      _results = await searchFn(_value);
+    } else {
+      items.forEach((item) => {
+        if (_results.length >= maxResults) return;
+        if (item.title.toUpperCase().includes(_value)) _results.push(item);
+      });
+    }
```
AndrwM, line 42: Very nice! I love it.

1 notes here – If the `searchFn` prop is used the component becomes no longer backed by the `items` array. Could we argument check and throw an error telling the developer that the component only supports `items` or `searchFn`?

https://github.com/Kajabi/sage-lib/blob/a16038af45688e3e4dd0aed7d69dc44c57660f26/packages/sage-react/lib/common/hooks/useFocusTrap.jsx#L22-L29

joedakroub: Good point @AndrwM. Could the `searchFn` push into the `items` array so that the component is still backed by `items`?

joedakroub (approved):
This is a nice improvement. Added one comment based on what Andrew mentioned, but not a blocker for me. 👍🏼

PR #303, opened 2021-02-12
Pagination - add option to hide page count

monkeypox8:
## Description
This update to the Sage Pagination component is in response to performance issues we've encountered when calculating the item count. For a site with large numbers of records in its database, this can end up causing a long lookup time, eventually resulting in a timeout and making the view inaccessible.

**The impact on current use should be minimal since this is an optional attribute, but views listed below should still be checked.**

Changes:
- attribute `hide_counter` has been added to prevent the counter from rendering
- modified CSS to force pagination link alignment when the counter is disabled
- added documentation around `hide_pages` and `hide_counter`

### Screenshots

pagination-counter

### Change impact: low
Affected locations (listed by view)
- [x] Coupons index: `app/views/admin/coupons/_new_coupons_list.html.erb`
- [x] Offer list:
  - [x] `app/views/admin/offers/_table.html.erb`
  - [x] `app/views/admin/offers/index.html.erb`
- [x] Payment transactions: `app/views/admin/payments/_transactions.html.erb`
- [ ] Podcasts: `app/views/admin/podcasts/show.html.erb`
- [ ] Product Offers: `app/views/admin/product_offers/index.html.erb`
- [x] Products: `app/views/admin/products/index.html.erb`
- [ ] SuperAdmin:
  - [ ] Search: `app/views/super_admin/shared/_dropdown_search.html.erb`
  - [ ] Users: `app/views/super_admin/users/index.html.erb`

### DIFF
--- a/docs/app/views/examples/objects/pagination/_markup.html.erb
+++ b/docs/app/views/examples/objects/pagination/_markup.html.erb
@@ -1,5 +1,8 @@
-<nav class="sage-pagination" aria-label="pagination">
-  <span class="sage-pagination__count"><strong><%= total_count %></strong> Records</span>
+<% hide_counter ||= false %>
+<nav class="sage-pagination<% if hide_counter %> sage-pagination--no-counter<% end %>" aria-label="pagination">
+  <% unless hide_counter %>
+    <span class="sage-pagination__count"><strong><%= total_count %></strong> Records</span>
+  <% end %>
   <ul class="sage-pagination__pages">
     <% pagination_items.each do |item| %>
       <li class="sage-pagination__item">
--- a/docs/app/views/examples/objects/pagination/_props.html.erb
+++ b/docs/app/views/examples/objects/pagination/_props.html.erb
@@ -1,4 +1,16 @@
 <tr>
+  <td><code>hide_pages</code></td>
+  <td>When set to <code>true</code>, links to individual pages will not be rendered, leaving only the "Back" and "Next" links visible.</td>
+  <td>Boolean</td>
+  <td>false</td>
+</tr>
+<tr>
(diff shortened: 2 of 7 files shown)


joedakroub (approved):
This works well locally. Very nice addition! Thanks Kevin! 👍🏼

docs/app/views/examples/objects/pagination/_markup.html.erb
```diff
@@ -1,5 +1,8 @@
-<nav class="sage-pagination" aria-label="pagination">
-  <span class="sage-pagination__count"><strong><%= total_count %></strong> Records</span>
+<% hide_counter ||= false %>
```
philschanely, line 1: Is this needed anymore?

docs/app/views/examples/objects/pagination/_props.html.erb
```diff
@@ -1,3 +1,15 @@
+<tr>
```
philschanely, line 1: Any chance you'd be willing to sync this up with the ruby component props? and note the `md()` rails function for faster formatting. :-D

philschanely (approved):
Looks great. A few questions/suggestions below but not blocking.

docs/app/views/examples/objects/pagination/_markup.html.erb
monkeypox8: Well, it does seem to be for the example in the docs, yes. My understanding of it is that the variable needs to be initialized before it can be called, but since we're using markup and not the Rails component, that step gets bypassed here. I had hoped to replace this markup with the Rails component, but figuring out how to format/manipulate the activerecord request is beyond my knowledge.

PR #305, opened 2021-02-12
Feature Toggle - image sizing to not stretch

QuintonJason:
## Description

- [x] - resolve image stretching issue

### Screenshots

|  before  |  after  |
|--------|--------|
|Screen Shot 2021-02-12 at 7 47 14 AM|Screen Shot 2021-02-12 at 7 42 38 AM|

### Steps for testing
1. In the app visit the Beta Features page: http://www.kajabi.test:3000/admin/sites/1/beta_features
2. Verify that the long copy doesn't cause the image to stretch

### DIFF
--- a/packages/sage-assets/lib/stylesheets/patterns/objects/_feature_toggle.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/objects/_feature_toggle.scss
@@ -85,17 +85,17 @@ $-image-height-mobile: rem(120px);
   display: flex;
   position: relative;
   grid-area: img;
+
+  @media (min-width: sage-breakpoint(md-min)) {
+    align-self: flex-start;
+  }
 }
 
 .sage-feature-toggle__image {
   align-self: flex-start;
   max-width: rem(140px);
-  width: auto;
-  height: 100%;
+  width: $-image-width;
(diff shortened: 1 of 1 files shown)


packages/sage-assets/lib/stylesheets/patterns/objects/_feature_toggle.scss
```diff
@@ -85,17 +85,17 @@ $-image-height-mobile: rem(120px);
...
+  }
 }
 
 .sage-feature-toggle__image {
   align-self: flex-start;
   max-width: rem(140px);
-  width: auto;
-  height: 100%;
+  width: $-image-width;
+  height: $-image-width;
```
philschanely, line 98: Is it indeed the case that we want to keep a square always? For example, what if instead you had a custom property like `--image-aspect-ratio` that defaults to 1 and is used to calculate the height? It could even be a property passed in and conditionally added inline? Just thinking out loud.

philschanely (approved):
Makes sense! One 'food for thought' non-blocking question but 👍  for sure.

joedakroub (approved):
No issues with these changes. @philschanely brings up a good question, but not a blocker. 👍🏼

PR #310, opened 2021-02-12
accordian-panel.js clean up/keyboard support updates.

pixelflips:
## Description
This is a follow up to the new expandable card adjustments #297 and includes the following adjustments:

- Removes code related to the `group-toggle-solo` check. 
This will no longer be needed with the latest updates to the expandable card.
- Improves keyboard support for the rails component.
`enter` and `space` keys should now expand and collapse the body properly.
If multiple expandable cards appear on the page, tabbing will navigate to each trigger if the body is not expanded.
If the body is expanded the child elements will receive focus/navigation as expected

Keyboard navigation updates based on:
 https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html

### Steps for testing
1. Visit http://localhost:4000/pages/breakout/object/expandable_card
2. Tab to the expandable card trigger (Expand)
3. Verify pressing `space` and `enter` expand/collapse the body.

--

1. Visit http://localhost:4000/pages/breakout/object/expandable_card
2. Collapse all cards
3. Verify pressing `tab` &  `shift+tab` navigates between triggers
Previously hidden children elements would receive focus even if not visible.

## Related

- N/A

### DIFF
--- a/packages/sage-assets/lib/stylesheets/patterns/objects/_expandable_card.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/objects/_expandable_card.scss
@@ -53,7 +53,7 @@ $-expandable-card-padding-xs: sage-spacing(xs);
   &::before {
     justify-content: center;
     margin-right: rem(4px);
-    font-size: rem(4px);
+    font-size: rem(8px);
     transition: transform 0.2s linear;
   }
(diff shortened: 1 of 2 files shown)


monkeypox8 (approved):
👍🏼 LGTM! Nicely done, this works great!

PR #317, opened 2021-02-17
Lists Documentation Updates

pixelflips:
## Description
Updates documentation and converts Lists to Sage component.

### Screenshots
No visual changes with these adjustments.

### Steps for testing
1. Visit http://localhost:4000/pages/breakout/element/lists
2. Verify UI and documentation updates.

## Related

- N/A


philschanely (approved):
This is progress for sure. We may want to eventually return to this and incorporate more into the component: modifiers, list generation, wrapping element, etc. The documentation may then be better synced. Ok for now!

PR #318, opened 2021-02-17
Sage form input property additions

monkeypox8:
## Description
This branch came about due to work in COMM-629, needing a `number` field to increment and preventing negative values.

- Updates Sage form input to allow use of `autocomplete`, `min`, `readonly`, and `step` values
- Documentation revised to match
- Fixes an issue with `value`, where setting a string with special characters like apostrophes or spaces would truncate prematurely

## Test notes
(MEDIUM) No known uses of `input_type: "number"` in the app at present. However, the change in output of `value` should be monitored. Some examples include:
- [x] Offer detail view:
  - [x] Affiliate settings (After Purchase panel)
  - [x] Days of Access (Product Access panel)
- [x] New coupon setup view (`app/views/admin/coupons/_form.html.erb`)

## Related
- COMM-629

Pending merge conflict from #323

### DIFF
--- a/docs/app/views/examples/elements/form_input/_preview.html.erb
+++ b/docs/app/views/examples/elements/form_input/_preview.html.erb
@@ -52,6 +52,21 @@
     has_placeholder: false
   } %>
 
+  <h3 class="t-sage-heading-6">Text (readonly)</h3>
+  <%= sage_component SageFormInput, {
+    id: "form__readonly",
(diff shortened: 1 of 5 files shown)


joedakroub (approved):
Tested it locally, and the new attributes work as expected. Nice updates to the docs as well. Thanks Kevin! 👍🏼

philschanely (approved):
No concerns here.

PR #321, opened 2021-02-17
Elements/Objects Props Updates

pixelflips:
## Description
This PR updates all the `_props.html.erb` files with the following changes to better standardize these files:

- Adds props documentation to elements/objects where missing.
- Converts all outstanding objects and elements to use markdown.
- Updates all files to use `nil` instead of `null` where applicable.
- Updates all files to use `'` for wrapping `md` content unless inner contents need `"` or `%(...)` as a last resort.

### Screenshots
No visual changes with these updates.

### Steps for testing
1. Visit any object or element breakout page.
2. Verify properties display and `nil` is used instead of `null`

## Related

- N/A


philschanely (approved):
🙇‍♂️ 🙇‍♂️ 🙇‍♂️

PR #324, opened 2021-02-18
Fix assistant image path

philschanely:
## Description

This PR updates the Assistant rails component to fix an issue with the logo image path and calling method. It includes a temporary shim/default value specifically for `kajabi-products` in order to avoid introducing a breaking change that would require a follow up in `kajabi-products`.

### Screenshots

N/A 

## Test notes

See Objects > Assistant and usage at top of Documentation site

### Steps for testing
1. SageAssistant component is updated. It is used exclusively in Super Admin at this time. 
    - [ ] Log in to `super_admin` and confirm the masthead appears as expected.


pixelflips (approved):
👍🏼  LGTM!! Thanks for the future-proofing as well.

PR #329, opened 2021-02-22
fix(rails empty state): collapse Rails EmptyState title when no title provided, matches React

AndrwM:
BUILD-619

## Description
Matches react implementation:
- Removes empty SageEmptyState title tag when no title defined
- Removes empty SageEmptyState text tag when no text defined
- Updates title tag to be an `h2` (matching React component)

### Screenshots

|  before  |  after  |
|--------|--------|
|image|image|

## Test notes
Ensure title & text are collapsing

## Related
BUILD-619

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_empty_state.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_empty_state.html.erb
@@ -11,7 +11,7 @@
     </div>
   <% end %>
   <%= "<i class='sage-empty-state__icon sage-icon-#{component.icon}-3xl'></i>".html_safe if component.icon %>
-  <h1 class="sage-empty-state__title"><%= component.title %></h1>
-  <p class="sage-empty-state__text"><%= component.text %></p>
+  <%= "<h2 class='sage-empty-state__title'>#{component.title}</h2>".html_safe if component.title %>
+  <%= "<p class='sage-empty-state__text'>#{component.text}</p>".html_safe if component.text %>
   <%= component.content %>
 </section>


joedakroub (approved):
This change looks good to me. Wondering if a conditional should be applied to the `component.text` as well?

AndrwM:
@joedakroub Ah, good call, adding 🙏

QuintonJason (approved):

docs/lib/sage_rails/app/views/sage_components/_sage_empty_state.html.erb
```diff
@@ -11,7 +11,7 @@
     </div>
   <% end %>
   <%= "<i class='sage-empty-state__icon sage-icon-#{component.icon}-3xl'></i>".html_safe if component.icon %>
-  <h1 class="sage-empty-state__title"><%= component.title %></h1>
-  <p class="sage-empty-state__text"><%= component.text %></p>
+  <%= "<h2 class='sage-empty-state__title'>#{component.title}</h2>".html_safe if component.title %>
```
philschanely, line 14: @AndrwM for curiosity's sake is there any difference in rails (interpolation, processing, etc.?) between using `'`, `"` or `%(...)`? In a case like this as an example, could we have done this with no particular side effect/cost?
```erb
<%= %(<h2 class="sage-empty-state__title">#{component.title}</h2>).html_safe if component.title %>
```

AndrwM: @philschanely Oh hmm! I'm not sure, from the ruby docs it doesn't seem like there is a difference. Using ruby percent strings would be a great idea here so that we don't need to use single quotes here.

PR #332, opened 2021-02-22
Adds gap for sage choice component group

kajabijamell:
### Screenshots

**Before**
image
**After**
image

## Test notes
Testing this fix is solely visual, you should be ablle to wrap 2 or more sage choice components inside of a btn-group and there should be a visible gap when adding `--gap` modifiers.

### Steps for testing
1. Add two or more sage choice components inside of a `btn-group` wrapper component
2. Apply a `--gap` modifier to the `btn-group` with a value of `xs`, `sm`, `md` or `lg`
3. Preview the page and you should see a gap with the corresponding size.

### DIFF
--- a/packages/sage-assets/lib/stylesheets/patterns/elements/_choice.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/elements/_choice.scss
@@ -5,6 +5,8 @@
 
 ================================================== */
 
+/* stylelint-disable max-nesting-depth */
+
 $item-radio: "before";
 $item-icon: "before";
 $item-arrow: "before";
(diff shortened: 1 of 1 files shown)


philschanely (approved):
No concerns here. Thanks for scoping this to `sage-btn-group`; great consideration for other uses for these.

In case it is relevant for your use, these can also be used inside `SageTabs` with the `style: "choice"` set and you gain some out of the box setup. You also then benefit from the layout options available for Tabs as well.

philschanely:
@kajabijamell in regard to testing steps, it appears this is a new use case and should not affect exitsing uses of SageChoice. For soundness of mind, consider double-checking some other uses of SageChoice or SageTabs with the `style: "choice"` to confirm and then include such an instance as a place QA can double-check as confirmation existing implementations are unaffected. See "Samples" under [#1 here](https://github.com/Kajabi/sage-lib/wiki/(WIP)-Version-Bump-Process#1-your-featurefix-pr)

QuintonJason (approved):

Issue #328, opened 2021-02-22
Tooltip on Dropdown items


philschanely:
## Description

- [ ] Look into whether or not Tooltips are supported on Dropdown items and add this ability if necessary.
- [ ] Ensure that disabled items allow this.
- [ ] Ensure both Rails and React components add support in similar fashion.

Related Component: #228

sdhull:
Additional request:
- [ ] Ensure that disabled buttons & links can have tooltips and/or popovers

We sometimes want to include a "More info" type of link in these

philschanely:
@pixelflips is this indeed closable?

PR #334, opened 2021-02-23
Add Tooltip Support for Disabled Dropdown Items & Links Component

pixelflips:
## Description
Adds support for tooltips in disabled dropdown items and in the link elements.

**Note:**
This will need to be followed up with support for disabled buttons at some point.

#### Disabled Dropdown Items
|  Rails  |  React  |
|--------|--------|
|Screen Shot 2021-02-23 at 1 31 45 PM|Screen Shot 2021-02-23 at 1 05 28 PM|

#### Links w/ Tooltips
|  Rails  |  React  |
|--------|--------|
|Screen Shot 2021-02-23 at 1 04 25 PM| Screen Shot 2021-02-23 at 1 34 08 PM|

### Rails Testing:

    Disabled Dropdown Options:
    
    1. Visit http://localhost:4000/pages/breakout/object/dropdown
    2. Click the "Add An Element" Button
    3. Verify the tooltip displays when hovering over the "Disabled w/ Tooltip" item.
    4. In addition, click the dropdown under the Select heading and verify when hovering over the disabled item that the tooltip appears.
    
    Link Tooltip:
    
    1. Visit http://localhost:4000/pages/breakout/element/link
    2. Hover over any links where link text indicates there is a tooltip.
    3. Verify tooltip appears when hovering over a link.
    4. Verify `attributes` documentation is correct under Properties.

 
### React Testing:

    Disabled Dropdown Options:
    
    1. Visit http://localhost:6006/?path=/story/sage-dropdown--default
    2. Click the feature dropdown button.
    3. Hover the "Disabled w/ tooltip" option.
    4. Verify tooltip appears.
    
    Link Tooltip:
    
    1. Visit http://localhost:6006/?path=/story/sage-link--tooltip
    2. Hover over the link.
    3. Verify tooltip appears.

## Related

- #328

### DIFF
--- a/packages/sage-react/lib/Dropdown/DropdownItem.jsx
+++ b/packages/sage-react/lib/Dropdown/DropdownItem.jsx
@@ -84,6 +84,15 @@ export const DropdownItem = ({
 
   const fieldId = `${groupId}-checkbox-${id}`;
 
+  const renderOptions = () => (
+    <OptionsDropdown
+      className="sage-dropdown__item-options"
+      id={`${groupId}-${id}-options`}
+      exitPanelHandler={onExitSubmenu}
+      options={options}
+    />
+  );
+
   const renderItem = () => {
     if (isHeading) {
       return <>{label}</>;
@@ -158,12 +167,7 @@ export const DropdownItem = ({
           <li className={classNames} role="none">
             {renderItem()}
             {options && (
-              <OptionsDropdown
-                className="sage-dropdown__item-options"
-                id={`${groupId}-${id}-options`}
-                exitPanelHandler={onExitSubmenu}
-                options={options}
-              />
+              renderOptions()
             )}
           </li>
         </Tooltip>
@@ -171,12 +175,7 @@ export const DropdownItem = ({
         <li className={classNames} role="none">
           {renderItem()}
           {options && (
-            <OptionsDropdown
-              className="sage-dropdown__item-options"
-              id={`${groupId}-${id}-options`}
-              exitPanelHandler={onExitSubmenu}
-              options={options}
-            />
+            renderOptions()
           )}
         </li>
       )}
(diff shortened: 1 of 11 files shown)


monkeypox8 (commented):
Hmm, this one is tricky. By allowing tooltips on the disabled elements, we have to re-enable the `pointer-events`, which means that they are then selected/actionable: 

disabled-dropdown-item

One way to circumvent this might be to check for a `disabled` class and add  `preventDefault` on the event listener(s), but I'm not certain adding that complexity is the way we want to go. Hoping the group can weigh in with thoughts here.

pixelflips:
@monkeypox8 great catch! I was only checking for the tooltip and missed that. I will also wait on others to chime in, but will start looking into a solution in the meantime.

pixelflips:
@monkeypox8 I've just pushed a fix for the issue you brought up. Thanks again for catching that and please check it out when you have some time.

packages/sage-system/lib/dropdown.js
```diff
@@ -73,6 +74,15 @@ Sage.dropdown = (function() {
       return;
     }
 
+    // Stop if the dropdown item clicked and parent is disabled
+    if (el.parentNode.classList.contains(dropdownDisabledItemClass)) {
```
monkeypox8, line 78: I think we can simplify this to:
```
    if (el.parentNode.classList.contains(dropdownDisabledItemClass)) {
      event.preventDefault();
      return;
    }
```
since the `preventDefault()` should be harmlessly ignored on a button (or at least that's my theory)

monkeypox8 (approved):
👍🏼 LGTM, excellent work! One item to consider on the added conditional, not a blocker. Thanks also for cleaning up the JS!

packages/sage-system/lib/dropdown.js
pixelflips: @monkeypox8 will test it out and update if no issues come up. Thanks for taking a look so quickly!

packages/sage-react/lib/Dropdown/DropdownItem.jsx
```diff
@@ -214,4 +234,10 @@ DropdownItem.propTypes = {
     PropTypes.func,
   ]),
   to: PropTypes.string,
+  tooltip: PropTypes.shape({
+    position: PropTypes.oneOf(Object.values(Tooltip.POSITIONS)),
+    size: PropTypes.oneOf(Object.values(Tooltip.SIZES)),
+    theme: PropTypes.oneOf(Object.values(Tooltip.THEMES)),
+  }),
```
joedakroub, line 241: Nice 👍🏼

joedakroub (approved):
Ran this locally; all looks great. Nice work on this implementation and nice cleanup on that JS too. Thanks Phillip! 👍🏼

PR #336, opened 2021-02-23
Storybook cleanup prep

philschanely:
## Description

This PR adds some utilities and example Storybook cleanup for a few components:

- Button and Button group are demonstrations of simple/standard cleanup
- Breadcrumbs shows how to document data structures passed into a component such as `items` here.
- Modal is documented using `subcomponents` and uses both a standard story component and a composed component for demonstrating the core component and a wired version.

## Testing

See Storybook for each of the components above.

No impact on existing components.


QuintonJason (approved):
Nice!

PR #339, opened 2021-02-24
Makes pagination less subjective and adds parameter for collection name

kajabijamell:
## Description 

The pagination component works great under the correct conditions. However, in an effort to make it more generalized, I made the following two adjustments.

**1. It currently only accepts a collection that's a descendant of an `Active Record` class.**
This requirement makes it impossible to feed collections items that are non-descendant of Active Record classes. For example, We tried feeding the pagination a Website Pages collection list that is a descendant from `Themefile`. Even though the collection is enumerable, the check explicitly tries for Active Record descendants, and it, therefore, raised an error. 

**2. It creates a name based on the collection `entry_name` without the option for an override.**
The `entry_name` for our collection was "Template Theme File", this is less than ideal from a UX perspective. To get around this I added a new parameter to override the collection name. 

### Example use case:
```
<%= sage_component SagePagination, {
  items: presenter.website_pages,
  window: 1,
  hide_pages: hide_pages,
  collection_name: "Collection Name"
} %>
```
**Collection name is pluralized and titleized

### Screenshots
|  before  |  after  |
|--------|--------|
|image|image|

### Steps for testing
1. The first fix should allow pagination for work with any enumerable object in rails. 
2. Pass the component a `colection_name` attribute as a string. 
3. Preview the page to reveal the pagination elements.
4. The name should be pluralized and display the name as a string.

### DIFF
--- a/docs/lib/sage_rails/app/sage_components/sage_pagination.rb
+++ b/docs/lib/sage_rails/app/sage_components/sage_pagination.rb
@@ -1,10 +1,11 @@
 class SagePagination < SageComponent
   set_attribute_schema({
-    items: -> (v) { SageSchemaHelper.is_active_record?(v) },
+    items: -> (v) { SageSchemaHelper.will_paginate?(v) },
     window: [:optional, Integer],
     hide_pages: [:optional, TrueClass],
     hide_counter: [:optional, TrueClass],
-    additional_params: [:optional, Hash]
+    additional_params: [:optional, Hash],
+    collection_name: [:optional, String]
   })
 
   def initialize(attributes = {})
@@ -21,7 +22,11 @@ class SagePagination < SageComponent
   end
 
   def page_count(collection)
-    entry_name = (collection.entry_name || "Record").pluralize(collection.total_count)
+    unless collection_name.blank?
+      entry_name = collection_name
+    else
+      entry_name = (collection.entry_name || "Record").pluralize(collection.total_count)
+    end
 
     if collection.total_pages < 2
       "<strong>#{collection.total_count}</strong> #{entry_name}"
--- a/docs/lib/sage_rails/app/sage_components/sage_pagination.rb
+++ b/docs/lib/sage_rails/app/sage_components/sage_pagination.rb
@@ -22,11 +22,9 @@ class SagePagination < SageComponent
   end
 
   def page_count(collection)
-    unless collection_name.blank?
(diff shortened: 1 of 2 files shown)


docs/lib/sage_rails/app/sage_components/sage_pagination.rb
```diff
@@ -21,7 +22,11 @@ def next_text
   end
 
   def page_count(collection)
-    entry_name = (collection.entry_name || "Record").pluralize(collection.total_count)
+    unless collection_name.blank?
+      entry_name = collection_name.titleize.pluralize(collection.total_count)
+    else
+      entry_name = (collection.entry_name || "Record").titleize.pluralize(collection.total_count)
+    end
```
anolson, line 29: Using `unless/else` can be difficult to grok, I'd go for an `if/else` here instead.

docs/lib/sage_rails/app/sage_components/sage_schema_helper.rb
```diff
@@ -19,8 +19,8 @@ module SageSchemaHelper
     left: [:optional, Set.new([:xs, :sm, :md, :lg, :xl, "xs", "sm", "md", "lg", "xl"])],
   }
 
-  # Accepts ActiveRecord descendants
-  def self.is_active_record?(value)
-    value.superclass == ActiveRecord::Base
+  # Accepts any Collection that can be paginated
+  def self.will_paginate?(value)
+    value.respond_to?(:total_pages)
```
anolson, line 24: This seems like a great improvement! 👍 

There's another Ruby pagination gem called `will_paginate` and this might cause confusion for someone new to the codebase. I'm fine with the name, but there might be a better option.

anolson (commented):
Looks good, just a couple notes.

docs/lib/sage_rails/app/sage_components/sage_pagination.rb
sdhull, line 29:
```suggestion
    name = collection_name.presence || collection.entry_name || "Record"
    entry_name = name.titleize.pluralize(collection.total_count)
```

sdhull (approved):
lgtm

little refactor suggestion if you like the looks of it

anolson (approved):

joedakroub (approved):
This looks great; a really nice addition to pagination. Thanks Jamell! 👍🏼

PR #343, opened 2021-02-25
[BUILD-792] Kajabi-Products Toast Usage Support

AndrwM:
BUILD-792

### Docs Preview
https://user-images.githubusercontent.com/565743/117182356-3af04a00-ada4-11eb-8856-271a213643a1.mp4
### Kajabi Products Context
https://user-images.githubusercontent.com/565743/117211603-3b4e0c80-adc7-11eb-8452-debac1a81420.mp4

https://user-images.githubusercontent.com/565743/117212774-c4b20e80-adc8-11eb-944a-319accd00fc0.mp4

Getting Toasts in a place where Kajabi-Products can use them within its `App.flash` abstraction.

**Goal Completion Pattern**
Aside from the addition of the "Goal Completion" CTA-link pattern that Kajabi-Products needs visually this does not change.

**DOM Injection within Kajabi-Products context**
The majority of the work here has been refactoring the current component to be compatible with Kajabi-Products. This now involves injecting a wrapper div to contain the Toasts rather than relying on `nth-of-type` for positioning which made styling fragile to the document structure.

**JS Templating Pattern**
Additionally this PR adds a templating pattern within our vanilla ES6 JS. This is currently 1 of 2 components (Toasts and Tooltips) that _do not_ have Rails component html. The markup for these originate exclusively from JS because we expect these components to be built dynamically on trigger rather than relying on the backend (Rails components) to render the HTML.

**Toast Removal**
When dismissed a toast is no longer removed automatically from the DOM, only hidden. This is to match the pattern in Kajabi-Products. Feature specs check often for the presence of flash/toast text in a test. Calling `reset()` will remove if needed.

**JS Interaction Documentation**
Included in this PR is JS documentation describing the recommended pattern for manually dismissible toasts.
```
    // Pattern for a manually dismissible toast:

    // Sage.toast.trigger()
    // -----------------------------------------------------------------------------------------------------
    // Triggers a toast with the arg and returns the unique id of the toast.
    //
    // @param {object} - Toast options, `message` key is required, all others are optional
    // @returns {unique id}
    //
    // Example:
    const myWonderfulToastId = Sage.toast.trigger({message: 'I will not dismiss via timer', timer: false});


    // Sage.toast.dismiss()
    // -----------------------------------------------------------------------------------------------------
    // Dismisses a toast with a unique id.
    //
    // @param {unique id}
    // @returns {boolean} – `false` if no toast of that ID exists, `true` if it does exist and was successfully dismissed
    //
    // Example:
    Sage.toast.dismiss(myWonderfulToastId);
```
Source: https://github.com/Kajabi/sage-lib/commit/14ddd49d44e4f06a3a904e28b9559d9352fca5de#diff-b95d8cd064f74ea1152f874dcf45e45980ef1ac07da97a706f9cc87718cf7af8R52-R73

-----------

### Details
- Ability to manually manage a toast as an alternative to the timer (e.g. "Uploading Media" and closing when complete)
- Vanilla JS templating solution
- Update React toast html to support new structure (non-breaking)
- Renamed argument `message` -> `text` so that the KJB Flash message arguments can directly map to these new Toasts
- Added CTA link support for Kajabi-Products' "goal completion" flash message pattern
- Built lightweight button styles within the Toast component so that we don't have to hardcode Sage Buttons within the template

Sage JS utilities added:
- Sage Util: small id token generator
- Sage Util: string to DOM fragment converter
- Sage Util: object to stringified html attributes ready for injection

Documentation for info on this components usage:
http://localhost:4000/pages/element/toast

## Test notes
http://localhost:4000/pages/element/toast

## Related
BUILD-792


packages/sage-react/lib/Toast/Toast.story.jsx
```diff
@@ -18,4 +18,19 @@ storiesOf('Sage/Toast', module)
         color={select('Color', Toast.COLORS, Toast.COLORS.DEFAULT)}
       />
     </div>
+  ))
+  .add('Trigger Pattern', () => (
+    <div style={{ marginTop: 50, maxWidth: 540 }}>
+      <p className="t-sage-body">
```
philschanely, line 24: Could you please provide an example of this in use in react context such as here in this story? If all is working well we could then mark this component as deprecated and assert this use as the official new way to use Toast.

JulaB (commented):
This looks great, @AndrwM!

I can see that it would be useful to have shortcut methods for common use cases:

`Sage.toast.error('Danger toast created.')`
for 
`Sage.toast.trigger({icon: 'danger', message: 'Danger toast created.', type: 'danger'})`

`Sage.toast.success('This operation was successful')`
for 
`Sage.toast.trigger({icon: 'success', message: 'This operation was successful', type: 'success'})`

`Sage.toast.notice('Some info message')`
for 
`Sage.toast.trigger({icon: 'notice', message: 'Some info message', type: 'notice'})`

For spinner toasts (they usually look like 'notice' + spinner/loader before the message):

`let toastId = Sage.toast.loading('Loading message....')`
for 
`let toastId = Sage.toast.trigger({message: 'Loading message', timer: false, icon: 'spinner?', type: 'notice'})`

Another usage example:
```
let toastId = Sage.toast.loading('Loading....');
Sage.toast.update(toastId, {message: 'Still loading.....' })
Sage.toast.update(toastId, {message: 'Installing something.....' })
Sage.toast.dismiss(toastId);
```

AndrwM:
@JulaB Thank you for the feedbacl! Great idea for the `update` function. Adding this.

I was originally thinking of having aliases for different toast configurations (success / error / notice / etc) within Kajabi-Products' `App.flash` here: https://github.com/Kajabi/kajabi-products/blob/0918bc9124e0c21d7da2bc3e5b3bec5906ad2930/app/assets/javascripts/admin/flash_messages.js

But maybe it would be best to have those aliases here, less complication. 🚀

docs/app/views/examples/elements/toast/_preview.html.erb
```diff
@@ -3,24 +3,72 @@
...
+    style: "primary",
+    raised: true,
+    attributes: {
+      "onClick": "alert('TOAST ID: ' + Sage.toast.trigger({message: 'I will not dismiss via timer', timer: false}))",
+    },
+  } %>
+<% end %>
+
+<pre class="prettyprint">
+  <code>
```
QuintonJason, line 51: Missing closing `</code>`

QuintonJason:
@AndrwM this is great! One accessibility enhancement. Since there's an interactive element within the toast, so the toast needs to receive focus on the first interactive element. If you don't want the toast to show with a focus state enabled on the interactive element, we can put the focus on the toast container. Upon closing the interactive toast, returning focus back to the interactive toast trigger.

packages/sage-react/lib/Toast/Toast.story.jsx
philschanely: With the approach set up in `kajabi-products` could that method be documented here so that folks trying to use Toast in React see the requirements and method distinctly here?

philschanely (commented):
LGTM but note the comments. My biggest concern is to ensure that the fact that the markup for this is in a (currently) unusual place inside a vanilla JS file is somehow called out and easy to find, and that the React-based method is explicitly documented, if not actually exacted for demo purpose in Storybook.

packages/sage-react/lib/Toast/Toast.story.jsx
AndrwM: On it!

packages/sage-system/lib/toast.js
```diff
@@ -1,76 +1,132 @@
...
+  const DATA_ATTR = 'data-js-toast';
+  const DATA_ATTR_CLOSE_BUTTON = 'data-js-toast-close';
+  const TOAST_CLASS_DISMISSED_STATE = 'sage-toast--animate-out';
+
+  // ==================================================
+  // Template
+  // ==================================================
+
+  const toastTemplate = ({id, type, icon, text, link, timer}) => (`
+    <dialog
```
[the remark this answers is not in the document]
AndrwM: I removed this concept. This PR is only going to handle the Rails side of this component for now.

docs/app/views/examples/shared/_code.html.erb
```diff
@@ -11,19 +11,20 @@
...
+  <%= sage_component SageCardHeader, { title: "Attribute Schema"} do %>
+    <code><%= component_classname %>::ATTRIBUTE_SCHEMA</code>
   <% end %>
+  <div class="example__code">
+    <button class="example__expand-btn" aria-expanded="false" aria-controls="example-rails-code-<%= title %>">Expand this code snippet</button>
+    <div id="example-rails-code-<%= title %>" class="sage-code-snippet" tabindex="-1">
+      <pre class="prettyprint"><code><%= component_classname.constantize::ATTRIBUTE_SCHEMA.pretty_inspect %></code></pre>
+    </div>
+  </div>
+<% end if component_classname.is_a?(Class) %>
```
AndrwM, line 27: Prevent the Docs code view from raising an exception if a Ruby class for the component doesn't exist.

docs/app/views/examples/shared/_code.html.erb
```diff
@@ -1,28 +1,33 @@
...
+          <pre class="prettyprint"><code><%= klass::ATTRIBUTE_SCHEMA.pretty_inspect %></code></pre>
+        </div>
       </div>
-    </div>
+    <% end %>
+  <% else %>
+    <%= sage_component SageAlert, {
+      color: "info",
+      title: "This component does not provide a SageRails class.",
+    } %>
```
AndrwM, line 30: Prevent raising a not found error and display a message to viewer if a rails component class doesn't exist for the documented component.

packages/sage-system/lib/utils/index.js
```diff
@@ -0,0 +1,15 @@
+export function generateId(prefix) {
+  const hash = Math.random().toString(36).substring(2, 15) + Math.random().toString(36).substring(2, 15);
+  return `${prefix}-${hash}`;
+};
+
+export function stringToHtmlFragment(string) {
+  return document.createRange().createContextualFragment(string);
```
joedakroub, line 7: Nice use of this rarely-used function! 💯

packages/sage-system/lib/utils/index.js
```diff
@@ -0,0 +1,15 @@
...
+};
+
+export function objectToHtmlAttributes(object) {
+  return Object.entries(object).map(attribute => {
+    const [key, value] = attribute;
+    return `${key}="${value}"`;
+  }).join(' ');
```
joedakroub, line 14: Super clean! 👍🏼

packages/sage-system/lib/toast/toast.template.js
```diff
@@ -0,0 +1,57 @@
...
+    class="sage-toast sage-toast--style-${type}"
+    id="${id}"
+    ${DATA_ATTR}
+  >
+    <i class="sage-toast__icon sage-icon-${icon}"></i>
+    <output
+      class="sage-toast__value"
+      aria-live="assertive"
+      aria-atomic="true"
+    >
```
joedakroub, line 22: This is terrific!

joedakroub (approved):
This is really solid work; much-needed! Thanks @AndrwM ! 👏🏼

philschanely (approved):
This is looking really good. Thanks, @AndrwM .

Side note FWIW: We may be able to use React Portals to help tie the Rails and React toast/flash together moving forward. More to come!

AndrwM:
Thank you everyone!

PR #352, opened 2021-03-02
Sync header to design specs

philschanely:
This PR brings the application header up to spec and distinguishes it from the Assistant. Future work will include work on the Search/Assistant itself and prep for the Account dropdown.

Screen Shot 2021-03-09 at 12 27 23 PM
Screen Shot 2021-03-09 at 12 27 45 PM

### Testing/Impact

See header in Docs and test responsive flexibility.

(Breaking) Adjustments to Sage Header and Assistant. Impacts implementation in Super Admin. Fix to be provided in [this PR](https://github.com/Kajabi/kajabi-products/pull/18071) along with testing steps.


packages/sage-assets/lib/stylesheets/layout/_header.scss
```diff
@@ -5,6 +5,118 @@
...
+.sage-header__branding {
+  display: flex;
+  grid-area: branding;
+  height: $-header-branding-height;
+}
+
+.sage-header__branding-img {
+  height: 100%;
+
+  @media (max-width: sage-breakpoint()) {
```
monkeypox8, line 75: What are your thoughts around having the mobile nav visible up through tablet-portrait breakpoints as the figma doc indicates? I'm not sure that we ever made that determination completely clear across the app due to the varying container sizes. I do think it makes sense here within the docs, though.

philschanely: Good call. I think I'd leave it where we have it for now until we can further settle the broader grid adjustment point too. If anything, in the recent Sage sync it sounded like responsive type were desired to nudge down to adjust below `768px`.

joedakroub (approved):
This looks great! No concerns other than what Kevin mentioned about the mobile nav being visible through table portrait mode. 👍🏼

packages/sage-assets/lib/stylesheets/layout/_header.scss
philschanely: @monkeypox8 after @joedakroub 's second I looked more and realized that hiding the menu at the larger breakpoint fits better with what we talked about a few months back in regard to future grid directions and helps the layout appear wider longer. These changes are drafted here now... what do you think?

packages/sage-assets/lib/stylesheets/layout/_header.scss
```diff
@@ -5,6 +5,118 @@
...
+
 .sage-header {
+  display: grid;
+  gap: sage-spacing(lg);
+  align-items: center;
   z-index: sage-z-index(nav);
+  height: $-header-height;
+  padding: 0 sage-spacing(lg);
+  background: sage-color(primary, 500);
+
```
monkeypox8, line 24: One last thing: we're setting a `z-index` but without a creating a stacking context since the parent is no longer a flex container, so I'd suggest setting `position` on this element. That's assuming we do actually want it to appear above the overlay blur, though. If not, we can safely remove the `z-index`.

monkeypox8 (approved):
👍🏼 LGTM, thanks for making these changes! One item to ponder around the overlay blur stacking order… I'm not sure what the expected behavior actually is.

PR #355, opened 2021-03-03
Remove manifest file

nobodyiscertain:
Found an empty manifest file in the public directory, I'm skeptical this is in the entire issue, but I'm hoping this will get us past the empty manifest error.


nobodyiscertain:
Handled in #358 

Closing out.

PR #358, opened 2021-03-03
Update docs config for Rails

nobodyiscertain:
This is a handful of updates to get the docs site serving up on Heroku:
* Add missing dependencies required to compile assets
* Update homepage image to use image_tag vs hardcoded image path
* Require `pp` for `#pretty_inspect` to work
* Removed empty manifest and ignore assets directory
* Update web process to use ruby instead of npm


docs/Procfile
```diff
@@ -1 +1 @@
-web: npm run rails:server:production
```
nobodyiscertain, line 1: This was requiring us to need the node buildpack, which I think we don't need if we just use Ruby.

docs/app/helpers/application_helper.rb
```diff
@@ -1,3 +1,5 @@
+require "pp"
```
nobodyiscertain, line 1: After assets were compiling, it was barking about not having `pretty_inspect`. I thought that Pretty Print was auto-included in Ruby, but thinking it might not be available in production? Little confused about this honestly, but this gets it working.

docs/Procfile
```diff
@@ -1 +1 @@
...
+web: bundle exec rails server -e production -b 0.0.0.0 -p $PORT
```
nobodyiscertain, line 1: I'm not sure we should be using WEBrick (rails s) for production. But thinking we could swap it to puma/unicorn in a follow-up.

docs/package.json
```diff
@@ -36,11 +36,15 @@
     "@rails/webpacker": "4.2.2",
     "arrive": "^2.4.1",
     "core-js": "^3.6.5",
+    "css-loader": "^5.1.1",
```
voodooGQ, line 39: Putting these as part of the primary dependencies may present conflicts when using the Sage bridge or when integrating into Kajabi Products. We should make sure to test these thoroughly before moving forward.

voodooGQ (changes requested):
Just one note, I have it as "request changes" until we can have someone in UX vet out the local Kajabi bridge and the Kajabi Products integration.

philschanely:
Pulled, bootstrapped, and checked `kajabi-products` with the bridge per @voodooGQ 's suggestion. No concerns here! 

👍  from me and THANK YOU!

docs/package.json
nobodyiscertain: Ah, good thinking! Looks like @philschanely was able to confirm we're good with the bridge. How can we check it with Kajabi Products? Should we get QA involved?

voodooGQ (approved):
Sorry for the late approve! If @philschanely feels good about it then I'm on board. Thank you!

PR #360, opened 2021-03-04
Table: fix for content overflow in grid-based parents

monkeypox8:
## Description
Sage responsive tables are causing [grid blowouts](https://css-tricks.com/preventing-a-grid-blowout/) when located inside of a card or panel. As I understand it, this is due to the way `display: grid` handles horizontal overflow. Though the grid templates default to a `min-content` which works with most inline content, in the case of tables the "minimum" content width cannot wrap and flows outside of the grid container.

The proposed solution here adds `.sage-table-wrapper--scroll` to the main table wrapper, which sets `overflow: hidden` along with a `max-width` when embedded in a card or panel.

This is a _partial_ fix for `COMM-780`, however the table `sage-table-wrapper__overflow` container also prevents the dropdown menu from fully rendering vertically outside of the wrapper. We'll need a separate, long-term solution to correct that issue.

### Screenshots

|  Before  |  After  |
|--------|--------|
|sage-table-before|sage-table-after|

### Estimated impact
- [ ] (MEDIUM) Table overflow wrapper modified to truncate content within a card or panel
   - [x] Coupons index (Sales -> Coupons)
   - [x] Offers index (Sales -> Offers) (wrapper will be removed in `COMM-780`)
   - [ ] Site custom domains 
   - [ ] Podcast episodes (can't tell if this is using a Card or Panel wrapper)


joedakroub (approved):
Very nice fix; works well. Thanks for looking into this Kevin! 👍🏼

philschanely (approved):
Thanks for this!

PR #364, opened 2021-03-04
Rails SageSortable: Fix bug where Sortable onload injects JS response twice.

AndrwM:
### Description
Initially identified in Toast development. In testing the new toast messages in Kajabi-Products I realized `xhr.onLoad` fires twice in Chrome.

### Solution
Switch to using an eventlistener for the `Sage.util.ajaxRequestWithJsInjection()` that is used internally with Sortable component.


philschanely (approved):
No concerns here 👍

joedakroub (approved):
Nice catch! Thanks Andrew! 👍🏼

pixelflips (approved):
🔥  LGTM!

PR #365, opened 2021-03-05
Add Icon Card option to FeatureToggle

joedakroub:
## Description
Adds the ability to assign an `IconCard` component to an `icon` property of the `FeatureToggle` component.

This feature enables you to specify an `IconCard` in place of an image like so:

```
<%= sage_component SageFeatureToggle, {
    description: "New navigation bar layout that places Offers, Coupons, and Affiliates under a brand new “Sales” tab.",
    icon: {
      color: "published",
      icon: "tag",
      label: "New Sales Tab",
      size: "lg"
    }
...
```

### Notes
- The component should only render an `image` or an `icon`, not both.
- The complete schema for an `IconCard` can be used to configure the icon
- The icon or image container shrinks to fit the content, so smaller/larger icon sizes WILL adjust the container width; this is by design
- Added `background_color` and `foreground_color` to `IconCard` which can be used in `FeatureToggle`

## Screenshots
| |Image|Icon|
|-|------|----|
|FeatureToggle|Image|Icon|

### Color states

The following states can be achieved based on setting `color`, `foreground_color` and/or `background_color`:

| `color` | `foreground` | `background` | output |
|---|---|---|---|
| not | not | not |  |
| not | not | provided |  |
| not | provided | not |  |
| not | provided | provided |  |
| provided | not | not |  |
| provided | not | provided |  |
| provided | provided | not |  |
| provided | provided | provided |  |

## Testing
1. Visit http://0.0.0.0:4000/pages/object/feature_toggle
2. Visit http://0.0.0.0:4000/pages/element/icon_card

## Affected Product Areas
1. (High) **Admin -> Labs** - Local: http://www.kajabi.test:3000/admin/sites/1/labs
2. (High) **Admin -> Products -> Edit Product** - Local: http://www.kajabi.test:3000/admin/products/12/edit

### DIFF
--- a/docs/app/views/examples/objects/feature_toggle/_preview.html.erb
+++ b/docs/app/views/examples/objects/feature_toggle/_preview.html.erb
@@ -1,4 +1,4 @@
-<h3 class="t-sage-heading-6">Feature Toggle
+<h3 class="t-sage-heading-6">Feature Toggle</h3>
 <%= sage_component SagePanel, {} do %>
   <%= sage_component SageFeatureToggle, {
     alt_text: "Text alt text",
--- a/docs/app/views/examples/objects/feature_toggle/_preview.html.erb
+++ b/docs/app/views/examples/objects/feature_toggle/_preview.html.erb
@@ -33,6 +33,39 @@
     <% end %>
   <% end %>
 <% end %>
+<h3 class="t-sage-heading-6">Feature Toggle with <code><a href="/pages/element/icon_card">Icon Card</a></code></h3>
+<%= sage_component SagePanel, {} do %>
+  <%= sage_component SageFeatureToggle, {
+    description: "New navigation bar layout that places Offers, Coupons, and Affiliates under a brand new “Sales” tab.",
+    icon: {
+      color: "published",
+      icon: "tag",
+      label: "New Sales Tab",
+    },
+    links: [
+      {
+        icon: "comment",
+        location: "#",
+        target: "_blank",
+        text: "Feedback",
+      },
+    ],
+    title: "New Sales Tab",
+  } do %>
+    <%= sage_component_section :feature_toggle_input do %>
+      <%= sage_component SageSwitch, {
+        type: "checkbox",
+        id: "feature-toggle-id-2",
+        hide_text: true,
+        label_text: "Label Title",
+        message: "",
+        name: "feature-toggle-id",
+        value: "feature-toggle-value",
+        css_classes: "sage-feature-toggle__switch"
+      } %>
+    <% end %>
+  <% end %>
+<% end %>
 <h3 class="t-sage-heading-6">Feature Toggle with <code>image</code></h3>
 <%= sage_component SagePanel, {} do %>
   <%= sage_component SageFeatureToggle, {
--- a/docs/app/views/examples/objects/feature_toggle/_preview.html.erb
+++ b/docs/app/views/examples/objects/feature_toggle/_preview.html.erb
@@ -40,7 +40,7 @@
     icon: {
       color: "published",
       icon: "tag",
-      label: "New Sales Tab",
+      label: "New Sales Tab"
     },
     links: [
       {
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_icon_card.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_icon_card.html.erb
@@ -1,9 +1,6 @@
 <%
-color = "draft"
-if component.color.present?
-  color = component.color
-end
-size = component.size.present? ? component.size : "3xl"
+color = component.color.present? ? component.color : "draft"
+size = component.size.present? ? component.size == "md" ? "" : "-#{component.size}" : "-3xl"
 style = {
   "--background-color": (component.background_color if component.background_color.present?),
   "--color": (component.foreground_color if component.foreground_color.present?),
@@ -23,7 +20,7 @@ style = {
   <% end %>
 >
   <i
-    class="sage-icon-<%= component.icon %>-<%= size %>"
+    class="sage-icon-<%= component.icon %><%= size %>"
     <% if component.label.present? %>
     aria-label="<%= component.label %>"
     <% else %>
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_icon_card.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_icon_card.html.erb
@@ -1,16 +1,26 @@
 <%
-size = component.size.present? ? component.size : '3xl'
-color = component.color.present? ? component.color : 'draft'
+size = component.size.present? ? component.size : "3xl"
+color = "draft"
+if component.color.present?
+  color = component.color
+elsif component.background_color.present? and component.foreground_color.present?
+  color = false
+end
 %>
 <div
   class="
     sage-icon-card
-    <%= "sage-icon-card--#{color}" %>
+    <% if component.color.present? %>
+      <%= "sage-icon-card--#{color}" %>
+    <% end %>
     <%= component.css_classes if component.css_classes.present? %>
   "
   <% component.attributes.each do |key, value| %>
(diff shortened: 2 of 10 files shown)


philschanely:
Great work here. A few general thoughts:

- do we need more color variations or color overrides? Seems to me that some of the existing icon sets demanded colors somewhat outside this set. If so, may be nice to have a foreground_color and background_color local override to pass into the icon card.
- do we need more flexibility in the size of the graphic itself? May be hand to pass in a size that controls the width/height of the icon card.
- do the current icon card's icon sizes afford what we need? (depends somewhat on the answer to preceding bullet

joedakroub:
@philschanely Great questions as always:

> do we need more color variations or color overrides? Seems to me that some of the existing icon sets demanded colors somewhat outside this set. If so, may be nice to have a foreground_color and background_color local override to pass into the icon card.

Yes, adding a `background_color`/`foreground_color` option is a good call. I will add that in.

> do we need more flexibility in the size of the graphic itself? May be hand to pass in a size that controls the width/height of the icon card.

The initial request from Court was to reduce the size of the image as he feels it is too large. 

We could add a prop to `FeatureToggle` to specify `image_size` (having a few size options possibly) and have that apply to both the `image` and the `icon` container and then make the children 100% width of the container.

> do the current icon card's icon sizes afford what we need? (depends somewhat on the answer to preceding bullet)

The icon sizes work well for making the icon smaller than the current image, which is what the initial request was. However, offering a few `image_size` options might be a better solution.

packages/sage-assets/lib/stylesheets/patterns/elements/_icon_card.scss
```diff
@@ -12,10 +12,11 @@
   align-self: stretch;
   padding: sage-spacing(lg);
   border-radius: sage-border(radius);
+  background-color: var(--background-color, inherit);
```
philschanely, line 15: 😍

docs/lib/sage_rails/app/views/sage_components/_sage_icon_card.html.erb
```diff
@@ -1,16 +1,26 @@
...
+  color = component.color
+elsif component.background_color.present? and component.foreground_color.present?
+  color = false
+end
 %>
 <div
   class="
     sage-icon-card
-    <%= "sage-icon-card--#{color}" %>
+    <% if component.color.present? %>
```
philschanely, line 13: Given your logic above would `<%= "sage-icon-card--#{color}" if color %>` suffice here?

docs/lib/sage_rails/app/views/sage_components/_sage_icon_card.html.erb
```diff
@@ -1,16 +1,26 @@
...
+      <%= "sage-icon-card--#{color}" %>
+    <% end %>
     <%= component.css_classes if component.css_classes.present? %>
   "
   <% component.attributes.each do |key, value| %>
     <%= "#{key}=\"#{value}\"".html_safe %>
   <% end if component.attributes&.is_a?(Hash) %>
+  <% if component.background_color.present? and component.foreground_color.present? %>
```
philschanely, line 21: Not sure this needs to require both background and foreground. In theory, they could provide a `color` preset. and then provide an override to just one of either foreground or background. 

| `color` | `foreground` | `background` | output |
|---|---|---|---|
| provided | provided | provided | all are set BUT `foreground` and `background` override `color` preset |
| not | provided | provided | `draft` for `color` is cancelled and `foreground` and `background` override |
| regardless | not | not | default `draft` (or provided `color`) takes full effect |
| regardless | not | provided | `color` applies to `foreground` but `background` overrides |
| regardless | provided | not | `color` applies to `background` but `foreground` overrides |

docs/lib/sage_rails/app/views/sage_components/_sage_icon_card.html.erb
```diff
@@ -1,16 +1,26 @@
 <%
-size = component.size.present? ? component.size : '3xl'
-color = component.color.present? ? component.color : 'draft'
+size = component.size.present? ? component.size : "3xl"
+color = "draft"
```
philschanely, line 3: Give discussion below, we may benefit from switching the order:

```rb
color = "draft"
if component.background_color.present? and component.foreground_color.present?
  color = false
elsif component.color.present?
  color = component.color
end
```

docs/lib/sage_rails/app/views/sage_components/_sage_icon_card.html.erb
```diff
@@ -1,16 +1,26 @@
...
+    <% if component.color.present? %>
+      <%= "sage-icon-card--#{color}" %>
+    <% end %>
     <%= component.css_classes if component.css_classes.present? %>
   "
   <% component.attributes.each do |key, value| %>
     <%= "#{key}=\"#{value}\"".html_safe %>
   <% end if component.attributes&.is_a?(Hash) %>
+  <% if component.background_color.present? and component.foreground_color.present? %>
+    style="--background-color: <%= component.background_color %>; --color: <%= component.foreground_color %>"
```
philschanely, line 22: If switch above to `or` is made then these could be `<%= component.background_color if component.background_color.present? %>`, etc.)

philschanely (approved):
Looking good. Not blocking but some thoughts on the logic behind how colors are applied.

docs/lib/sage_rails/app/views/sage_components/_sage_icon_card.html.erb
joedakroub: Thanks so much for walking through this Phil! You're right, the logic can definitely change here. I'll update it.

docs/lib/sage_rails/app/views/sage_components/_sage_icon_card.html.erb
```diff
@@ -1,16 +1,26 @@
...
-color = component.color.present? ? component.color : 'draft'
+color = "draft"
+if component.color.present?
+  color = component.color
+end
+size = component.size.present? ? component.size : "3xl"
+style = {
+  "--background-color": (component.background_color if component.background_color.present?),
+  "--color": (component.foreground_color if component.foreground_color.present?),
+}.compact.map{|k,v| "#{k}:#{v}"}.join(";")
```
joedakroub, line 10: Here we build a string of optional custom props

docs/app/views/examples/objects/feature_toggle/_preview.html.erb
```diff
@@ -1,4 +1,4 @@
-<h3 class="t-sage-heading-6">Feature Toggle
+<h3 class="t-sage-heading-6">Feature Toggle</h3>
 <%= sage_component SagePanel, {} do %>
   <%= sage_component SageFeatureToggle, {
     alt_text: "Text alt text",
```
monkeypox8, line 4: Minor detail here, I think this can be removed since there isn't an image.

packages/sage-assets/lib/stylesheets/patterns/objects/_feature_toggle.scss
```diff
@@ -16,7 +16,7 @@ $-feature-toggle-image-height-mobile: rem(120px);
   grid-row-gap: sage-spacing(sm);
 
   @media (max-width: sage-breakpoint(md-min)) {
```
monkeypox8, line 18: Outside the bounds of the current work, but is this media query intended for a smaller breakpoint? Around mid to large mobile widths (480 - 767px), the component seems oddly confined. I think the second `min-content` can be removed, which should help some. Can then also set `.sage-feature-toggle__aside` to `justify-content: flex-end` to emulate the desktop layout

monkeypox8 (approved):
👍🏼 LGTM, beautifully done. Hoping we can make more use of CSS custom props like this!

packages/sage-assets/lib/stylesheets/patterns/objects/_feature_toggle.scss
joedakroub: Another nice catch! I've removed the 2nd `min-content` and it works well. 👍🏼

PR #367, opened 2021-03-08
Panel Controls: Page size dropdown

philschanely:
This PR adds a Page size dropdown selector option to React Panel Controls.

Also exposes a few classes for aligning text in React.

Screen Shot 2021-03-08 at 2 47 38 PM

### Impact and Testing

(LOW) Adds new page size dropdown option to Panel Controls. 
   - [ ] Only use of this in `kajabi-products` is in the new async Contacts table: ensure it functions as intended (no change: new feature to be incorporated in follow)


joedakroub (approved):
Tested locally in Storybook and all looks as it should. Nice implementation Phil! 👍🏼

monkeypox8 (approved):
👍🏼 LGTM, works great!

monkeypox8:
One item to investigate as a follow-up: we're using `role="dialog"` for many of the dropdown menus in React (including this one), but `role="menu"` for the same component in Rails. Whichever one we'd want to run with, I suspect we'd want to have an `aria-labelledby` to associate the menu with the trigger/button.

PR #368, opened 2021-03-08
Fix logo color

philschanely:
This PR patches a few spots where older Sage colors were in use.

### Testing and impact

(Low) Patches to Sage logo and Loader colors.


monkeypox8 (approved):
👍🏼 Looks like it's time to replace the slack sage emoji

PR #369, opened 2021-03-10
Button group - add modifier for space-between

QuintonJason:
## Description

 - [x] - add modifier for `space-between` alignment - rails
 - [x]  - update `alignEnd` prop to be an array to allow the addition of `space-between` - react

#### Align Space-Between
|  before  |  after  |
|--------|--------|
|does not exist|Screen Shot 2021-03-15 at 9 18 17 AM|

##### Align Space-Between with top border

|  before  |  after  |
|--------|--------|
|does not exist|Screen Shot 2021-03-15 at 9 16 25 AM|

### Steps for testing
1. Visit the button group rails and react page.
    - select react page: http://localhost:4100/?path=/story/sage-button-group--default
    - select rails page: http://localhost:4000/pages/element/button
    
### QA Steps
#369 (Medium) Updates existing instances of `<ButtonGroup>` that have the `alignend` property set
    - [ ] - Product Outlines page, click into an item and hover the row to verify the button group
    - [ ] - Product Outlines page, click into an item, and observe the `Post Thumbnail` panel. Verify that it is unchanged

### PR to apply changes
- https://github.com/Kajabi/kajabi-products/pull/18204

### App checks
- all instances of `alignEnd` pertain to `<Button>` not `<ButtonGroup>` so good here
- all instances of `.sage-bnt-group` have either no modifiers or only modify `.sage-btn-group--gap-*` so no existing button groups are affected

### DIFF
--- a/packages/sage-assets/lib/stylesheets/patterns/elements/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/elements/_button.scss
@@ -436,3 +436,7 @@ $-btn-interactive-label-icon-size: rem(24px);
 .sage-btn-group--align-end {
   justify-content: flex-end;
 }
+
+.sage-btn-group--align-space-between {
+  justify-content: space-between;
+}
--- a/packages/sage-assets/lib/stylesheets/patterns/elements/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/patterns/elements/_button.scss
@@ -440,3 +440,8 @@ $-btn-interactive-label-icon-size: rem(24px);
 .sage-btn-group--align-space-between {
   justify-content: space-between;
 }
+
+.sage-btn-group--border-top {
+  padding-top: sage-spacing(lg);
+  border-top: 1px solid sage-color(grey, 300);
+}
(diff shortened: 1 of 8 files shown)


philschanely (approved):
Looks good. Just to be sure though : this does not affect existing implementations, but requires one to add the `align` property in order to enact the `space-between` setting?

QuintonJason:
> Looks good. Just to be sure though : this does not affect existing implementations, but requires one to add the `align` property in order to enact the `space-between` setting?

@philschanely Yes, since the modification is a new string to the array and the only one prior had that value in the rule declaration, it ensures it doesn't break the existing implementation. But happy to talk through anything if you like.

joedakroub (approved):
Nice addition; much-needed. No concerns here. Thanks Quinton! 👍🏼

packages/sage-assets/lib/stylesheets/patterns/elements/_button.scss
```diff
@@ -436,3 +436,12 @@ $-btn-interactive-label-icon-size: rem(24px);
 .sage-btn-group--align-end {
   justify-content: flex-end;
 }
+
+.sage-btn-group--align-space-between {
+  justify-content: space-between;
+}
+
+.sage-btn-group--border-top {
+  padding-top: sage-spacing(lg);
```
monkeypox8, line 445: As a follow-up, I'm wondering if the padding here could be configurable. For instance, when used within a form to match the margins there.

monkeypox8 (approved):
👍🏼 LGTM!

PR #370, opened 2021-03-11
Modal transition & animation settings

monkeypox8:
## Description
This branch adds a default fade transition to all modals along with options for animated movement. Screens attached below, though the effect should be easier to see on-page.

Transitions are controlled through `data` attributes. Using this implementation as a test case for an extended Sage animation library. 

Notes:
- huge thanks to @philschanely for assistance with the React version 🙇🏼 
- added ability to disable background blur: in testing, the css `filter` tends to increase GPU/CPU usage when animating, resulting in dropped frames

### Screenshots

|  Before  |  After  |
|--------|--------|
|modal-before|modal-after|

## Test notes
In order to implement these transitions, the inactive modal CSS was changed from `display: none` to instead use `visibility: hidden` and a negative `z-index`. This should still be sufficient to "hide" the modal from accidental user interaction, but we'll need to ensure the stacking order is working as expected depending on placement of the modal's markup. In Rails, if `SageModal` is nested within a `<% content_for :modal do %>`, there should not be an issue.

All transitions are disabled when "reduce motion" is enabled on a user's OS, or on browsers that support the feature. Modal backdrop opacity is also increased for additional contrast.

### Steps to test
- Navigate to http://localhost:4000/pages/object/modal
- Open some modals
- (optional) Enable "reduce motion" in your MacOS system preferences -> Accessibility (screen below). All transitions should now be disabled

### Estimated impact
- (HIGH) Modal CSS display property has been altered to allow for transitioning effects. Confirm that modal is displayed on page as expected, and cannot be accessed via user interaction or keyboard navigation. Examples of use:
  - [x] Beta features modal 
  - [x] Offer price modal
  - [x] Create site modal
  - [ ] Coupons: delete Coupon modal
  - [ ] Coupons: add Offer modal

### DIFF
--- a/docs/app/views/examples/objects/modal/_preview.html.erb
+++ b/docs/app/views/examples/objects/modal/_preview.html.erb
@@ -11,14 +11,6 @@
     <%= sage_component SageButton, {
       style: "primary",
       icon: { name: "launch", style: "right" },
-      value: "Modal with Animation",
-      attributes: {
-        "data-js-modaltrigger": "cool-modal-animate",
-      }
-    } %>
-    <%= sage_component SageButton, {
-      style: "primary",
-      icon: { name: "launch", style: "right" },
       value: "Large Modal",
       attributes: {
         "data-js-modaltrigger": "cool-modal-large",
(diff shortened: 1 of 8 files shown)


docs/app/views/examples/objects/modal/_preview.html.erb
```diff
@@ -158,4 +158,185 @@
       <% end %>
     <% end %>
   <% end %>
+
+<% end %>
+
+<%= sage_component SagePanelBlock, {} do %>
+  <h3 class="t-sage-heading-6">Animated</h3>
+  <div class="sage-btn-group">
```
pixelflips, line 166: Could we use the ButtonGroup component here?

`<%= sage_component SageButtonGroup, { gap: :sm, spacer: { bottom: :sm }} do %>`

pixelflips (approved):
👍🏼  LGTM! Checked this out locally and this is awesome! The reduced motion is 🔥  too btw.

Left one minor comment, but definitely non-blocking. Nice work!

joedakroub (approved):
This is terrific work Kevin. I really like the reduced motion option as well. Looks great in the docs and within the app. Nice work! 👍🏼

QuintonJason (approved):
Is the fading out of the modal something we'd be able to handle in a followup? I know there's some complexity around incorporating `requestAnimationFrame`, so it may be more trouble than worth currently.

monkeypox8:
> Is the fading out of the modal something we'd be able to handle in a followup? I know there's some complexity around incorporating `requestAnimationFrame`, so it may be more trouble than worth currently.

Great call, that's something I'd like to see as well. I considered adding an new event before `sage.modal.closeAll` to help set off a closing timer, but it seemed better suited for a follow-up.

PR #378, opened 2021-03-17
Expand card tweaks

philschanely:
This PR adds a few tweaks to Expandable Card:

- Full width button
- Option to add sage type on the card body

Plus:
- Fixes breaking linter on `develop`
- Adds generated_css_class onto Button and Copy Button Card components.

No visual changes otherwise

### Testing and Impact

See Expandable Card in Docs site and Storybook.

1. (LOW) Adds additional class settings to Button and Copy Button Card. No impact on implementations
1. (LOW) Adds additional settings on Expandable Card. No impact on current non-production uses.

### DIFF
--- a/docs/app/views/examples/objects/expandable_card/_props.html.erb
+++ b/docs/app/views/examples/objects/expandable_card/_props.html.erb
@@ -5,6 +5,12 @@
   <td><%= md('`nil`') %></td>
 </tr>
 <tr>
+  <td><%= md('`sage_type`') %></td>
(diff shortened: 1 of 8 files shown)


joedakroub (commented):
These changes look good to me overall. I did notice that the new `sage_type` prop is missing from the Rails docs, however.

joedakroub (approved):
These changes look good to me overall. I did notice that the new `sage_type` prop is missing from the Rails docs, however.

PR #383, opened 2021-03-23
Github PR template corrections

monkeypox8:
## Description
In a previous update to our github PR template, the revised template file(s) were moved to a subdirectory, which caused github to lose the template's association completely. As a result, creating new PRs left the body blank. It appears that multiple templates are supported for issues, but not with pull requests.

This change moves the template file back to the root `.github` directory.

### Estimated impact
n/a


pixelflips (approved):
LGTM! 👍🏼

joedakroub (approved):
Thanks Kevin! 👍🏼

PR #371, opened 2021-03-11
Rails components - verify existence of html_attributes

QuintonJason:
## Description

Added `<%= component.generated_html_attributes.html_safe %>` to all remaining props to ensure that any rails props can be passed any html attributes needed
 - [x] - updated assistant
 - [x] - updated avatar
 - [x] - updated banner
 - [x] - updated billboard
 - [x] - updated body
 - [x] - updated breadcrumbs
 - [x] - updated buttongroup
 - [x] - updated catalog item
 - [x] - updated container
 - [x] - updated content
 - [x] - updated copybutton
 - [x] - updated copytext
 - [x] - updated copytextcard
 - [x] - updated datacard
 - [x] - updated description
 - [x] - updated dropdown
 - [x] - updated expandablecard
 - [x] - updated featuretoggle
 - [x] - updated forminput
 - [x] - updated formsection
 - [x] - updated formselect
 - [x] - updated formtextarea
 - [x] - updated gridcol
 - [x] - updated gridrow
 - [x] - updated header
 - [x] - updated hero
 - [x] - updated iconlist
 - [x] - updated indicator
 - [x] - updated inputgroup
 - [x] - updated inputhelper
 - [x] - updated labelgroup
 - [x] - updated loader
 - [x] - updated meter
 - [x] - updated modal
 - [x] - updated modalcontent
 - [x] - updated nav
 - [x] - updated nullview
 - [x] - updated overlay
 - [x] - updated pageheading
 - [x] - updated pagination
 - [x] - updated panel controls
 - [x] - update popover
 - [x] - updated search
 - [x] - updated sidebar
 - [x] - updated sortable
 - [x] - updated sortableitem
 - [x] - updated stage
 - [x] - updated statusicon
 - [x] - updated switch
 - [x] - updated tabspane
 - [x] - updated tooltip
 - [x] - updated transactioncard
 - [x] - updated uploadcard

### Screenshots

No visual changes

### Steps for testing
There is no visual change. This PR allow for adding `html_attributes` to any Rails component to add custom data-attributes and other properties.

The best way to test is to turn on the link and add a prop to any component -> inspect that element -> verify that the `html_attributes` have been added
    
### QA Steps
(Low) This updates existing Rails components but props are new features so there should be no difference.

PR #386, opened 2021-03-25
Add attrs to sortable

philschanely:
No visual change... adds `html_attributes` output to `SageSortable`.

### Testing and Impact

Adds html attributes output on Sage Sortable. No impact on existing components. Check the following to confirm they behave as expected:
   - [ ] Website > Navigation ... sortable items retain existing functionality


pixelflips:
@philschanely it looks like you guys might have had the same idea: https://github.com/Kajabi/sage-lib/pull/371

PR #390, opened 2021-03-29
Modal - add header image to modalcontent

QuintonJason:
## Description

- [x] - added an optional header image to the modal content
- [x] - updated docs and demo

### Screenshots

|  Before Modal Trigger Click  |  After Modal Trigger Click  |
|--------|--------|
|Screen Shot 2021-03-29 at 4 58 49 PM|Screen Shot 2021-03-29 at 4 58 54 PM|

## Test notes

- Visit the Modal Rails page: http://localhost:4000/pages/object/modal
- Click the Modal with header image and verify the image is present 

### Estimated impact

- [ ] (**LOW**) This is a new feature and does not affect any instances of modals

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_modal_content.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_modal_content.html.erb
@@ -1,5 +1,8 @@
 <% if component.title.present? %>
   <header class="sage-modal__header">
+    <% if component.header_image.present? %>
+      <img class="sage-modal__header-image" src=<%= component.header_image %> alt="" aria-hidden="true"/>
+    <% end %>
     <h1 class="t-sage-heading-4">
       <%= component.title %>
     </h1>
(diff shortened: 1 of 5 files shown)


docs/lib/sage_rails/app/views/sage_components/_sage_modal_content.html.erb
```diff
@@ -1,5 +1,8 @@
 <% if component.title.present? %>
   <header class="sage-modal__header">
+    <% if component.header_image.present? %>
+      <img class="sage-modal__header-image" src=<%= component.header_image %> alt="" aria-hidden="true"/>
```
monkeypox8, line 4: Could there be a case where we might want to treat the image as non-decorative? I think the answer for now is "no", just putting it out there.

monkeypox8 (approved):
👍🏼 LGTM!

docs/lib/sage_rails/app/views/sage_components/_sage_modal_content.html.erb
QuintonJason: Good point. My thinking upon creation centered around it's location. Since it's in the header of a modal, I couldn't see an instance in which a modal image would not be a supporting image of the title. We can definitely address this in a follow when necessary.

PR #393, opened 2021-03-30
Next version

philschanely:
### Estimated impact

1. #389 (LOW) Updates to Radio, Switch, and Toggle to include a message/hint and correct state coloring. No adverse effects expected. Investigate any Sage-based forms such as the following for continued expected behavior:
   - [ ] Blog post editor form (radio for publishing on the right sidebar)
   - [ ] Navigation editor, edit a link modal (checkbox for whether or not to be only visible to logged-in users)
1. #392 (MEDIUM) This style update affects all instances of `SagePagination` that exists within `SagePanel` instead of `SagePanelControls`.
    - [ ] - Product Index page
    - [ ] - Podcast Index page
1. - #391 (LOW) Modal: enable custom CSS classes. No expected impact from adding this optional param. Check the following implementations to verify expected functionality:
    - [ ] Sites -> New Site modal
    - [ ] Product blueprint -> New Product modal
1. #390 (LOW) Modal: add header image. This is a new feature and does not affect any instances of modals
1. #388 (**LOW**) Added `generated_css_classes` props to grid and panel rails components. Shouldn't affect any sage components
    - [ ] Product index page
    - [ ] Podcast index page
1. #395 (LOW) Adjusts pagination to allow for a prefix or suffix string to be added to the page count section. This should not affect any current implementations as it is a new feature; verify the following still behave as expected:
   - [ ] People > pagination below table
   - [ ] Offers > pagination below list of offers
1. #397 (LOW) Added optional radio buttons behind "choice" tab components for use in form contexts. Confirm no effect on existing uses of choice tabs in:
   - [ ] Website > Navigation > Add new item and in modal, choice tabs function as planned.
1. #399 (LOW) Adds ability to suppress searching again when search terms are cleared. No impact on existing implementations; verify the following search components continue to behave as expected:
   - [ ] Go to: `/admin/sites/SITE_ID/email_campaigns` and clearing the search field should execute the search
   - [ ] Go to: `/admin/sites/SITE_ID/products` and clearing the search field should execute the search

PR #398, opened 2021-04-01
Adds remote_url option for modals

sdhull:
## Description
There are plenty of places in the app where it would be useful to load modal content via ajax. This has been done in the past in an (imo) unnecessarily complicated way. For example, see [this open pr](https://github.com/Kajabi/kajabi-products/pull/18437), which took this basic idea from elsewhere in the app.

One related problem is that there is no `open` event for the modal (this would be my next PR, regardless). With an `open` event, devs could do the footwork. But why should we make them re-invent the wheel over and over again?

Making "modals with remote content" should be easy & ergonomic. I can think of at least two other modals our team worked on recently that would benefit from this.

This PR makes it drop-dead simple to make a modal with content that loads via ajax. Simply pass `remote_url` as an option to the modal and hook up the endpoint to render (without layout of course) and BOOM! Remote modal.

### Screenshots
N/A no visual changes

## Test notes
Make sure your controller endpoint does not render layout :grimacing:

### Estimated impact
Adds ability to load content asynchronously in Modals easily. No impact on existing implementations; Verify the following modals continue to behave as expected:
- [x] Website > Navigation > + Add New Navigation Menu
- [x] Website > Navigation > + Add (Button on a navigation panel) > Add Product

### DIFF
--- a/docs/app/views/examples/objects/modal/_props.html.erb
+++ b/docs/app/views/examples/objects/modal/_props.html.erb
@@ -53,6 +53,12 @@
   <td><%= md('`nil`') %></td>
 </tr>
 <tr>
+  <td><%= md('`remote_url`') %></td>
+  <td><%= md('Specify a url which from which to load the modal content on open. This will be requested & content replaced every time the modal is opened.') %></td>
+  <td><%= md('Boolean') %></td>
+  <td><%= md('`nil`') %></td>
+</tr>
+<tr>
   <td colspan="4">
     <strong>Modal Content</strong>
   </td>
--- a/docs/app/views/examples/objects/modal/_props.html.erb
+++ b/docs/app/views/examples/objects/modal/_props.html.erb
@@ -55,7 +55,7 @@
 <tr>
   <td><%= md('`remote_url`') %></td>
   <td><%= md('Specify a url which from which to load the modal content on open. This will be requested & content replaced every time the modal is opened.') %></td>
-  <td><%= md('Boolean') %></td>
+  <td><%= md('String') %></td>
   <td><%= md('`nil`') %></td>
 </tr>
 <tr>
--- a/packages/sage-system/lib/modal.js
+++ b/packages/sage-system/lib/modal.js
@@ -104,11 +104,12 @@ Sage.modal = (function() {
   }
 
   function fetchModalContent(el) {
+    let elContainer = el.querySelector(`[${SELECTOR_MODAL_CONTAINER}]`);
     let url = el.getAttribute(SELECTOR_MODAL_REMOTE_URL);
     const xhr = new XMLHttpRequest();
 
     xhr.addEventListener('load', (evt) => {
-      el.innerHTML = evt.currentTarget.response;
+      elContainer.innerHTML = evt.currentTarget.response;
     }, { once: true });
 
     xhr.open('GET', url, true);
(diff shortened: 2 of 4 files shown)


packages/sage-system/lib/modal.js
```diff
@@ -5,8 +5,9 @@ Sage.modal = (function() {
   // ==================================================
 
   const SELECTOR_MODAL = 'data-js-modal';
-  const SELECTOR_MODAL_CONTAINER = '.sage-modal__container';
+  const SELECTOR_MODAL_CONTAINER = 'data-js-modal-container';
```
sdhull, line 8: Note: changed this selector to follow the same convention as the others (eg, `data-js-` attribute rather than css class)

packages/sage-system/lib/modal.js
```diff
@@ -99,9 +103,22 @@ Sage.modal = (function() {
     removeModalContents(el);
   }
 
+  function fetchModalContent(el) {
```
philschanely, line 106: Nice addition! 🥇

philschanely (approved):
Thanks for this, @sdhull ! Excited for folks to be contributing ideas like this here!. Couple of quick notes for tidying the PR per our current practices:

- [x] **Merge target:** Go ahead and rebase this against the `develop` branch. We're using that as a buffer for new work since merging to master automatically triggers new package version numbers and we found we preferred to intentionally release sets of work together on a weekly cycle. We'll keep you looped in as that progresses and your work lands in `kajabi-products`. You can see our current "Next version" PR here #393 
- [x] **Testing steps:** If you haven't already consider using the "bridge" to see and test your work in `kajabi-products`. What we've been doing lately even when our work should not cause any issues on any existing implementations is to provide an example or two of the affected component in use for QA to validate no impact. So in this case you could add something like this under your Description's "Testing" heading:

> Adds ability to load content asynchronously in Modals easily. No impact on existing implementations; Verify the following modals continue to behave as expected:
>   - [ ] Website > Navigation > + Add New Navigation Menu
>   - [ ] Website > Navigation > + Add (Button on a navigation panel) > Add Product

Then, once this PR merges into `develop` just copy and paste this same testing instruction into the description of the Next version bump PR linked above following the pattern of the others there (you'll see your work alongside the others we've merged in there so far for this effort). Whomever of us UXD that runs the week's version bump eventually grabs all these and puts them in the `products` PR for QA to use to check/validate the version bump is safe. 

Again, many thanks for this! Should be a great simplification of a common need!

QuintonJason (approved):
Very nice addition

joedakroub (approved):
Beautiful! Thanks for adding this @sdhull ! Nicely done. 👍🏼

docs/app/views/examples/objects/modal/_props.html.erb
```diff
@@ -52,6 +52,12 @@
   <td><%= md('Boolean') %></td>
   <td><%= md('`nil`') %></td>
 </tr>
+<tr>
+  <td><%= md('`remote_url`') %></td>
+  <td><%= md('Specify a url which from which to load the modal content on open. This will be requested & content replaced every time the modal is opened.') %></td>
+  <td><%= md('Boolean') %></td>
```
anolson, line 58: Believe this should be `String`, right?

sdhull: gah, thanks 🤦

PR #399, opened 2021-04-01
fix(sagesearch): adds option to not execute search on clear

darrellrivera:
## Description
JIRA: [TME-959](https://kajabi.atlassian.net/browse/TME-959)
This adds the option to add `disable_search_on_clear` attribute on the SageSearch component so that the search does not run when clearing.

### Screenshots
after

## Test notes
1.) On a rails `SageSearch` component, add `disable_search_on_clear: true` to the attributes
2.) Type in anything in the search field, and press enter (so that search executes)
3.) Click on the `X` to clear the search field, but you should notice that the search does not execute (same results as before should still be appearing)

### Estimated impact
Adds ability to suppress searching again when search terms are cleared. No impact on existing implementations; verify the following search components continue to behave as expected:

- [ ] Go to: `/admin/sites/SITE_ID/email_campaigns` and clearing the search field should execute the search
- [ ] Go to: `/admin/sites/SITE_ID/products` and clearing the search field should execute the search

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_search.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_search.html.erb
@@ -14,9 +14,8 @@
   </label>
   <input
     id="<%= component.id %>"
-    class="sage-search__input
-    <%= "sage-search--disable_search_on_clear" if component.disable_search_on_clear %>
-    "
+    class="sage-search__input"
+    <%= "data-js-disable-search-on-clear=true" if component.disable_search_on_clear %>
     type="search"
     placeholder="<%= component.placeholder.present? ? component.placeholder : "Search" %>"
     value="<%= component.value %>"
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_search.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_search.html.erb
@@ -14,7 +14,9 @@
   </label>
   <input
     id="<%= component.id %>"
-    class="sage-search__input"
+    class="sage-search__input
+    <%= "sage-search--disable_search_on_clear" if component.disable_search_on_clear %>
+    "
     type="search"
     placeholder="<%= component.placeholder.present? ? component.placeholder : "Search" %>"
     value="<%= component.value %>"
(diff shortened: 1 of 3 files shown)


docs/lib/sage_rails/app/views/sage_components/_sage_search.html.erb
```diff
@@ -14,7 +14,9 @@
   </label>
   <input
     id="<%= component.id %>"
-    class="sage-search__input"
+    class="sage-search__input
+    <%= "sage-search--disable_search_on_clear" if component.disable_search_on_clear %>
```
philschanely, line 18: Could we switch to use `data-js-disable-search-on-clear="true"` here so as to reserve the class for styling?

packages/sage-system/lib/search.js
```diff
@@ -5,6 +5,7 @@ Sage.search = (function() {
 
   const SELECTOR_CLEAR_BUTTON = 'data-js-search-clear';
   const CLASS_HAS_TEXT = 'sage-search--has-text';
+  const DISABLE_SEARCH_ON_CLEAR = 'sage-search--disable_search_on_clear'
```
philschanely, line 8: As noted above, lets switch to using a `data-js` attribute here and its use below.

philschanely (commented):
Hello @darrellrivera and thanks for adding this feature! Excited for folks to be contributing ideas like this here! Couple of comments below and a note for tidying the testing steps.

What we've been doing lately even when our work should not cause any issues on any existing implementations is to provide an example or two of the affected component in use for QA to validate no impact. So in this case you could add something like this under your Description's "Estimated impact/testing" heading:

> Adds ability to suppress searching again when search terms are cleared. No impact on existing implementations; verify the following modals continue to behave as expected:
>   - [ ] (list ideally two specific example(s) from `kajabi-products` in ways a QE could find and verify existing behavior persists)

Then, once this PR merges into `develop` just copy and paste this same testing instruction into the description of the Next version bump PR linked above following the pattern of the others there (you'll see your work alongside the others we've merged in there so far for this effort). Whomever of us UXD that runs the week's version bump eventually grabs all these and puts them in the `products` PR for QA to use to check/validate the version bump is safe.

darrellrivera:
@philschanely thanks for the help on this! I've updated the code and also updated the `Estimated impact/testing` section.  Let me know if there is anything else 😄

philschanely (approved):
Looks good, @darrellrivera. Thanks for this!!

PR #400, opened 2021-04-01
Sortable Item Custom - added new rails subcomponent

QuintonJason:
## Description

This PR is intended to set a pattern when creating a custom panel. The aim is to reduce some of the rigidity with layouts being tightly coupled in the component. In some cases, the design pattern of the component already exists within our [Grid Templates](http://sage-design-system.kajabi.com/pages/layout/grid_templates). Interested in feedback on this approach to add flexibility while still confined to our sage principles

### Screenshots
Screen Shot 2021-04-01 at 3 46 03 PM

## Test notes

1. Visit the Sortable Item Rails page: http://localhost:4000/pages/object/sortable
2. Check the `Using a custom grid template` section and verify

### Estimated impact

- [ ] #400 (**LOW**) This is a new component that does not currently exist within the app.

### DIFF
--- a/docs/app/views/examples/objects/sortable/_preview.html.erb
+++ b/docs/app/views/examples/objects/sortable/_preview.html.erb
@@ -119,42 +119,41 @@
 <h3 class="t-sage-heading-6">Using a custom <code>grid template : "te"</code></h3>
 <%= sage_component SageSortable, { resource_name: "link" } do %>
   <% 3.times do %>
-    <%= sage_component SageSortableItemCustom, {
+    <%= sage_component SageSortableRow, {
+      grid_template: "te",
       id: "id",
       url_update: "update/endpoint",
     } do %>
-      <%= sage_component SageCardRow, { grid_template: "te" } do %>
-        <a href="#">Field 1</a>
-        <%= sage_component SageDropdown, {
-          align: "right",
-          items: [
-            {
-              value: "Edit",
-              icon: "pen",
-              attributes: {},
-            },
-            {
-              value: "Preview",
-              icon: "preview-on",
-              attributes: {},
-            },
-            {
-              value: "Delete",
-              style: "danger",
-              border_before: true,
-              modifiers: ["border-before"],
-              icon: "trash",
-              attributes: {},
-            },
-          ]
-        } do %>
-          <%= sage_component SageButton,  {
-            style: "secondary",
-            subtle: true,
-            icon: { name: "dot-menu-horizontal", style: "only" },
-            value: "edit",
-          } %>
-        <% end %>
+      <a href="#">Field 1</a>
+      <%= sage_component SageDropdown, {
+        align: "right",
+        items: [
+          {
+            value: "Edit",
+            icon: "pen",
+            attributes: {},
+          },
+          {
+            value: "Preview",
+            icon: "preview-on",
+            attributes: {},
+          },
+          {
+            value: "Delete",
+            style: "danger",
+            border_before: true,
+            modifiers: ["border-before"],
+            icon: "trash",
+            attributes: {},
+          },
+        ]
+      } do %>
+        <%= sage_component SageButton,  {
+          style: "secondary",
+          subtle: true,
+          icon: { name: "dot-menu-horizontal", style: "only" },
+          value: "edit",
+        } %>
       <% end %>
     <% end %>
   <% end %>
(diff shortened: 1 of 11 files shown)


docs/app/views/examples/objects/sortable/_preview.html.erb
```diff
@@ -114,3 +114,107 @@
...
+
+
+<h3 class="t-sage-heading-6">Using a custom <code>grid template : "te"</code></h3>
+<%= sage_component SageSortable, { resource_name: "link" } do %>
+  <% 3.times do %>
+    <%= sage_component SageSortableItemCustom, {
+      id: "id",
+      url_update: "update/endpoint",
+    } do %>
+      <%= sage_component SageCardRow, { grid_template: "te" } do %>
```
AndrwM, line 126: Digging this solution! I wonder if it would make more sense to marry the concept of `SageCardRow` into `SageSortableItemCustom` as `SageSortableItemRow`.

It would follow a similar naming pattern to `SagePanelRow` and `SageCardRow`.

```
<%= sage_component SageSortable, { resource_name: "link" } do %>
  <%= sage_component SageSortableItemRow, { grid_template: "te" } do %>
  <% end %>
<% end %>
```

AndrwM (commented):
LGTM, left some remarks on the naming. Wishing now that we had picked a more specific name than `SageSortableItem` for sortable children components in `SageSortable` at the time to make this naming schematic a bit easier on ourselves. (Pretty sure this was my doing originally).

👍  Curious what @philschanely thinks.

joedakroub (approved):
Overall, this approach seems like the correct one. I think @AndrwM proposed a great naming scheme as well...which looks like you just refreshed with the new names. 👍🏼

packages/sage-assets/lib/stylesheets/patterns/objects/_sortable.scss
```diff
@@ -57,6 +58,10 @@ $-sortable-image-height: rem(48px);
     }
   }
 
+  &.sage-sortable__row {
+    display: flex;
```
monkeypox8, line 62: This looks good aside from Safari, which doesn't support flex + `grid-gap`:

safari-grid-gap

docs/lib/sage_rails/app/views/sage_components/_sage_sortable_row.html.erb
```diff
@@ -0,0 +1,13 @@
+<section
```
monkeypox8, line 1: Per our discussion, wondering if `section` should be replaced with a generic div here since we're not able to reference a heading in the custom content. Or better yet, we can make it an actual `li` inside a parent ordered/unordered list.

monkeypox8 (commented):
Looking great! Couple of items to check out.

docs/app/views/examples/objects/sortable/_props.html.erb
```diff
@@ -4,3 +4,12 @@
   <td><%= md('String') %></td>
   <td><%= md('`nil`') %></td>
 </tr>
+<tr>
+  <td colspan="4"><%= md('**Sortable Row**') %></td>
+</tr>
+<tr>
```
philschanely, line 10: Should we be documenting the `url_update` and `id` props here too?

docs/lib/sage_rails/app/views/sage_components/_sage_sortable_row.html.erb
philschanely: I'm in favor of this too. I'm afraid we're over-using the semantically opinionated `section` tags in general in our components ATM.

docs/lib/sage_rails/app/views/sage_components/_sage_sortable_row.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  class="
+    sage-sortable__item
+    sage-sortable__row
+  "
+  data-js-sortable-update-url="<%= component.url_update %>"
+  <%= "id=#{component.id}" if component.id %>
+  <%= component.generated_html_attributes.html_safe %>
+>
+  <%= sage_component SageCardRow, { grid_template: component.grid_template } do %>
```
philschanely, line 10: I would prefer we avoid the double-nesting here. Any chance we could bake this affordance into the sortable row in some fashion?

philschanely (commented):
Totally cool with the idea and think this is a big win for layout flexibility for this element. See notes below for a few refinements.

docs/app/views/examples/objects/sortable/_preview.html.erb
philschanely: I will say some clarification in the documentation about when to use `SortableItem` vs. `SortableRow` would be helpful. In this way I think the `Custom` label perhaps helped a little to clarify whereas `Item` and `Row` are more ambiguous. If there's a way that the same component could be used but perhaps with a `custom_layout` flag that might be worth exploring eventually.

docs/lib/sage_rails/app/views/sage_components/_sage_sortable_row.html.erb
QuintonJason: Excellent callout. Upon investigating usage throughout the app, updating the parent to `<ul>`  and the children to `<li>` will cause no semantic issues

monkeypox8:
The `.sage-sortable` parent class could use a `padding-left` reset:

sortable-padding

docs/app/views/examples/objects/sortable/_preview.html.erb
```diff
@@ -114,3 +114,111 @@
     } %>
   <% end %>
 <% end %>
+
+
+<%= md('
+`SageSortableRow` should be used when you want to create a layout within the Sortable Item that follows any pattern using specified dot and dash pattern denoted in the [Sage Grid Templates](http://sage-design-system.kajabi.com/pages/layout/grid_templates) docs. It its simpliest form, `SageSortableRow` is `SageSortableItem` as cards with no image or subtitles, but any combination Sage Grid Template.
```
monkeypox8, line 120: Small nitpick here "At its simplest…"

packages/sage-assets/lib/stylesheets/core/mixins/_utilities.scss
```diff
@@ -152,3 +152,22 @@
   white-space: nowrap;
   border: 0;
 }
+
+///
+/// Media Queries to target Safari browser
+///
+@mixin target-safari {
```
monkeypox8, line 159: 🎯

monkeypox8 (approved):
👍🏼  LGTM, pending the modification to the `ul` padding. One teeny tiny spelling edit nitpick as well, but that's not a blocker. Great work!

docs/lib/sage_rails/app/views/sage_components/_sage_sortable_row.html.erb
philschanely: I rest this concern for now. can address in a deeper look at this component and its uses.

philschanely (approved):

docs/app/views/examples/objects/sortable/_preview.html.erb
```diff
@@ -114,3 +114,111 @@
     } %>
   <% end %>
 <% end %>
+
+
+<%= md('
+`SageSortableItemCustom` should be used when you want to create a layout within the Sortable Item that follows any pattern using specified dot and dash pattern denoted in the [Sage Grid Templates](http://sage-design-system.kajabi.com/pages/layout/grid_templates) docs. At its simpliest form, `SageSortableItemCustom` is `SageSortableItem` as cards with no image or subtitles, but any combination Sage Grid Template.
```
philschanely, line 120: ❤️

packages/sage-assets/lib/stylesheets/patterns/objects/_card.scss
```diff
@@ -135,6 +135,10 @@
 
 .sage-card__row {
   @include sage-grid-card-row();
+
+  .sage-sortable__item--custom & {
```
philschanely, line 139: 😍

philschanely (approved):
My concerns are addressed. Well done!

PR #403, opened 2021-04-02
Jcm tab examples

cmcfadzean:
## Description
Updating the example Pages to utilize tabs. 
This is in alignment with the end direction that design has created. 

### Screenshots
Screen Shot 2021-04-02 at 1 13 17 PM
Screen Shot 2021-04-02 at 1 13 26 PM
Screen Shot 2021-04-02 at 1 13 34 PM


pixelflips (approved):
This looks great and I really like how it reduces the information overload when checking out the examples! Left a couple of non-blocking comments but overall a great update. 👍🏼  

**Possible Follow-up Consideration:**
I did notice that due to the smaller overall panel sizes, increased padding and now placing examples within the tabs that some components are starting to wrap due to the decreased overall sizing of their containers as a whole. Would we be open to possibly using a larger panel (`--md` vs. `--sm`) or maybe reducing the `5rem` padding to provide a little more room on larger screens?  This also brings things more in line with the app and real-world usage unless the plan might be to use smaller widths in the app at some point in the future.

A few examples: 
- http://localhost:4000/pages/object/catalog_item
- http://localhost:4000/pages/object/modal
- http://localhost:4000/pages/object/pagination
- http://localhost:4000/pages/object/panel_controls
- http://localhost:4000/pages/design/icon

monkeypox8 (approved):
👍🏼  LGTM, this looks and works great! I think we may need improvements on the mobile styling of tab navigation items, but that's well beyond the scope here.

Sidenote: will the quick links sidebar be returning? And if not, should the commented code be removed?

PR #404, opened 2021-04-04
Jcm homepage

cmcfadzean:
### Screenshots

|  Before  |  After  |
|--------|--------|
|||

### Estimated impact

- [ ] (**LOW/MEDIUM/HIGH/BREAKING** impact level) Description of change and examples of use. Address to cross-functional groups when needed.


docs/app/views/application/_home_sections.html.erb
```diff
@@ -0,0 +1,141 @@
+<div class="docs-section">
+  <%= sage_component SageContainer, { size: "md" } do %>
+    <h2 class="docs-section__title t-sage-heading-1">Our Design System</h2>
+    <%= sage_component SageGridRow, {} do %> 
+      <%= sage_component SageGridCol, { breakpoint: "md", size: "4" } do %> 
+        <%= sage_component SagePanel, { spacer: { bottom: :lg } } do %>
+          <%= sage_component SagePanelStack, {} do %>
+            <i class="sage-icon-form-3xl t-sage--color-sage"></i>
+            <h3 class="t-sage-heading-3">Foundations</h3>
+            <p class="t-sage-body">Lorem ipsum dolor sit amet, non at set consectetur adipiscing elit. </p>
```
pixelflips, line 10: Seeing a few spots of lorem ipsum in this file.

pixelflips (approved):
👍🏼  - This looks great! Nice work! A couple of small comments below.

In addition, I've optimized the images to reduce their overall size. The new images can be downloaded here (from 389kb to around 131kb): https://drive.google.com/drive/folders/19H67RAbxz-j4k5i8Esl7salLZsqpWTas?usp=sharing

docs/app/views/application/_home_sections.html.erb
```diff
@@ -0,0 +1,141 @@
...
+        <%= sage_component SagePanel, { spacer: { bottom: :lg } } do %>
+          <%= sage_component SagePanelStack, {} do %>
+            <i class="sage-icon-form-3xl t-sage--color-sage"></i>
+            <h3 class="t-sage-heading-3">Foundations</h3>
+            <p class="t-sage-body">Principles that reflect our design philosophy and inspire our team to shape great experiences for Knowledge Commerce businesses of any scale.</p>
+            <%= sage_component SageButton, {
+              value: "Explore Foundations",
+              attributes: {
+                href: pages_foundation_path(:design_principles),
+                title: "Explore Foundations "
```
joedakroub, line 15: Not sure if this `title` value was meant to have the extra space or not.

docs/app/views/pages/index.html.erb
```diff
@@ -1,75 +1,4 @@
...
-<%= md("
-
-This system is a work of shared ownership and we encourage any member of the Kajabi team to feel empowered to participate in its ongoing improvement. Understanding and contributing to the system are the keys to the system's longevity. We encourage anyone planning on contributing to the system to first read over our Guides:
-
-[Code Guidelines](https://github.com/Kajabi/sage-lib/wiki).
-", use_sage_type: true) %>
+<%= render "application/home_hero" %>
+<%= render "application/home_sections" %>
+<%= render "application/home_developer" %>
+<%= render "application/home_code" %>
```
joedakroub, line 4: Nice cleanup! 🙌🏼

joedakroub (approved):
Look fantastic Court; nice work! Only one minor issue I found. Thank you! 👍🏼

docs/app/views/application/_home_hero.html.erb
```diff
@@ -0,0 +1,44 @@
...
+              style: "right",
+              name: "launch"
+            },
+            style: "secondary",
+          } %>
+        <% end %>
+      <% end %>
+      <%= sage_component SageGridCol, { breakpoint: "md", size: "1" } do %><% end %>
+      <%= sage_component SageGridCol, { breakpoint: "md", size: "7" } do %> 
+        <%= image_tag "sage_ui.png", class: "docs-hero__img", loading: "lazy" %>
```
monkeypox8, line 40: We'll want to either add an empty `alt` attribute here or provide a more detailed description, something like "An example of a Sage component and its HTML markup"

docs/app/views/application/_nav.html.erb
```diff
@@ -0,0 +1,95 @@
+<div class="docs-nav">
+  <%= sage_component SageButtonGroup, { gap: :sm } do %>
+    <%= sage_component SageButton, {
+      value: "Home",
+      attributes: {
+        href: "/",
+        title: "Home"
```
monkeypox8, line 7: 👍🏼

docs/app/views/layouts/home.html.erb
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <%= render "application/meta" %>
+    <%= render "application/styles" %>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/devicons/devicon@master/devicon.min.css">
```
monkeypox8, line 6: This works for now, though it is a render-blocking resource. I'm not sure whether it'd be worth applying a [deferred rendering approach](https://web.dev/defer-non-critical-css/) since it's so small. Hopefully someone more knowledgeable about that can weigh in.

As a follow-up, I think it'd be 😎  to have these icons added as local SVGs rather than an external resource, since we're loading the entire stylesheet for just the 4 or so that we use on the site. That's also assuming we can decouple the embedded fonts/assets in the CSS.

docs/app/views/application/_home_hero.html.erb
```diff
@@ -0,0 +1,44 @@
...
+        <%= image_tag "sage_icon.png", style: "max-width:70px;height:auto", loading: "lazy" %>
+        <div class="sage-type">
+          <h1 class="docs-hero__title">Building with Sage</h1>
+          <p>Our system is our source of truth, providing everything you need to build the best experiences for all Kajabi entrepreneurs.</p>
+        </div>
+        <%= sage_component SageButtonGroup, { gap: :sm, spacer: { top: :lg } } do %>
+          <%= sage_component SageButton, {
+            value: "Get Started",
+            attributes: {
+              href: pages_element_path(sorted_sage_elements.first[:title]),
```
monkeypox8, line 17: Question for the group: should this link take the visitor directly to the elements, or more of the basics such as design/tokens? What location might be best suited for someone starting out building with the system?

monkeypox8 (commented):
Really digging this new layout! Left a few notes on cleanup items to consider. I'd suggest also running the images through [imageoptim](https://imageoptim.com/mac) or https://squoosh.app/, to compress them down as much as possible.

pixelflips:
Very minor but the images could still be reduced a bit. See the link in my comment above to download. The 316kb can be reduced to around 131kb.

docs/lib/sage-frontend/stylesheets/docs/_home.scss
```diff
@@ -0,0 +1,65 @@
...
+  padding: sage-spacing();
+  padding-bottom: rem(100px);
+
+  @media (max-width: sage-breakpoint(md-min)) {
+    padding-bottom: sage-spacing(xl);
+  }
+
+  .sage-btn--subtle.sage-btn--secondary {
+    color: #fff;
+  }
```
monkeypox8, line 22: Something that I overlooked previously is that we don't have either a `hover` or `focus` style on these links.

If we don't have a design to go off of, I think at the very least we'll want to add a `focus` outline to emulate the existing styles. So, assuming a new class is added:

```
.docs-nav__link {
  &:focus::after {
    color: currentColor;
  }
}
```

docs/lib/sage-frontend/stylesheets/docs/_home.scss
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.docs-hero__title {
+  font-size: 4rem;
+  line-height: 4.5rem;
+  margin-bottom: sage-spacing(sm);
+  color: #fff;
```
monkeypox8, line 29: No need to set color here since it will inherit from `docs-hero`

docs/lib/sage-frontend/stylesheets/docs/_home.scss
```diff
@@ -0,0 +1,65 @@
...
+  @media (max-width: sage-breakpoint(md-min)) {
+    display: none;
+  }
+}
+
+.docs-section {
+  background: sage-color(gray, 200);
+  padding: sage-spacing(2xl) sage-spacing();
+
+  i {
```
monkeypox8, line 46: Our overall preference here would be to use BEM on this guy rather than using element selectors. Something like `docs-section__icon` in this case.

docs/lib/sage-frontend/stylesheets/docs/_home.scss
```diff
@@ -0,0 +1,65 @@
...
+    margin-bottom: sage-spacing();
+  }
+  
+  .devicon {
```
monkeypox8, line 50: Same as above here, possibly `docs-section__logo` or `docs-section__platform-icon`?

docs/lib/sage-frontend/stylesheets/docs/_home.scss
```diff
@@ -0,0 +1,65 @@
...
+  background-color: sage-color(primary, 500);
+  color: #fff;
+  padding: sage-spacing();
+  padding-bottom: rem(100px);
+
+  @media (max-width: sage-breakpoint(md-min)) {
+    padding-bottom: sage-spacing(xl);
+  }
+
+  .sage-btn--subtle.sage-btn--secondary {
```
monkeypox8, line 20: I'm not sure why it's not in the docs, but rather than reference the `.sage-btn` styles directly, I'd recommend using the `SageButton` `css_classes` prop on these links to add a BEM class, such as `docs-nav__link`. Then this style can move into the `docs/_nav.scss` file.

docs/lib/sage-frontend/stylesheets/docs/_nav.scss
```diff
@@ -0,0 +1,25 @@
+/* ==================================================
+  ** _nav.scss
+
+  For Sage documentation use
+================================================== */
+
+.docs-nav {
+
+  .sage-btn::before {
```
monkeypox8, line 9: Continuing my note above about adding a custom class. Ideally this would be `docs-nav__link` as well.

monkeypox8 (approved):
👍🏼 LGTM, added some comments, mostly dealing with our BEM expectations. I think the only item I'd consider a blocker would be the focus styling.

PR #407, opened 2021-04-06
[TME-959] Fix data-kjb-disable-search placement

darrellrivera:
## Description
JIRA: [TME-959](https://kajabi.atlassian.net/browse/TME-959)
In SageSearch, we added the attribute of `data-kjb-disable-search` to  `<form>` to "Prevent the default Kajabi-Products Search.js from binding to this input"; however, in [products' search](https://github.com/Kajabi/kajabi-products/blob/master/app/assets/javascripts/search.js#L1), we are checking for `data-kjb-disable-search` on an `input` instead. 

### Screenshots
n/a

## Test notes
This should be tested against products
1.) Go to https://www.kajabi.test/admin/sites/1/landing_pages?website_page=all, and type `Home`
2.) Click outside of the text input and it should no longer auto submit
3.) Go back to the search input and press enter, and it should now submit.

### Estimated impact
Fixes bug where in products, any page that was using the `SageSearch` component was autosubmitting when it should not have been (users will now have to press enter to submit the form)
- [ ] Go to `https://www.kajabi.test/admin/sites/1/products` and search for a product.  It should not execute query until you press enter

### DIFF
--- a/docs/lib/sage_rails/app/sage_components/sage_search.rb
+++ b/docs/lib/sage_rails/app/sage_components/sage_search.rb
@@ -5,6 +5,5 @@ class SageSearch < SageComponent
     label_text: [:optional, String],
     placeholder: [:optional, String],
     value: [:optional, String, NilClass],
-    disable_search_on_clear: [:optional, TrueClass],
   })
 end
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_search.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_search.html.erb
@@ -5,8 +5,6 @@
   "
   data-js-search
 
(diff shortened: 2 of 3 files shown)


philschanely (approved):
No concerns here. Did we also mean to remove the conditionality here? AKA are we sure we want this on always?

darrellrivera:
@philschanely yes.  it seems like this was always on (but just on the wrong element).  i also just reverted the work i did [here](https://github.com/Kajabi/sage-lib/pull/399) because I thought that was the initial fix but the search.js file here already takes care of not executing search correctly when canceling

philschanely:
@darrellrivera Ah ok!

PR #418, opened 2021-04-13
feat(catalog-item): gives the Catalog Item Optional Image and Icon

cmcfadzean:
## Description
The Catalog image might not always need to be included in the catalog item so this makes it optional
and adds a formatting class
The Catalog item also might want to have an icon instead of an image. This supports this. 

### Screenshots
Screen Shot 2021-04-13 at 10 48 43 AM
Screen Shot 2021-04-13 at 10 48 59 AM
Screen Shot 2021-04-13 at 11 07 33 AM

## Test notes
Create A Catalog Item with an included image URL. The image will show. Do not include a URL and no image will show.

### DIFF
--- a/docs/app/views/examples/objects/catalog_item/_preview.html.erb
+++ b/docs/app/views/examples/objects/catalog_item/_preview.html.erb
@@ -1,4 +1,4 @@
-<% 3.times do %>
+<% 2.times do %>
   <%= sage_component SageCatalogItem, {
     title: "Fun For All – The Children Call: Their Favorite Time Of Year",
     image: "/assets/card-placeholder-sm.png",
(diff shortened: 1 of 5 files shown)


pixelflips (approved):
LGTM! 👍🏼  This is great and working as expected. Two small nitpicks to consider:

1. Should we have an example of an item with no image or icon in the preview?
2. I noticed that the CatalogItems are contained in `<li>` elements, but seem to be missing the parent `<ul>` or `<ol>`. See screenshot below.

Screen Shot 2021-04-13 at 1 39 11 PM

joedakroub (approved):
These changes look good to me. Thanks for the docs updates too! 👍🏼

PR #419, opened 2021-04-14
Tooltip - increase z-index

QuintonJason:
## Description

Increase the z-index to be higher than modals in the case a modal has a tooltip. This will not cause an adverse affect if the tooltip is under the modal since it only shows on hover

### Screenshots

|  Before  |  After  |
|--------|--------|
|Screen_Shot_2021-04-14_at_7_26_21_AM|Screen_Shot_2021-04-14_at_7_26_37_AM|

## Test notes

To test the existing implementation follow the steps from the video in this ticket, [BUILD-995](https://kajabi.atlassian.net/browse/BUILD-995)
- Visit podcast distribution page
- Expand one of the vendor accordians
- Click on `View Steps`
- Hover the copy button to the right of the `RSS Feed URL` and notice the tooltip
**or **
On the modal's preview page, temporarily include a tooltip within one of the `SageModalContent`s to test

### Estimated impact

- (**MEDIUM** impact level) This change increases the `z-index` for all tooltips to resolve them hiding behind modals.
- [ ] Podcast Distribution page

## Related

[BUILD-995](https://kajabi.atlassian.net/browse/BUILD-995)

### DIFF
--- a/docs/app/views/examples/objects/modal/_preview.html.erb
+++ b/docs/app/views/examples/objects/modal/_preview.html.erb
@@ -54,11 +54,6 @@
         sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
       </p>
 
-      <%= sage_component SageTooltip, {
-        position: "left",
-        text: "Bibendum Condimentum",
-      } %>
-
       <% content_for :sage_footer do %>
         <% content_for :sage_footer_aside do %>
           <%= sage_component SageButton, {
--- a/docs/app/views/examples/objects/modal/_preview.html.erb
+++ b/docs/app/views/examples/objects/modal/_preview.html.erb
@@ -54,6 +54,11 @@
         sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
       </p>
 
+      <%= sage_component SageTooltip, {
+        position: "left",
+        text: "Bibendum Condimentum",
+      } %>
+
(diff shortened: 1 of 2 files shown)


monicawheeler (approved):

joedakroub (approved):
Nice catch; thanks Quinton! 👍🏼

monkeypox8 (commented):
LGTM. At one point in the recent past, we had advocated incrementing by no less than a factor of `100` (ex. `sage-z_index(modal, 100)`) except to prevent a `z-index` collision. In those cases incrementing by `10` would be acceptable. This can still be seen in the "old" repo's wiki, but I'm not sure whether we want to keep it as a standard moving forward: https://github.com/Kajabi/sage/wiki/Z-index-Scale

QuintonJason:
> LGTM. At one point in the recent past, we had advocated incrementing by no less than a factor of `100` (ex. `sage-z_index(modal, 100)`) except to prevent a `z-index` collision. In those cases incrementing by `10` would be acceptable. This can still be seen in the "old" repo's wiki, but I'm not sure whether we want to keep it as a standard moving forward: https://github.com/Kajabi/sage/wiki/Z-index-Scale

Excellent callout! There's no reason to change the rule. Updated and thanks for reiterating the wiki reference.

monkeypox8 (approved):
👍🏼  🔥

PR #423, opened 2021-04-20
Ensure expandable card content is truly hidden by default

philschanely:
## Description

Current implementations of this component were exhibiting a temporary flash of the panel content. This PR ensures the content is set to hide by default. JS or React will then enable it to be toggled on as appropriate. 

No visual difference.

## Test notes

See Expandable Card in Rails and React to see no visual or functional difference other than absence of possible flash of content.

### Estimated impact
(LOW) Removes undesired flash of content in expandable card component. No adverse effects expected. Verify the following behave as expected but without the initial flash of content:
   - [ ] New custom domain flow "Need help" toggles
   - [ ] (if relevant) Old custom domain flow step "5. Setup Page Rules & Orange Cloud" toggles


monkeypox8 (approved):
👍🏼 LGTM!

QuintonJason (approved):

PR #424, opened 2021-04-21
Panel controls: toggle default toolbar

philschanely:
## Description

This PR adds a new configuration on Panel Controls in React that allows the "default toolbar" to be easily toggled off and on such as to hide it when there are no items to show (and thus no reason to show pagination, bulk actions, sort, etc.)

## Test notes

See the [Panel Controls component in Storybook](http://localhost:4100/?path=/story/sage-panelcontrols--default). You should now see both a new note about needing to configure the Newscatcher API and an updated "empty state" view in lieueu of a bunch of disabled controls. If you add a Rapid API key as indicated in `packages/sage-react/lib/services/newsapi.js` the api should activate and show the table results as it did in the past. (reach out to Phil if you need a demo api key; it is excluded from file source for security). Then if you change the search terms `q` param inside `getNews` to something like `'glup'` the API should return no results and show the empty state again. 

### Estimated impact
(BREAKING) Updates React Panel Controls with new feature to easily toggle the Default Toolbar (bulk actions, sort, pagination) off and on. [This PR](https://github.com/Kajabi/kajabi-products/pull/18717) provides the patch to restore existing behavior to the table as well as to toggle this off and on in response to search and filer requests.
   - [ ] Verify the People contacts table behaves as expected, as well as now hides the bulk actions, pagination, and sort controls when there are no results to a search or filter.


monkeypox8 (approved):
👍🏼 LGTM, working as expected. Took me a while to figure out I needed to have `use_async_contacts_ui` enabled to get the proper view 😆

PR #427, opened 2021-04-22
Jcm example cleanup

cmcfadzean:
## Description
Some simple clean-up on the code examples for the elements and objects. This is just a small step toward more clear and consistent code examples and documentation.

Screen Shot 2021-04-22 at 2 47 16 PM

### DIFF
--- a/docs/app/views/examples/elements/breadcrumbs/_preview.html.erb
+++ b/docs/app/views/examples/elements/breadcrumbs/_preview.html.erb
@@ -1,9 +1,7 @@
-<%= sage_component SageCardBlock, {} do %>
-  <%= sage_component SageLabel, {
-    color: "draft",
-    value: "Default"
-  } %>
-<% end %>
+<!-- Docs Only Label -->
+<%= sage_component SageCardBlock, {} do %><%= sage_component SageLabel, {color: "draft",value: "Default"} %><% end %>
+<!-- Docs Only Label End -->
+
 <%= sage_component SageBreadcrumbs, {
   items: [
     {
@@ -21,12 +19,10 @@
   ]
 } %>
 
-<%= sage_component SageCardBlock, {} do %>
-  <%= sage_component SageLabel, {
-    color: "draft",
-    value: "Back Link Variation"
-  } %>
-<% end %>
+<!-- Docs Only Label -->
+<%= sage_component SageCardBlock, {} do %><%= sage_component SageLabel, {color: "draft",value: "Back Link"} %><% end %>
+<!-- Docs Only Label End -->
+
 <%= sage_component SageBreadcrumbs, {
   items: [
     {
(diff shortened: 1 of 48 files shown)


docs/app/views/examples/elements/breadcrumbs/_preview.html.erb
```diff
@@ -1,9 +1,7 @@
-<%= sage_component SageCardBlock, {} do %>
-  <%= sage_component SageLabel, {
-    color: "draft",
-    value: "Default"
-  } %>
-<% end %>
+<!-- Docs Only Label -->
+<%= sage_component SageCardBlock, {} do %><%= sage_component SageLabel, {color: "draft",value: "Default"} %><% end %>
```
monkeypox8, line 2: I'm reasonably sure this was intended to be primarily a visual change, but I'd recommend keeping the previous heading structure intact for grouping. We do need to correct much of our heading hierarchy across the docs site (we skip `h2` on most of the example sections, for example), but the change here to a label would appear to remove the association between the description and example.

For example, in firefox's accessibility inspector for "Form Input", the version with labels appears as disassociated text sections due to the card structure:

|  current   |  proposed  |
|--------|--------|
|before|after|

docs/app/views/examples/elements/button/_preview.html.erb
```diff
@@ -244,3 +226,14 @@ demo_configs = [
     raised: true,
   } %>
 <% end %>
+
+<!-- Docs Only Note -->
+<%= sage_component SageAlert, {
+  color: "info",
+  title: "Note:",
+  desc: "Regular buttons allow for a 'raised' shadow treatement to be toggled on or off.
+  This is on by default for 'rimary' button styles and can be toggled off with 'no_shadow'.
```
monkeypox8, line 235: Missing "P" in "Primary" here

docs/app/views/examples/elements/breadcrumbs/_preview.html.erb
philschanely: I agree here and was wondering about specifying a tag for the label if we want the visual styling but the semantic tag. I hope to get a PR up early next week with a proposal for breaking apart things a bit more too per our discussion elsewhere.

docs/CHANGELOG.md
```diff
@@ -3,6 +3,35 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [3.22.0](https://github.com/Kajabi/sage-lib/compare/@kajabi/sage@3.21.0...@kajabi/sage@3.22.0) (2021-04-21)
```
philschanely, line 6: 🤔 do we know why these changes got included? Perhaps this needs to be rebased against `develop`?

PR #429, opened 2021-04-23
React unit testing

philschanely:
## Description

This PR adds boilerplate unit testing files for all React components and some developer tools to aid the test development process.

(May need to run `yarn lerna bootstap` from root to update to new packages here.)

Use the new `yarn test:dev` script from within the `packages/sage-react` directory to fire up a live reload server and Jest testing in watch mode. You'll then see this coverage report in the browser:

Screen Shot 2021-04-23 at 3 30 59 PM

You can click into any folder for more details:

Screen Shot 2021-04-23 at 3 31 23 PM

And then a specific file will show coverage line by line:

Screen Shot 2021-04-23 at 3 31 32 PM

### New to Jest Unit testing?

This is a neat way to add stability and checks to React components and JS files. [Learn more about Jest](https://jestjs.io/). Guidance to be provided for developing and maintaining Jest tests here in Sage.

Some key concepts:

- Tests: written in `*.spec.jsx` files using the Jest `describe`, `it`, and `expect` syntax. The goal should be to create tests that validate expectations for how a component should be constructed, varied, and change on interaction. When you run `yarn test` you will get a report of tests run and their statuses:

  Screen Shot 2021-04-23 at 3 34 01 PM

  Ironically, testing can be subjective in terms of what counts as a valid test. There is definitely an art to master with it. As we set out we may want to aim for a mixture of tests that provide coverage (see below) and tests that validate what we know to be the core aspects of a given component.

- Coverage: one metric beyond just passing tests is to gauge how much of the code is "covered" by tests. Coverage provides such a metric, as seen in the screenshots above. This evaluates lines covered, and even breaks out into functions and branches of code covered! While 100% coverage in all categories may be ideal, 80% may be a realistic benchmark to aim for.

## Test notes
(for reals, yo?)

### Estimated impact
(LOW) Added unit testing framework for Sage React. No impact; development tool only


philschanely:
@voodooGQ Any idea why this has caused CI to freeze up?

monkeypox8 (approved):
👍🏼 LGTM and thanks for the jest 101 runthrough!

Would there be value in having the new scripts also be applied to the main `package.json`? For example, `yarn test:dev` is as far as I know only accessible within the react dir.

PR #434, opened 2021-04-27
Rails Components generated_css_classes to all remaining components

QuintonJason:
## Description

~This is a followup PR to: https://github.com/Kajabi/sage-lib/pull/432~. Since we'll be adding more flexibility to what can be styled in `generated_css_classes`, we want to be sure it's on all components. 

##### Excluded components
- `NavLink`
- `Link`

## Test notes

On any component add a `spacer` property with a corresponding value and verify that the corresponding classes are present with the appropriate styling

### Estimated impact

- [ ] (**LOW**) Adds `generated_css_classes` where currently missing so this should have no visual affect

## Related

https://github.com/Kajabi/sage-lib/pull/432


docs/lib/sage_rails/app/views/sage_components/_sage_avatar.html.erb
```diff
@@ -3,7 +3,7 @@
     sage-avatar
     <%= "sage-avatar--#{component.color}" if component.color.present? %>
     <%= "sage-avatar--centered" if component.centered.present? && component.centered %>
-    <%= component.css_classes %>
```
philschanely, line 6: Probably shouldn't be deleted yet here.

docs/lib/sage_rails/app/views/sage_components/_sage_body.html.erb
```diff
@@ -1,3 +1,7 @@
-<body class="sage-page <%= component.css_classes if component.css_classes.present? %>" <%= component.generated_html_attributes.html_safe %>>
```
philschanely, line 1: probably shouldn't delete `css_classes` here yet.

docs/lib/sage_rails/app/views/sage_components/_sage_dropdown.html.erb
```diff
@@ -1,12 +1,13 @@
 <div
-  class="sage-dropdown
-    <%= "sage-dropdown--anchor-right" if component.align == "right" %>
+  class="sage-dro    <%= "sage-dropdown--anchor-right" if component.align == "right" %>pdown
```
philschanely, line 2: Looks like an accidental copy paste error here.

docs/lib/sage_rails/app/sage_components/sage_schema_helper.rb
```diff
@@ -12,6 +12,11 @@ module SageSchemaHelper
     "ste", "sti", "sts", "sme", "smi", "sms", "soe", "soi", "sos",
   ])
 
+  SAGE_ALIGN = {
+    align_items: [:optional, Set.new([:"flex-start", :"flex-end", :center, :baseline, :stretch, "flex-start", "flex-end", "center", "baseline", "stretch"])],
```
philschanely, line 16: This is flex-focused ATM but may also be useful with `grid` setups. Only a few are different such as `start` vs. `flex-start`, etc. Would you consider adding grid-compatible options here too? or perhaps we need separate options for grid align versus flex align? 🤔

philschanely (commented):
Fabulous ideas here all around! 🙌  Some thoughts and questions below.

In regard to the generated_css_classes I noted a few places I caught where a component's `css_classes` was removed. I think it would be great for this to be one of the standard props on all components, fed into the generated_css_classes in some fashion, but for now we should at least be careful we're not removing that by accident unless we also replace it.

docs/lib/sage_rails/app/views/sage_components/_sage_body.html.erb
QuintonJason: It's still there for this one

philschanely (approved):
🙌

PR #433, opened 2021-04-26
Transaction card updates

pixelflips:
## Description
The transaction card design has been updated. This PR updates the component to match these changes.

- Removes dropdown
- Adds option to add a link to the `name` through `name_href` attribute
- Renames `product` element and attributes to `related_property`
- Adds option to add a link to the `related_property` through `related_property_href` attribute
- Adds `transaction_state` section (replaces dropdown)
- Adds `name_tag` (rails) `NameTag` (react) to allow for heading element to be set and not be hard-coded. Will default to h4.

### Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2021-04-26 at 12 42 00 PM|Screen Shot 2021-04-26 at 12 41 36 PM|

## Test notes

REACT:
- Visit: http://localhost:4100/?path=/story/sage-transactioncard--default
 - Verify dropdown is no longer available
 - Verify link is added to the `name`
 - Verify link is added to the `related_property`
 - Verify `transaction_state` text displays in place of dropdown
 - Verify product-related naming is updated to `related_property` (elements and attributes)
 - Adjust controls to ensure things work as expected.

RAILS: 
 - Visit http://localhost:4000/pages/object/transaction_card
 - Verify dropdown is no longer available
 - Verify link is added to the `name`
 - Verify link is added to the `related_property`
 - Verify `transaction_state` text displays in place of dropdown
 - Verify product-related naming is updated to `related_property` (elements and attributes)

### Estimated impact
- [ ] (**LOW**) Removes dropdown, adds an optional link to name and related_property, and adds new transaction_state text for Transaction Card. This component is not currently in use and these changes are being made in preparation for work to begin during the next cycle.

PR #445, opened 2021-04-29
feat(transaction card): updating the transaction card to include links

cmcfadzean:
Link for the name and a link for the offer
BREAKING CHANGE: The product title has been changed to offer title

## Description
Transaction Card can now take a URL for the Name and the Offer.
This component is not yet used anywhere so while it is a breaking change nothing should break.

### DIFF
--- a/docs/app/views/examples/objects/transaction_card/_preview.html.erb
+++ b/docs/app/views/examples/objects/transaction_card/_preview.html.erb
@@ -2,9 +2,11 @@
   label_color: "published",
   label_text: "New",
   name: "Lilly Jones",
+  name_link: "#",
   amount: "+ 450.00",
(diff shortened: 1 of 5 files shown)


pixelflips:
@cmcfadzean We should sync up. I have a similar but slightly different set of requirements in review right now.

https://github.com/Kajabi/sage-lib/pull/433

pixelflips:
@cmcfadzean I've incorporated these exact changes into https://github.com/Kajabi/sage-lib/pull/433 so I think it will be safe to close this one once 433 is approved.

pixelflips:
@cmcfadzean #433 has been approved and merged so you can close this PR when time permits. It contains the changes outlined above along with others discussed in the #433 PR.

PR #455, opened 2021-04-30
Patch item count logic with page size in Panel Controls

philschanely:
## Description

This PR makes a fix to a bug in calculating the current page range to show given recent changes to add page size.

## Testing in `sage-lib`

N/A

(**LOW**) Fixes a bug with panel controls' current display item count. 
   - [ ] Verify that pagination displays correct value for all pages in a result set on People page.


kimjaej (approved):
🚀

PR #457, opened 2021-05-03
Responsive type and type pairings

philschanely:
## Description

This PR updates the type specs to match latest from design including making both font size and line height responsive. 

See updated [Figma specs](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/Sage-3.0-WIP?node-id=3655%3A15435)

**So now when working with type settings:**

- (Preferred) Current system for applying type specs using `@extend %t-sage-[spec-name]` is unchanged and is still preferred; now it uses updated responsive values behind the scenes.
- (Next best) A new mixin `sage-type-pairing` is available for setting up type on an element to where both font size and line height are set according to system specs for such pairings. Pass a spec such as `body` to retrieve a responsive pairing, or pass a scale such as `md` and a second parameter of `false` for a non-responsive pairing.
- (Least preferred) When only font sizes or only line-heights are desired use `sage-font-size()` or `sage-line-height()` function can be used. Now, however, if a responsive setting is desired, the type spec should be provided as the key such as `sage-font-size(body)` whereas if a non-responsive value is desired use a scaled key and pass `false` as a second parameter such as `sage-font-size(md, false)`.

## Screenshots
|  Before  |  After  |
|--------|--------|
| Screen Shot 2021-05-03 at 1 32 53 PM | Screen Shot 2021-05-03 at 1 23 52 PM |

## Testing in `sage-lib`

See Typography page under Design and generally poke around to observe all aspects should remain as before.

(BREAKING) Type spec setup is updated with minimal visible adjustments; general inspection of `kajabi-products` pages that use Sage should show no negative side effects. [This PR](https://github.com/Kajabi/kajabi-products/pull/18868) will merge in a few necessary patches.

### DIFF
--- a/docs/app/views/pages/typography.html.erb
+++ b/docs/app/views/pages/typography.html.erb
@@ -4,7 +4,8 @@ type_specs = [
     style: md(%(
       <span class="t-sage-heading-1">Heading 1</span><br/>
       HTML: `t-sage-heading-1`<br/>
-      Rails: `SageClassnames::TYPE::HEADING_1`
+      Rails: `SageClassnames::TYPE::HEADING_1`<br/>
+      React: `SageClassnames.TYPE.HEADING_1`
     )),
     weight: "700",
     kerning: "0.3px",
@@ -15,7 +16,8 @@ type_specs = [
     style: md(%(
       <span class="t-sage-heading-2">Heading 2</span><br/>
       HTML: `t-sage-heading-2`<br/>
-      Rails: `SageClassnames::TYPE::HEADING_2`
+      Rails: `SageClassnames::TYPE::HEADING_2`<br/>
+      React: `SageClassnames.TYPE.HEADING_2`
     )),
     weight: "700",
     kerning: "0.3px",
@@ -26,7 +28,8 @@ type_specs = [
     style: md(%(
       <span class="t-sage-heading-3">Heading 3</span><br/>
       HTML: `t-sage-heading-3`<br/>
-      Rails: `SageClassnames::TYPE::HEADING_3`
+      Rails: `SageClassnames::TYPE::HEADING_3`<br/>
+      React: `SageClassnames.TYPE.HEADING_3`
     )),
     weight: "700",
     kerning: "0.3px",
@@ -37,7 +40,8 @@ type_specs = [
     style: md(%(
       <span class="t-sage-heading-4">Heading 4</span><br/>
       HTML: `t-sage-heading-4`<br/>
-      Rails: `SageClassnames::TYPE::HEADING_4`
+      Rails: `SageClassnames::TYPE::HEADING_4`<br/>
+      React: `SageClassnames.TYPE.HEADING_4`
     )),
     weight: "600",
     kerning: "0.3px",
@@ -48,7 +52,8 @@ type_specs = [
     style: md(%(
       <span class="t-sage-heading-5">Heading 5</span><br/>
       HTML: `t-sage-heading-5`<br/>
-      Rails: `SageClassnames::TYPE::HEADING_5`
+      Rails: `SageClassnames::TYPE::HEADING_5`<br/>
+      React: `SageClassnames.TYPE.HEADING_5`
     )),
     weight: "600",
     kerning: "0.3px",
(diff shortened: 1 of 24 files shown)


packages/sage-assets/lib/stylesheets/components/_pagination.scss
```diff
@@ -78,7 +78,6 @@ $-pagination-bg-color-dark: sage-color(grey, 300);
   align-items: center;
   min-height: rem(40px);
   padding: $-pagination-padding;
-  font-size: sage-font-size(lg);
```
philschanely, line 81: This one appeared redundant.

packages/sage-assets/lib/stylesheets/components/_toast.scss
```diff
@@ -79,6 +79,6 @@ $-toast-viewport-spacing-bottom: sage-spacing();
   margin-right: sage-spacing(xs);
   padding: 0;
   color: sage-color(white);
-  line-height: sage-line-height(2xs);
```
philschanely, line 82: This also appeared redundant given new spec settings.

packages/sage-assets/lib/stylesheets/tokens/_line_height.scss
```diff
@@ -6,25 +6,59 @@
 
 
 ///
-/// Sage line height/leading token
+/// Sage raw line height/leading token (non-responsive)
 ///
 $sage-line-heights: (
-  2xs: sage-baseline(4),
```
philschanely, line 12: This one is not in spec

monkeypox8 (approved):
❤️ Love this change. Specs match perfectly, and I'm happy to see how well this all fits together!

packages/sage-assets/lib/stylesheets/tokens/_type_specs.scss
```diff
@@ -125,10 +129,13 @@ $sage-type-specs: (
 @mixin sage-type-spec($spec-name) {
   $spec-map: map-get($sage-type-specs, $spec-name);
 
+  @include sage-type-pairing(
```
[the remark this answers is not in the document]
philschanely: @monkeypox8 Great point on the fallbacks showing up... I am in favor of this but don't know how to do it... is it in a default configuration in webpack perhaps that we just need to override?

In terms of defaults I was thinking about this too as I refactored the usages in `kajabi-products`. here in `sage-lib` a fallback is not as much of a concern to me, but over in `kajabi-products` we may want to consider fallbacks if we use these custom properties directly. We may want to avoid that though and use the mixins/functions instead. And if we do a fallback should it be to the mobile or desktop value?

docs/app/views/pages/typography.html.erb
```diff
@@ -4,152 +4,211 @@ type_specs = [
...
-    size_mobile: "16px",
-    line_height: "24px",
     kerning: "0.3px",
+    desktop: "18px / 28px",
+    mobile: "16px / 28px",
   },
   {
     style: md(%(
       <span class="t-sage-heading-6">Heading 6</span><br/>
       HTML: `<-sage-heading-6`<br/>
```
joedakroub, line 66: Looks like a typo here `<-sage-heading-6`

packages/sage-assets/lib/stylesheets/tokens/_type_specs.scss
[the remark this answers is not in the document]
monkeypox8: Oh, I see. I was wondering if it was buried in the config somehow, since the style generation was tough to source. I'm hoping it's an easy change to make in webpack—can dig into it more later this morning, but also happy to jump on a call and see what we can uncover if you'd like.

As far as the fallbacks go, I'm in favor of going all-in with allowing the use of custom props directly, since they're more easily accessed from outside of Sage on the Rails side. But we may want a more in-depth discussion within our ranks before moving forward there.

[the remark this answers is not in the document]
philschanely: @monkeypox8 I think I'll need to look at this config in a follow up.

PR #465, opened 2021-05-05
Adding resolve-url-loader to docs devDependencies for deployment

voodooGQ:
## Description
Deploys are currently broken

image

### History
`resolve-url-loader` was added to `docs/config/webpack/environment.js` as a fix in this PR: https://github.com/Kajabi/sage-lib/pull/285. This appears to work correctly in local environments because `resolve-url-loader` is available through yarn workspaces (https://classic.yarnpkg.com/en/docs/workspaces/) but still needs to be part of the `docs/package.json` if it is to be installed on deployment via Git Subtree pushes (https://github.com/Kajabi/sage-lib/blob/master/package.json#L44).

### Solution
Add `resolve-url-loader` to the `docs/package.json`

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -55,6 +55,7 @@
     "eslint-plugin-react": "^7.22.0",
     "eslint-plugin-react-hooks": "^4.2.0",
     "npm-run-all": "^4.1.5",
+    "resolve-url-loader": "^3.1.1",
     "right-pad": "^1.0.1",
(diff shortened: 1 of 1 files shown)


monkeypox8 (approved):
👍🏼

joedakroub (approved):
Looks good to me 👍🏼

pixelflips (approved):
🎉

PR #477, opened 2021-05-12
Added file-loader to deps of sage-assets

voodooGQ:
## Description
Same problem we had here: https://github.com/Kajabi/sage-lib/pull/465 except this time its the `file-loader` in `sage-assets`

image

## Solution
Added `file-loader` dependency to `sage-assets` package.json


philschanely (approved):
@voodooGQ if you want feel free to rebase this one against master so we can test triggering a new deploy?

PR #488, opened 2021-05-18
Stat Box Updates

anechol:
## Description

Fixes Stat Box styling to be more in line with Figma mockup. Data attribute has been updated to type "String."

[Stat Box in Figma](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/%5BWIP%5D-Sage-3-%E2%80%94-Admin-Components?node-id=3435%3A5)

## Screenshots

|  Before  |  After  |
|--------|--------|
statbox|Screen Shot 2021-05-18 at 3 36 16 PM

### Rails
- View [Rails component](http://localhost:4000/pages/component/stat_box)
- Check that component UI exists and nothing is broken

### React
- View [Storybook](http://localhost:4100/?path=/docs/sage-statbox--default)
- Check that component UI exists and nothing is broken

(LOW) Updates styling and data attribute. No effect on existing work in Kajabi Products.

## Related

[MAN-1385](https://kajabi.atlassian.net/browse/MAN-1385)

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_stat_box.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_stat_box.html.erb
@@ -1,14 +1,13 @@
 <% 
 change_value = component.change.present? ? component.change[:value] : "No change"
 change_type = component.change.present? ? component.change[:type] : "neutral"
-change_icon = component.change.present? ? component.change[:icon] : ""
 label_configs = { color: "draft", value: change_value } 
 if change_type == "positive"
-  label_configs = { color: "published", value: change_value, icon: change_icon } 
+  label_configs = { color: "published", value: change_value, icon: "caret-up" } 
 elsif change_type == "negative"
-  label_configs = { color: "danger", value: change_value, icon: change_icon } 
+  label_configs = { color: "danger", value: change_value, icon: "caret-down" } 
 elsif change_type == "neutral"
-  label_configs = { color: "draft", value: change_value, icon: change_icon } 
+  label_configs = { color: "draft", value: change_value, } 
 end
 %>
 <article
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_stat_box.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_stat_box.html.erb
@@ -20,7 +20,7 @@ end
       <%= component.popover.html_safe %>
     <% end %>
   </header>
-  <main class="sage-stat-box__body sage-grid-template-te">
+  <div class="sage-stat-box__body sage-grid-template-te">
     <p class="sage-stat-box__data">
       <%= component.data %>
       <% if component.timeframe.present? %>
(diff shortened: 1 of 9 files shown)


docs/lib/sage_rails/app/views/sage_components/_sage_stat_box.html.erb
```diff
@@ -1,13 +1,14 @@
...
 change_value = component.change.present? ? component.change[:value] : "No change"
 change_type = component.change.present? ? component.change[:type] : "neutral"
+change_icon = component.change.present? ? component.change[:icon] : ""
 label_configs = { color: "draft", value: change_value } 
-if change_type == "up"
-  label_configs = { color: "published", value: change_value, icon: "caret-up" } 
-elsif change_type == "down"
-  label_configs = { color: "danger", value: change_value, icon: "caret-down" } 
+if change_type == "positive"
+  label_configs = { color: "published", value: change_value, icon: change_icon }
```
philschanely, line 7: Here is the hope to allow for an icon override? Seems we may still want to provide preset icons in the conditionals below like:

`... icon: (change_icon ? change_icon : "caret-up") ...` etc.

philschanely (commented):
Great style changes! Some comments below regarding icon setup but otherwise ready to go. LMK if you'd like to discuss more in a call.

anechol:
Per brainstorming session with @philschanely, I've removed the option for a custom icon and reverted back to defaults/fallbacks. If a custom scenario or icon is required for the label, developers have the option to pass in whatever override they feel is appropriate.

monicawheeler:
Should the "After" screenshot show an arrow on the percent increase?

anechol:
Oops! Good catch @monicawheeler. Updated with new image!

docs/lib/sage_rails/app/views/sage_components/_sage_stat_box.html.erb
```diff
@@ -20,11 +20,13 @@ end
       <%= component.popover.html_safe %>
     <% end %>
   </header>
-  <main class="sage-stat-box__body sage-grid-template-ete">
-    <p class="sage-stat-box__data"><%= component.data %></p>
-    <% if component.timeframe.present? %>
-      <span class="sage-stat-box__timeframe"><%= component.timeframe %></span>
-    <% end %>
+  <main class="sage-stat-box__body sage-grid-template-te">
```
monkeypox8, line 23: Hmm, I may be mistaken but is it acceptable to use multiple `main` elements on a page and/or as a descendant of `article`? At least per the current W3C spec this has been disallowed: https://www.w3.org/TR/html52/grouping-content.html#the-main-element, but I recall it being a point of contention between the W3C and WHATWG in the past

anechol: Oh, thanks so much for catching this! I don't know why I thought I read a `<main>` element as an acceptable child of `<article>`. Looks like this isn't the case. I've changed `<main>` to `<div>`.

monkeypox8 (approved):
👍🏼 🔥  LGTM, great work!

philschanely (approved):
Looking good!

PR #533, opened 2021-05-21
Update rails and ensure rack is updated as well

bemurphy:
## Description
We need to update Rails for 2 primary reasons:

- Production is updating next week, so this should be synched with that for compatibility reasons
- Rack is vulnerable on this app, and requires either a risky monkeypatch or an upgrade, which is not possible with 4.x series Rails


philschanely (approved):
No concerns here. Assuming pulling this and running `bundle install` is sufficient to get up to date, i was able to start the server, turn on the bridge, and still see changes come through in `kajabi-products` so everything seemed 👍 

Thanks for doing this for Sage @bemurphy !

bemurphy:
thanks @philschanely for those extra steps! That sounds right to me.

The app appears setup with Heroku to autodeploy the docs as well, so that'll be the ultimate test ;) Will squash merge this shortly then and follow along with the Heroku deploy to ensure it's rendering right there as well.

PR #536, opened 2021-05-24
Modal - update spacing between adjacent p tags

QuintonJason:
## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2021-05-24 at 2 13 12 PM|Screen Shot 2021-05-24 at 2 12 57 PM|

## Testing in `sage-lib`

Visit the modal page and click on the Standard Modal demo
Verify that the spacing between the adjacent `<p>` tags matching the spec value, which is `24px`

1. (LOW) Update spacing between `<p>` tags within modal content. This affect any of the modals since there are no instances of adjacent `<p>` tags currently within any modal in the app.
   - [ ] Delete confirmation modal once sagified - Podcast episode delete episode modal
   - [ ] Delete confirmation modal text match once sagified - Site delete modal

## Related

[BUILD-1112: https://github.com/Kajabi/kajabi-products/pull/19033/

### DIFF
--- a/docs/app/views/examples/components/modal/_preview.html.erb
+++ b/docs/app/views/examples/components/modal/_preview.html.erb
@@ -54,6 +54,11 @@
         sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
       </p>
 
+      <p class="t-sage-body">
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit,
+        sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+      </p>
+
       <% content_for :sage_footer do %>
         <% content_for :sage_footer_aside do %>
           <%= sage_component SageButton, {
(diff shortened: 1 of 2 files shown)


philschanely (approved):
Concept is fine. Two things to look at a little more just in case:

- is this spacing design-reviewed? It seems a little large.
- does adding a class of `sage-type` on the modal content do the trick by chance? Also with this on the particular classes on type elements are not always needed either.

QuintonJason:
@philschanely 
> * is this spacing design-reviewed? It seems a little large.

Yep. I also agree that it seems large, but 24px was design's feedback when I originally suggested `8px`
> * does adding a class of `sage-type` on the modal content do the trick by chance? Also with this on the particular classes on type elements are not always needed either.

It does do the trick if `8px` is to spec.

philschanely:
@QuintonJason ok interesting. If design says its right I guess go with it. But I agree with you it feels too open. 🤷‍♂️

monkeypox8 (approved):
👍🏼 LGTM. I also agree that the spacing feels a bit large. Hopefully this won't dramatically impact some of the lengthier modals.

joedakroub (approved):
Changes themselves look good. I'll 4th that this spacing is quite open and wandering, but will defer to design's recommendation. 👍🏼

PR #537, opened 2021-05-24
Stray copytext whitespace

philschanely:
## Description

This PR fixes a bug in Copy buttons where a formatting error led to undesired whitespace in the copied content.

## Testing in `sage-lib`

See Copy Text and verify that the copy button still performs as expected (but now should not contain extra whitespace after copied content).

(LOW) Fixes a bug in Copy buttons where a formatting error led to undesired whitespace in the copied content. See any copy buttons in the app to verify they behave as expected but also 
   - [ ] Fixes MAN-1550 where additional whitespace was coming through in copy text buttons as a part of the Custom Domain nameserver setup step.


pixelflips (approved):
LGTM!

monkeypox8 (approved):
👍🏼 LGTM!

PR #543, opened 2021-05-25
BUILD-1145: podcast search not working

justinjkim:
## Description
Podcast episode search was not working when you searched for something on a page different than what the episode was on. This was because adding the search filter on top of a pagination tried to look for that search result on that specific page. The solution was to remove the `page` key from the `URLSearchParams`.

## Screenshots

|  Before  |  After  |
|--------|--------|
| Screen Shot 2021-05-25 at 2 35 33 PM | Screen Shot 2021-05-25 at 2 36 20 PM |

## Testing in `sage-lib`
I don't think the current Sage public docs show any examples of actual search results, so doesn't seem like you need to test here.

(**MEDIUM**)
1. Create a podcast (can be either episodic or serial)
2. Generate at least 30 published episodes, so that the pagination kicks in
3. Name one of your episodes something easy to remember (`EX// "dogs are cool"`), since the automatically generated ones use the `lorem ipsum` fillers.
4. Navigate to a page other than the first one
5. In the search bar, enter your episode's name
6. You should get a search result with your episode

   - [x] One more examples of the component in use to either test the change or verify the change has not had adverse effects.
**I tested this on the sage product admin view of the product list**

## Related
BUILD-1145

### DIFF
--- a/packages/sage-system/lib/search.js
+++ b/packages/sage-system/lib/search.js
@@ -57,6 +57,7 @@ Sage.search = (function() {
       searchParams.delete('search');
     } else {
       searchParams.set('search', value);
+      searchParams.delete('page');
     }
 
     window.location.search = searchParams.toString();


packages/sage-system/lib/search.js
```diff
@@ -57,6 +57,7 @@ Sage.search = (function() {
       searchParams.delete('search');
     } else {
       searchParams.set('search', value);
+      searchParams.delete('page'); // BUILD-1145: pagination for search results won't work if the `page` key is set to something other than 1, so we need this deletion of the `page` key
```
philschanely, line 60: I don't know much about how this is working in `products` so trust to your needs here. Has this addition been tested in any other pages that use the SageSearch to verify it does not create a breaking change elsewhere?

philschanely (commented):
Thanks @justinjkim . One question below.

(looks like a rebase may be needed too)

packages/sage-system/lib/search.js
justinjkim: Hey Phil! I did test on some other searches, but will verify with Monica and Quinton to make sure. Thanks!

monicawheeler (changes requested):
I tested within Contacts in `kajabi-products` and it appears the issue is not resolved.

Testing steps:
1. 
step1
2. 
step1 result
3. 
step2
4. 
step2 result

monicawheeler (approved):
Verified with @justinjkim on pairing session that the `SageSearch` component does work properly after this fix.

packages/sage-system/lib/search.js
justinjkim: Hi @philschanely , looks like on the Rails pages that utilize `SageSearch`, this fix does work. Can't speak to the React version of the components, they still seem to be WIP, but I can check with the devs working on those pages.

PR #544, opened 2021-05-26
Remove Stylelint from docs site

voodooGQ:
## Description
A bug was identified by @QuintonJason where his docs instance would time out after certain actions were performed. This required him to completely reset the instance in order to get it back up and running. Through debugging was able to identify that the Stylelint plugin appears to be running against the minified version of the css file:

image

As this is just a docs site I'm not sure if the stylelinting is strictly necessary. To alleviate the problem I removed it from the system.


philschanely (changes requested):
(Do we need to rebase before this goes in?)

Not opposed to this completely, but could it instead be revised to only lint `docs/lib/sage-frontend/styles` instead of pulled out altogether? While def less crucial it would be nice to keep linting on those files. 🙏  

I'm willing to pair to help figure that out if desired.

voodooGQ:
@philschanely we can certainly work on that, but if we're going to keep it in place we probably should just have it be a `yarn` script that can be run rather than tying it into the webpack config.

I'll get some time on your calendar and we can go over it together.

voodooGQ:
@philschanely I had a little extra time so I just coded it out. Take a look at it now and see if that will meet the need.

QuintonJason (approved):

philschanely (approved):
🙌  Thanks for this @voodooGQ !

PR #551, opened 2021-06-01
Copy button: fill container

philschanely:
## Description

Adds a new optional parameter to CopyButton to allow it to fill its container when desired.

## Screenshots

Screen Shot 2021-06-01 at 6 21 42 PM

## Testing in `sage-lib`

See Components > Copy Text

1. (LOW) Adds a new property to Copy Button allowing it to fill its container. No impact on existing uses.


pixelflips (approved):
LGTM! Nice work!

PR #552, opened 2021-06-02
Update state color settings

philschanely:
## Description

Makes a color change to Transaction card: blue color will not be used but a green on is needed for the state color.

## Testing in `sage-lib`

See Transaction Card and the new state color configs.

(LOW/MEDIUM/HIGH/BREAKING) Non-breaking color change to Transaction Card. Does not affect any existing uses.


pixelflips (approved):
There seems to be a lint failure, but other than that LGTM! Nice work!

philschanely:
FYI this is set to `master` as a quick patch for running version bump. @monicawheeler

Issue #491, opened 2021-05-20
Update Interactive states


philschanely:
## Description

(Possibly in partnership with [Form field L1](https://github.com/Kajabi/sage-lib/issues/496))

Consider a mixin or similar approach to implement new focus styles consistently. Investigate how current focus is set up and build a new approach in a non-destructive/backwards-compliant way that can be implemented as each component is updated in follow-up tickets.

Also consider whether other common interactivity such as `:hover` and "active" (selected state, not `:active`).

QuintonJason:
Per conversation with design, we'll need to remove the `box-shadow` from the `<button>`s default state. It will also required the raised attribute to be removed. May handle that portion in a followup PR to reduce the effect of the breaking change.
<img width="829" alt="Screen_Shot_2021-06-03_at_8_18_24_AM" src="https://user-images.githubusercontent.com/1241836/120651676-a6314880-c444-11eb-99f3-f3b1e2cf9d32.png">

QuintonJason:
This issue stems from the fact that the focus ring and `box-shadow` will exist in the interactive element, not a wrapper or container. We have two options to handle currently:
1. add an inner `<span>` or `<div>` for the sole purpose of one of the box-shadow
2. remove the `box-shadow` from interactive elements that have a `raised` state
We're choosing option 2 for now

FuturaExtraBold:
Closing this, as the Button updates branch has been merged.

PR #554, opened 2021-06-03
[SAGE-491] interaction pattern

QuintonJason:
## Description

[SAGE-491](https://github.com/Kajabi/sage-lib/issues/491)

- [x] - expandable card -> update `display` property, it's currently full width
- [x] - add `tabindex` to StatusIcon for keyboard focus
- [ ] - ~catalog item - add additional 8px between to the left of the aside~ [SAGE-576](https://github.com/Kajabi/sage-lib/issues/576)
- [ ] - ~update stat box and transaction card link to be a subtle button for consistent focus state~ [SAGE-577](https://github.com/Kajabi/sage-lib/issues/577)
- [ ] - ~resolve our non-existent focus state on `<a>`'s that aren't button styled~ [SAGE-577](https://github.com/Kajabi/sage-lib/issues/577)

##### Buttons
|  Before  |  After  |
|--------|--------|
|buttonInteractionPatternBefore|buttonInteractionPattern|

##### Labels
|  Before  |  After  |
|--------|--------|
|labelInteractionPatternBefore|labelInteractionPatternAfter|

## Testing in `sage-lib`

[Sage 3](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/%5BWIP%5D-Sage-3-%E2%80%94-Admin-Components?node-id=3435%3A17) -> Elements

1. #554 (LOW) Interaction pattern(focus state) adjusted to latest specs. Interaction pattern, namely focus states changes will occur as a result in kajabi-products but no adverse effects expected.

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_expandable_card.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_expandable_card.html.erb
@@ -9,7 +9,6 @@
       "data-js-accordion": "header",
       "aria-expanded": "false",
     },
-    full_width: true,
   } %>
   <div class="
       <% if component.body_bordered.present? %>sage-expandable-card__body-bordered<% else %>sage-expandable-card__body<% end %>
--- a/packages/sage-assets/lib/stylesheets/components/_alert.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_alert.scss
@@ -90,16 +90,23 @@ $-alert-icon-colors: (
   height: rem(20px);
   width: rem(20px);
   padding: 0;
+  align-self: flex-start;
   font-size: sage-font-size(xl, false);
   line-height: 1;
   background-color: transparent;
   border: 0;
   appearance: none;
+
   @each $name, $color in $-alert-colors {
     .sage-alert--#{$name} & {
       color: sage-color($color, 400);
+
+      &:focus {
+        @include sage-focus-shadow(sage-color($color, 400));
+      }
     }
   }
+
   &:hover {
     opacity: 0.7;
   }
--- a/packages/sage-assets/lib/stylesheets/components/_alert.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_alert.scss
@@ -86,11 +86,10 @@ $-alert-icon-colors: (
 }
 
 .sage-alert__close {
-  overflow: hidden;
+  align-self: flex-start;
   height: rem(20px);
   width: rem(20px);
   padding: 0;
-  align-self: flex-start;
   font-size: sage-font-size(xl, false);
   line-height: 1;
   background-color: transparent;
@@ -101,9 +100,8 @@ $-alert-icon-colors: (
     .sage-alert--#{$name} & {
       color: sage-color($color, 400);
 
-      &:focus {
-        @include sage-focus-shadow(sage-color($color, 400));
-      }
+      @include sage-focus-outline($outline-offset-inline: 4px, $outline-offset-block: 2px);
+      @include sage-focus-outline--update-color(sage-color($color, 400));
     }
   }
 
--- a/packages/sage-assets/lib/stylesheets/components/_breadcrumbs.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_breadcrumbs.scss
@@ -72,7 +72,7 @@ $-breadcrumb-disabled-color: sage-color(grey, 500);
   }
 
   .sage-breadcrumbs:not(.sage-breadcrumbs--progressbar) &:focus {
-    @include sage-focus-outline($outline-offset-inline: 6px, $outline-offset-block: 0px);
+    @include sage-focus-outline($outline-offset-inline: 6px, $outline-offset-block: 0);
     @include sage-focus-outline--update-color($sage-focus-color);
   }
 }
--- a/packages/sage-assets/lib/stylesheets/components/_breadcrumbs.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_breadcrumbs.scss
@@ -1,4 +1,4 @@
-////
+//
 /// Breadcrumbs
 ///
 /// @group sage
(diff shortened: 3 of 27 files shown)


docs/lib/sage_rails/app/views/sage_components/_sage_expandable_card.html.erb
```diff
@@ -9,7 +9,6 @@
       "data-js-accordion": "header",
       "aria-expanded": "false",
     },
-    full_width: true,
```
QuintonJason, line 12: Need to check with @philschanely on impact

philschanely: The impact here is that in the paces this is used this button is only as wide as its contents demand, so it looks a little funny atop an expanding content area. not awful:

Screen Shot 2021-06-14 at 3 06 23 PM

QuintonJason: @philschanely cool deal. This was design feedback so making sure it doesn't negatively affect layout in your use case

philschanely: Nah, if they're ok its fine by me.

packages/sage-assets/lib/stylesheets/components/_breadcrumbs.scss
```diff
@@ -70,6 +70,11 @@ $-breadcrumb-disabled-color: sage-color(grey, 500);
 
     @include sage-tab-underline(".sage-breadcrumbs__link--current");
   }
+
+  .sage-breadcrumbs:not(.sage-breadcrumbs--progressbar) &:focus {
```
philschanely, line 74: The `--progressbar` type has a darker blue border atm but not sure if that is intended. Do you think it should have the new color too?

Screen Shot 2021-06-15 at 11 00 24 AM

philschanely:
@QuintonJason FWIW per the Catalog items I observed this partial/old highlight on the title when focused (see dark bark on right)

Screen Shot 2021-06-15 at 11 02 12 AM

packages/sage-assets/lib/stylesheets/components/_choice.scss
```diff
@@ -60,20 +60,14 @@ $-choice-radio-color-checked-inner: map-get($sage-radio-colors, checked-inner);
...
-    outline: none;
-
-    &::after {
-      border-color: sage-color(grey, 400);
-    }
-  }
-
   &:active,
+  &:focus,
   &.sage-choice--active {
```
philschanely, line 65: This area may need a follow up as it seems this has taken out the more prominent highlight on active choices.

In browser:

Screen Shot 2021-06-15 at 11 08 00 AM


Spec:

Screen Shot 2021-06-15 at 11 06 33 AM

philschanely:
@QuintonJason also possibly a follow-up I noticed hyperlinks that don't have otehr styles specified also have a default focus effect. Do you think this designed effect should be added to generic hyperlinks as well? Places like this one may also benefit:

Screen Shot 2021-06-15 at 11 11 22 AM

QuintonJason:
> @QuintonJason FWIW per the Catalog items I observed this partial/old highlight on the title when focused (see dark bark on right)
> 
> Screen Shot 2021-06-15 at 11 02 12 AM

Yep, handling this one in this PR, https://github.com/Kajabi/sage-lib/pull/554#issuecomment-861576815. I'll need to work with design to find a pattern for interactive elements that truncate, the `overflow: hidden` on the parent will hide `box-shadow`s and `::after`s. The compromise may be having a pattern that resides within that element's `content-box`, maybe link an underline, `border-bottom`, inverted colors, not sure.

QuintonJason:
> @QuintonJason also possibly a follow-up I noticed hyperlinks that don't have otehr styles specified also have a default focus effect. Do you think this designed effect should be added to generic hyperlinks as well?
> 
> Screen Shot 2021-06-15 at 11 15 14 AM
> 
> Places like this one may also benefit:
> 
> Screen Shot 2021-06-15 at 11 11 22 AM

https://github.com/Kajabi/sage-lib/pull/554#issuecomment-861576815 will also have the fix for text links which currently have an issue if the current icon it uses resides on the `::after` pseudo element. Will bring this one up to design as well

Great callout on the hero, that seems to be an edge case since the play button is using that element's `::after`. I'll add to the current followup PR

philschanely (approved):
AMAZING and thorough work here! I've noted a few more little things I noticed as I went through each of the components. Things are looking in really good shape! 🙇‍♂️  🏅

packages/sage-assets/lib/stylesheets/components/_alert.scss
```diff
@@ -95,11 +95,16 @@ $-alert-icon-colors: (
   background-color: transparent;
   border: 0;
   appearance: none;
+
   @each $name, $color in $-alert-colors {
     .sage-alert--#{$name} & {
       color: sage-color($color, 400);
+
```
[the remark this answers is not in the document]
QuintonJason: You are correct. My initial thought was that since this was a text link it fell into the followup category; however since the parent modifier dictates the color palette for all descendants, they should also follow that rule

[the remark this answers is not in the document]
QuintonJason: Updated in the followup PR: https://github.com/Kajabi/sage-lib/pull/580

packages/sage-assets/lib/stylesheets/components/_breadcrumbs.scss
QuintonJason: This is edge case in which the `::after` is being occupied by the `border-bottom` effect. I'll need to check with design as this also is the case for `SageTabs`. Will handle this in the followup PR

packages/sage-assets/lib/stylesheets/components/_choice.scss
QuintonJason: pushed an update

monkeypox8 (commented):
Overall this is working great so far! As we'd discussed, it seems a decision between design and the @Kajabi/ux-devs group is needed on whether the updated focus outline fully satisfies the [WCAG standards](https://www.w3.org/WAI/WCAG21/Techniques/general/G195), since a 2px outline width can override the 3:1 contrast ratio requirements specified.

PR #555, opened 2021-06-03
fix(modal js): prevent modal JS from erroring on close when `active: true` attribute set

AndrwM:
BUILD-1112

## Description

The modal JS fails to close if a previously selected element does not exist, in the instance of
`active: true` there is not a last focused element. This changes allows the modal JS to fallback
gracefully rather than fire errors when a close is attempted

## Screenshots
n/a

## Testing in `sage-lib`
Ensure modals which have `active :true` can be closed.

1. (**LOW**) Prevent modals which have `active :true` from erroring out. 
   - [ ] Check modals that modals work as expected, incoming work makes use of `active: true` in https://github.com/Kajabi/kajabi-products/pull/19033

## Related
https://github.com/Kajabi/kajabi-products/pull/19033

### DIFF
--- a/packages/sage-system/lib/modal.js
+++ b/packages/sage-system/lib/modal.js
@@ -17,7 +17,7 @@ Sage.modal = (function() {
   const EVENT_ACTIVE = "sage.modal.active";
   const EVENT_OPENING = "sage.modal.opening";
   const EVENT_OPEN = "sage.modal.open";
-  let SELECTOR_LAST_FOCUSED;
+  let selectorLastFocused;
 
   // ==================================================
   // Functions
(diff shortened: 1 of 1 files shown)


QuintonJason (approved):
💥  Tested in `kajabi-products` and closes as expected!

pixelflips (approved):
LGTM!

pixelflips:
@AndrwM unless you're trying to get this in asap, you may want to merge it into the develop branch instead of master.

AndrwM:
@pixelflips Apologies on this! Switching the target branch. Thanks!

pixelflips:
> @pixelflips Apologies on this! Switching the target branch. Thanks!

All good and no apology needed. I've done it many times!!

PR #559, opened 2021-06-03
JBH - New Hint Component

FuturaExtraBold:
## Description
Closes #521 

Adds new Hint component to Sage library.

## Screenshots
image

### Rails

- View [Rails component](http://localhost:4000/pages/component/hint)
- Check that component UI exists and nothing is broken
- Check docs

### React
- View [Storybook](http://localhost:4100/?path=/docs/sage-hint--default)
- Check that component UI exists and nothing is broken
- Check docs

## Testing in `kajabi-products`
(LOW) Adds new Hint component. No effect on existing work in Kajabi Products.

PR #553, opened 2021-06-02
Indicator - add prop to show text representation of the current page

QuintonJason:
## Description

- [x] add a text variant for the indicator component based on the current page

## Screenshots

|  On demo site  |  In modal header  |
|--------|--------|
|Screen Shot 2021-06-02 at 1 13 27 PM|Screen_Shot_2021-06-02_at_11_10_33_AM|

## Testing in `sage-lib`

- Visit the Indicator component page and see the `show-text` variant
- Visit the modal page and click on the standard modal, then verify that the modal indicator is place correctly

## Testing in `kajabi-products`

1. (LOW) This is a new property and has no current affect with the app

PR #550, opened 2021-06-01
Sage 3 - align border radius to spec

QuintonJason:
## Description

- [x] - update tokens from `2px`, `4px`, `6px` -> `4px`, `6px`, `8px` respectively
- [x] - check all components and update anything not align to spec

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2021-06-01 at 3 00 11 PM|Screen Shot 2021-06-01 at 2 59 58 PM|

## Testing in `sage-lib`

Verify that components with the `border-radius` property set are aligned to spec. Some elements of note are buttons, form fields, and panel

## Testing in `kajabi-products`

1. (MEDIUM) Updates the `border-radius` property appwide, affecting buttons, cards, panels, and form fields
   - [ ] Open the app on your respective team's page and verify that the `border-radius` is updated 

## Related

#472

PR #561, opened 2021-06-07
Next version bump (~v4.3.0) 

QuintonJason:
1. #555  (LOW) Prevent modals which have `active :true` from erroring out. 
   - [ ] BUILD: Check that modals work as expected, incoming work makes use of `active: true` in https://github.com/Kajabi/kajabi-products/pull/19033
1. #550  (LOW) Updates the `border-radius` property app-wide, affecting buttons, cards, panels, and form fields in minor ways with new border radii.
1. #553  (LOW) Adds new property to indicator to show text version when desired. This is a new property and has no current affect with the app.
1. #562  (LOW) Updates existing and new tooltips to show on either hover or focus. No affect to the tooltip structure or content.
1. #559  (LOW) Adds new Hint component. No effect on existing work in Kajabi Products.


pixelflips (approved):
LGTM! 🎉

Issue #490, opened 2021-05-20
Update Colors


philschanely:
## Description

Refresh color values including those listed in Figma. **NOTE:** Be sure to check the actual hex values of the color style and if needed update the text label over the value in Figma (the text label may not match the actual values but they need to do so).

**Links**
- [Figma specs](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/?node-id=3554%3A25)
- [Existing](https://sage-design-system.kajabi.com/pages/design/color)

PR #563, opened 2021-06-07
style(colors): updates the colors to sage 3 colors

cmcfadzean:
re #490

## Description
Updates the colors to use the new sage 3 colors. 

## Testing in Sage

See Design > Colors and elsewhere colors are applied.

(LOW) Color palette adjusted to latest specs. Color changes will occur as a result in `kajabi-products` but no adverse effects expected.


packages/sage-assets/lib/stylesheets/tokens/_color_palette.scss
```diff
@@ -10,58 +10,58 @@
 ///
 $sage-colors: (
   primary: (
-    100: #e6f4fe,
-    200: #8ecafb,
+    100: #E6F4FE,
```
philschanely, line 13: @cmcfadzean can you please adjust all letters to lower-case in keeping with the precedent before? 🙏

philschanely (commented):
Thanks @cmcfadzean one short comment for adjustment. 👍 otherwise.

packages/sage-assets/lib/stylesheets/tokens/_color_palette.scss
```diff
@@ -82,12 +82,12 @@ $sage-colors: (
 );
 
 ///
-/// Sage breakpoints token utility
+/// sage breakpoints token utility
```
philschanely, line 85: haha looks like a plugin challenged all letters to lowercase now 😂

philschanely (approved):
Not blocking, but it seems the uncaps you applied was a bit greedy :)

QuintonJason (approved):
nice!

packages/sage-assets/lib/stylesheets/tokens/_color_palette.scss
cmcfadzean: hahaha

PR #574, opened 2021-06-12
SPIKE: a11y-friendly disabled state

monkeypox8:
## Description
Testing an option for more inclusive `disabled` controls. Should help allow secondary actions (tooltips, etc) but still prevent form submissions/actions unless requirements are met.

References:
- https://css-tricks.com/making-disabled-buttons-more-inclusive/
- https://codepen.io/stevef/post/short-note-on-the-disabled-attribute

## Testing in `sage-lib`
1. Navigate to the "button" component in the sage docs
2. The last button in each example row has a tooltip enabled. Hover your mouse over the active (non-disabled) example to confirm a tooltip appears, and the button can be clicked/toggled via keypress
3. Hover over the "disabled" (`aria-disabled`) tooltip example, and confirm a tooltip appears but the button **cannot** be actuated

1. (**MEDIUM**) Disabled button states use a combination of JS and `aria-disabled` to emulate the `disabled` attribute


monkeypox8:
@QuintonJason

QuintonJason:
@monkeypox8 This does seem to resolve the core issue with disabled button not have interaction on child elements. This is great as now we have that consistent behavior between buttons in isolation and button within dropdowns. I'm liking it!

disabledButtonInteraction

monkeypox8:
I'm not too familiar with the `arrive` JS library, so we'll need to confirm that it listens for `aria` changes. I think we may have to change the config and set `fireOnAttributesModification`  to `true`, assuming it doesn't have a dramatic impact on performance.

QuintonJason:
@monkeypox8 are you able to update the button preview page to contain a button with a tooltip, independent of being in a dropdown?

monkeypox8:
> @monkeypox8 are you able to update the button preview page to contain a button with a tooltip, independent of being in a dropdown?

Updated. Also fixed an issue with the small icon previews not rendering their disabled states.

monkeypox8:
Closing this for now as I believe most of the concerns have been addressed and the branch has gone stale. Can revisit/reopen if necessary.

Issue #497, opened 2021-05-20
Label


philschanely:
## Description

Update component to match latest specs in Sage 3.0. Be sure to check the variations for this component and set up follow-up issues in Tier 4 as needed.

**Dependencies**:
- [Color updates](https://github.com/Kajabi/sage-lib/issues/490)
- [Border radius](https://github.com/Kajabi/sage-lib/issues/472)
- [Interactive states](https://github.com/Kajabi/sage-lib/issues/491)

**Links**
- [Figma specs](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/WIP-Sage-3-%E2%80%94-Admin-Components?node-id=3655%3A16569)
- [Existing component/s](https://sage-design-system.kajabi.com/pages/component/label)

FuturaExtraBold:
Closed in #584. Thanks, @QuintonJason!

PR #584, opened 2021-06-17
[SAGE-497] Label - align to spec

QuintonJason:
Closes [SAGE-497](https://github.com/Kajabi/sage-lib/pull/497)
Dependent on [SAGE-554](https://github.com/Kajabi/sage-lib/pull/554)
## Description

- [x] align label to spec
    - [x] update interactive border-radius
    - [x] reduce label text font-weight

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2021-06-21 at 8 51 06 AM|Screen Shot 2021-06-21 at 8 50 26 AM|

## Testing in `sage-lib`

Visit the label rails and react page 

1. (LOW) Label component adjusted to latest specs. Label changes will occur as a result in kajabi-products but no adverse effects expected.

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_button.scss
@@ -177,7 +177,8 @@ $-btn-interactive-label-icon-size: rem(24px);
     border-radius: 0 sage-border(radius-x-large) sage-border(radius-x-large) 0;
 
     &::before {
-      font-size: sage-font-size(body-sm);
+      margin-top: 1px; /* needed as sage-font-size(sm) is equates 13px and not 14px. This 1px is accounting for that 1px */
+      font-size: sage-font-size(sm);
     }
 
     &::after {
--- a/packages/sage-assets/lib/stylesheets/components/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_button.scss
@@ -174,7 +174,11 @@ $-btn-interactive-label-icon-size: rem(24px);
     right: 0;
     width: $-btn-interactive-label-icon-size;
     margin: auto 0;
-    border-radius: 0 sage-border(radius) sage-border(radius) 0;
+    border-radius: 0 sage-border(radius-x-large) sage-border(radius-x-large) 0;
+
+    &::before {
+      font-size: sage-font-size(body-sm);
+    }
 
     &::after {
       width: $-btn-interactive-label-icon-size;
@@ -184,10 +188,6 @@ $-btn-interactive-label-icon-size: rem(24px);
 
     &:first-child:not(:last-child) {
       margin-right: sage-spacing(xs);
-    }
-
-    &:last-child:not(:first-child) {
-      margin-left: sage-spacing(xs);
     }
 
     + & {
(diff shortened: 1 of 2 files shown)


packages/sage-assets/lib/stylesheets/components/_button.scss
```diff
@@ -174,22 +174,23 @@ $-btn-interactive-label-icon-size: rem(24px);
     right: 0;
     width: $-btn-interactive-label-icon-size;
     margin: auto 0;
-    border-radius: 0 sage-border(radius) sage-border(radius) 0;
+    border-radius: 0 sage-border(radius-x-large) sage-border(radius-x-large) 0;
+
+    &::before {
+      margin-top: 1px; /* needed as sage-font-size(sm) is equates 13px and not 14px. This 1px is accounting for that 1px */
```
monkeypox8, line 180: I think we'll want to handle this with a `transform`, since the margin adds to the secondary button height on hover:

monkeypox8 (commented):
Looking fantastic! Just to confirm, will the focus outline states be handled in a separate button/interaction PR?

packages/sage-assets/lib/stylesheets/components/_button.scss
QuintonJason: Great catch! Pushed the fix

monkeypox8 (approved):
👍🏼 LGTM!

Firefox seems to be rendering the alignment in its own way from Chrome/Safari, so the added transform/margin ends up pushing the button an additional 1px off the vertical line. I don't think it's a major concern, though:

philschanely (approved):
Looks and works well. 🙌  One non-blocker is this older interactive—effect which I imagine will go away soon given your fabulous new interactive effects PR—does look a little odd with the new pill shape:

Screen Shot 2021-06-24 at 10 27 05 AM

Definitely an optional fix but wanted to raise awareness just in case it was of interest to fix now.

philschanely (approved):
Looking good, @QuintonJason . 

These may deserve a follow-up or two, but I noted the following:

1. To match the specs closer it appears we need to revise the secondary button hover settings a bit:

    Spec:

    Screen Shot 2021-07-19 at 12 01 42 PM

    As built to date:
    Screen Shot 2021-07-19 at 12 10 55 PM

2. There are also some aspects of how the colors and the mixture of sage btn within this component for the secondary button feel potentially leaky that might be nice revisit. See color settings in loop in `_label.scss` that override those in `_button.scss`. I wonder if we indeed need to use `SageButton` there or if the tradeoff is in our favor to use a custom styled HTML `button` instead?

Bottom line is nothing blocking this PR but some ideas about potential follow up. If you agree, I'm happy to file them in an issue for later. Thoughts?

QuintonJason:
> Looking good, @QuintonJason .
> 
> These may deserve a follow-up or two, but I noted the following:
> 
> 1. To match the specs closer it appears we need to revise the secondary button hover settings a bit:
>    Spec:
>     Screen Shot 2021-07-19 at 12 01 42 PM
>    As built to date:
>    Screen Shot 2021-07-19 at 12 10 55 PM
> 2. There are also some aspects of how the colors and the mixture of sage btn within this component for the secondary button feel potentially leaky that might be nice revisit. See color settings in loop in `_label.scss` that override those in `_button.scss`. I wonder if we indeed need to use `SageButton` there or if the tradeoff is in our favor to use a custom styled HTML `button` instead?
> 
> Bottom line is nothing blocking this PR but some ideas about potential follow up. If you agree, I'm happy to file them in an issue for later. Thoughts?

@philschanely Great callout! Made one commit to update hover colors, but we can revisit the need for the `SageButton` in a followup

QuintonJason:
@philschanely https://github.com/Kajabi/sage-lib/issues/640

PR #585, opened 2021-06-21
Next version bump (~v4.5.0)

QuintonJason:
1. (LOW) Documentation only, nothing affected within the app
2. (LOW) Adds a new property to Dropdown that allows for toggling the panel from the outside. No expected side effects for current usages. Validated by UXD by checking the various dropdowns in the People page.


monicawheeler (approved):
🚀

PR #586, opened 2021-06-21
Sage Button - Custom Content

joedakroub:
## Description
This PR makes the `value` prop optional and adds the ability to provide custom content and components to the body of the Sage Button component.

This feature will be necessary to accommodate the new design for the user dropdown that will reside in the top right of the application frame.

## Testing in `sage-lib`
1. Visit http://0.0.0.0:4000/pages/component/button and ensure that the `Custom content` example renders as expected.

1. **LOW** As this is a new feature, it should not affect current implementations, however testing in areas that use buttons should be done to be safe.
   - [ ] Settings - Site Details/Blog Settings/Marketing Details
   - [ ] Website page
   - [ ] Products page

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_button.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_button.html.erb
@@ -17,6 +17,7 @@
     <%= "sage-btn--#{component.style}" if component.style %>
     <%= "sage-btn--align-end" if (component.align == "end") %>
     <%= "sage-btn--icon-#{component.icon[:style]}-#{component.icon[:name]}" if component.icon %>
+    <%= "sage-btn--custom-content" if component.content %>
     <%= component.generated_css_classes %>
   "
 >
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_button.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_button.html.erb
@@ -17,7 +17,6 @@
     <%= "sage-btn--#{component.style}" if component.style %>
     <%= "sage-btn--align-end" if (component.align == "end") %>
     <%= "sage-btn--icon-#{component.icon[:style]}-#{component.icon[:name]}" if component.icon %>
-    <%= "sage-btn--custom-content" if component.content %>
     <%= component.generated_css_classes %>
   "
 >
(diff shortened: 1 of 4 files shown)


packages/sage-assets/lib/stylesheets/components/_button.scss
```diff
@@ -439,6 +439,10 @@ $-btn-interactive-label-icon-size: rem(24px);
   }
 }
 
+.sage-btn--custom-content {
+  display: inline-block;
+}
```
joedakroub, line 444: This is set because `display: inline-flex` (the original setting for Sage button) causes all of the nodes to pack together. This will enable the nodes to be spaced normally.

philschanely (approved):
Great work here. Two questions (not blocking this PR per se):

1. Are additional styles needed for when you break completely away from these color styles, etc. such as for the users dropdown? In some ways, that may be accomplishable using an Avatar as the trigger for a dropdown. But say you pass an avatar as the content for a Button, does this styling suffice, or do you intend a follow up?
2. Would you be willing to make similar modifications to the React component to ensure they align? Could def be a follow-up, or at least add an Issue to remind us to do it.

joedakroub:
@philschanely Good questions. 

1. In the PR I'm putting together for the user dropdown, I am using an Avatar and Icon component inside of a Button. I did need to wrap them with a `div` that has some light layout styling, but otherwise nothing else is needed. I kept this PR open-ended on that front because I think it may be difficult to anticipate how often having custom content will be used and what the use cases will be when used.
2. Happy to make the React updates in a follow-up PR 👍🏼

docs/lib/sage_rails/app/views/sage_components/_sage_button.html.erb
```diff
@@ -25,9 +26,13 @@
       <%= component.value.html_safe %>
     </span>
   <% else %>
+    <% if component.value %>
     <span class="sage-btn__truncate-text">
```
QuintonJason, line 30: indentation nitpick

QuintonJason:
@joedakroub Not sure if this is an edge case but when custom content is present in additional to an icon, the alignment is off due to `display: inline-block`. There may be a need a wrapper of the custom content to reduce the need for the `display` override.

joedakroub:
@QuintonJason Good find. I added a wrapper around the custom content and removed the custom styling. Seems to be working well. 👍🏼

QuintonJason (approved):
Nice!

PR #587, opened 2021-06-22
External dropdown triggers

### LINKED ISSUE #582: Trigger dropdown from outside
## Description

Need to add ability to trigger closing (and possibly opening) a dropdown panel from outside the panel itself. Consider a triggerToken like what we've used elsewhere.

philschanely:
## Description

This PR adds a new prop to React Dropdown that allows for triggering toggling the panel from the outside without requiring it to be bound to a boolean state variable. Now simply toggle the panel's visibility by changing the value of a token string such as by using `react-uuid` to set a unique universal id and changing it when needed. 

## Testing in `sage-lib`

See the `CustomPanelStory` in storybook for an example of this usage.

(LOW) Adds a new property to Dropdown that allows for toggling the panel from the outside. No expected side effects for current usages. Validated by UXD by checking the various dropdowns in the People page.

## Related
Closes #582

### DIFF
--- a/packages/sage-react/lib/Dropdown/Dropdown.jsx
+++ b/packages/sage-react/lib/Dropdown/Dropdown.jsx
@@ -29,6 +29,7 @@ export const Dropdown = ({
   panelModifier,
   panelMaxWidth,
   panelSize,
+  panelStateToken,
   triggerButtonSubtle,
   triggerModifier,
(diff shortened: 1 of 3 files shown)


monicawheeler (approved):
👍🏻 LGTM

Issue #591, opened 2021-06-22
Ensure `html_attributes` can be reliably used on all Components


philschanely:
## Description

Some elements such as `Button` either do not implement `generated_html_attributes` or use their own such as `attributes`. This needs to be tidied so that we can use a consistent interface for common attributes such as `id` or testing handles such as `data-kjb-element`.

philschanely:
Shim uses of `attribute`:

- [ ] Button
- [ ] Checkbox
- [ ] Choice
- [ ] Dropdown.items
- [ ] IconCard
- [ ] LabelSecondaryButton
- [ ] Label.secondaryButton
- [ ] Link
- [ ] NavLink
- [ ] Nav.items and Nav.items[n].items
- [ ] Radio
- [ ] Tab
- [ ] Tabs.items

Add `generated_html_attributes`:

TBD

PR #594, opened 2021-06-24
Only allow assignment for assignable component attributes

prsimp:
## Description
This ensures that no unassignable attributes will attempt to be assigned when instantiating a component. An attempt to assign unassignable attributes will result in an `ActiveModel::UnknownAttributeError` which will typically present as an `ActionView::TemplateError` when encountered during rendering. This allows for a more graceful "failure" where unassignable attributes will be (silently) ignored, similar to how invalid HTML attributes are treated.

We'll still validate the full schema in non-production environments, but only while passing in assignable attributes.

## Testing in `sage-lib`
Attempt to render a component with attributes not supported by the components schema. Rather than failing, Sage should simply ignore the unknown attributes.

Attempt to render a component with attributes not supported by the components schema. Rather than failing, Sage should simply ignore the unknown attributes.

## Related
N/A

### DIFF
--- a/docs/lib/sage_rails/app/helpers/sage_component_helper.rb
+++ b/docs/lib/sage_rails/app/helpers/sage_component_helper.rb
@@ -1,11 +1,12 @@
 module SageComponentHelper
   def sage_component(component, attributes, &block)
-    validate_sage_component_attributes(component, attributes)
+    relevant_attributes = attributes.slice(*component::ATTRIBUTE_SCHEMA.keys)
+    validate_sage_component_attributes(component, relevant_attributes)
 
     component.new({
       context: self,
       content: block_given? ? capture(&block) : nil
-    }.merge(attributes)).render
+    }.merge(relevant_attributes)).render
   end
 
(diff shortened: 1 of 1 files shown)


prsimp:
@monkeypox8 -- I left it as a draft for now. Feel free to tag whoever else you want to bring in for discussion 🤘

monkeypox8:
Will bring this up for discussion with the group for next steps. As described above, this change brings sage rails components closer to HTML markup in behavior with unrecognized attribute definitions. While we should still reasonably expect spec coverage to handle cases where a component is failing to render, the worst case scenario will now result in an unstyled component rather than an entirely broken view.

Affects #591

prsimp:
> While we should still reasonably expect spec coverage to handle cases where a component is failing to render, the worst case scenario will now result in an unstyled component rather than an entirely broken view.

If we wanted, we could only slice the attributes in production. This way the validation method would raise on either bad attrs or schema type mismatches. With this change, in non-prod, only schema type issues will raise.

Issue #479, opened 2021-05-12
Add `testing_handle` as global property on SageComponent


philschanely:
## Description

To aid in testing in `kajabi-products` it is desirable to have a global attribute available on all components `testing_handle`. It should be added as another part of the `generated_html_attributes` that is output on all components and defined in `SageComponent`.

NOTE: Future work may consider how to also provide testing attributes to children within a component.

philschanely:
Requirements for the handle:
- Needs to be a single attribute used for all QE testing
- Needs to be clear and explicit by its name what its for
- Needs to be attachable to every component across the system

PR #601, opened 2021-06-29
Testing attribute and `generated_html_attributes` cleanup

philschanely:
## Description

This is one part of an effort to a) implement a testing id on all components and b) tidy up how general attributes are provided on elements. This one accomplishes the following:

- [x] Ensure all Rails elements output `generated_html_attributes`, even though some also use `attributes` in addition to the centralized `html_attributes` available from `SageComponent`
- [x] Add a `test_id` attribute to all Rails components that outputs the value into `generated_html_attributes` with `data-kjb-element-"[value]"`

Followup work will include:

- [ ] #604 and #605 Deprecating uses of `attributes` in favor of `html_attributes`
- [ ] #603 Add a similar `testId` property on all React components.
- [ ] #602 Document common attributes in the site such as `html_attributes`, `css_classes`, and `test_id`

## Testing in `sage-lib`

Observe changes do not break any existing component pages and experiment locally with the new `test_id` attribute.

(LOW) Additional attribute added for specs and other automated testing. No impact expected on existing component usage.

## Related
- #591 
- #479

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_checkbox.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_checkbox.html.erb
@@ -18,6 +18,7 @@
     <%= "disabled" if component.disabled %>
     <%= "required" if component.required %>
     aria-label="<%= component.label_text %>"
+    <%= component.generated_html_attributes.html_safe %>
   >
 <% else %>
   <div class="
@@ -26,6 +27,7 @@
       <%= component.has_error ? "sage-checkbox--error" : "" %>
       <%= component.generated_css_classes %>
     "
+    <%= component.generated_html_attributes.html_safe %>
   >
     <input
       id="<%= component.id %>"
--- a/docs/lib/sage_rails/app/sage_components/sage_component.rb
+++ b/docs/lib/sage_rails/app/sage_components/sage_component.rb
@@ -4,6 +4,7 @@ class SageComponent
   attr_accessor :content
 
   ATTRIBUTE_SCHEMA = {
+    test_id: [:optional, NilClass, String],
     html_attributes: [:optional, Hash],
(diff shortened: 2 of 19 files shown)


docs/lib/sage_rails/app/views/sage_components/_sage_checkbox.html.erb
```diff
@@ -26,6 +27,7 @@
       <%= component.has_error ? "sage-checkbox--error" : "" %>
       <%= component.generated_css_classes %>
     "
+    <%= component.generated_html_attributes.html_safe %>
```
monkeypox8, line 30: For form inputs, is QE expecting `data-kjb-element` to live on the container, or on the input itself? I may be missing context, but as Button and Link apply the attribute directly to the element, I'd expect that to carry across here as well?

monkeypox8 (approved):
👍🏼 LGTM, working as expected. Tested on various views throughout the app, only question I had was around the output location of the element on inputs.

monkeypox8:
Related to the attribute assignments here: I meant to bring it up in this week's UXD meeting, but @prsimp  put together a PR modifying the components to silently fail any attributes it doesn't recognize in the schema: https://github.com/Kajabi/sage-lib/pull/594

The idea being that any sage components lacking spec coverage will still at the very least render arkup rather than throwing an error and breaking the view entirely.

docs/lib/sage_rails/app/views/sage_components/_sage_checkbox.html.erb
philschanely: Good point. I will note a follow up. I'd like to get this out this week so that was can start adding handles. But I will follow up with them about this attributes placement for future work. I believe for now if needed they can use selectors to drill in.

PR #613, opened 2021-07-06
[Sage-491] interaction pattern

QuintonJason:
## Description

[SAGE-491](https://github.com/Kajabi/sage-lib/issues/491)

Previously closed original ticket for comment context: [SAGE-554](https://github.com/Kajabi/sage-lib/pull/554)
Previously closed followup ticket for comment context: [SAGE-580](https://github.com/Kajabi/sage-lib/pull/580)

- [x] - expandable card -> update `display` property, it's currently full width
- [x] - add `tabindex` to StatusIcon for keyboard focus
- [x] create focus styles for links that aren't buttons
- [x] create focus to links in TransactionCard
- [x] create focus to links in StatBox
- [x] alert - update focus state for action buttons
- [x] hero - resolve focus state on hero artwork
- [x] catalog item - resolve focus state on title
- [x] button / link truncation is not functioning properly
- [x] resolve react helpLink shadow sizing

##### Buttons
|  Before  |  After  |
|--------|--------|
|buttonInteractionPatternBefore|interactiveafter|

##### Labels
|  Before  |  After  |
|--------|--------|
|labelInteractionPatternBefore|interactivelabelafter|

##### Catalog Item Interaction
|  Before  |  After  |
|--------|--------|
|catalogItemBefore|interactivecatalogafter|

##### Text links
|  Before  |  After  |
|--------|--------|
|linksBefore|interactivelinksafter|

##### Stat Box link
|  Before  |  After  |
|--------|--------|
|Screen Shot 2021-06-14 at 4 57 38 PM|Screen Shot 2021-07-08 at 12 55 52 PM|

##### Alert action buttons
|  Before  |  After  |
|--------|--------|
|alertFocusBefore|interactivealertafter|

## Testing in `sage-lib`

[Sage 3 specs](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/%5BWIP%5D-Sage-3-%E2%80%94-Admin-Components?node-id=3435%3A17) -> Elements
The best way to test is to visit all pages and tab through the docs site

1. #613 (LOW) Interaction pattern(focus state) adjusted to latest specs. Interaction pattern, namely focus states changes will occur as a result in kajabi-products but no adverse effects expected.

### DIFF
--- a/docs/app/views/examples/components/icon_card/_preview.html.erb
+++ b/docs/app/views/examples/components/icon_card/_preview.html.erb
@@ -33,13 +33,11 @@
 <h3 class="t-sage-heading-6">Icon Card - round</h3>
 <div class="sage-row">
   <% SageTokens::STATUSES.each do | color | %>
-    <div class="sage-col-2">
-      <%= sage_component SageIconCard, {
-        color: color,
-        icon: "file",
-        label: "File graphic",
-        round: true
-      } %>
-    </div>
+    <%= sage_component SageIconCard, {
+      color: color,
+      icon: "file",
+      label: "File graphic",
+      round: true
+    } %>
   <% end %>
 </div>
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_banner.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_banner.html.erb
@@ -23,14 +23,12 @@
   <% end %>
 
   <% if component.link.present? %>
-    <a
-      class="sage-banner__link sage-btn--subtle"
-      <% component.link[:attributes].each do |key, value| %>
-        <%= "#{key}='#{value}'".html_safe %>
-      <% end if component.link[:attributes]&.is_a?(Hash) %>
-    >
-      <%= component.link[:name] %>
-    </a>
+    <%= sage_component SageButton, {
+      value: component.link[:name],
+      subtle: true,
+      css_classes: "sage-banner__link",
+      attributes: component.link[:attributes]
+    } %>
   <% end %>
 
   <%= component.content if component.content.present? %>
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_banner.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_banner.html.erb
@@ -1,4 +1,4 @@
-<% if component.banner_context.present? %>
+  <% if component.banner_context.present? %>
   <div class="
     sage-banner-wrapper
     <%= "sage-banner-wrapper--context-#{component.banner_context}" if component.banner_context.present? %>
@@ -24,7 +24,7 @@
 
   <% if component.link.present? %>
     <a
-      class="sage-banner__link"
+      class="sage-banner__link sage-btn--subtle"
       <% component.link[:attributes].each do |key, value| %>
         <%= "#{key}='#{value}'".html_safe %>
       <% end if component.link[:attributes]&.is_a?(Hash) %>
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_transaction_card.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_transaction_card.html.erb
@@ -12,22 +12,24 @@
     <% end %>
   </div>
   <div class="sage-transaction-card__body">
-    <<%= html_tag %> class="sage-transaction-card__name">
-      <% if component.name_href.present? %>
-        <a class="sage-link sage-transaction-card__name-link" href="<%= component.name_href %>">
-          <span class="t-sage--truncate">
-            <%= component.name %>
-          </span>
-        </a>
-      <% else %>
-        <div class="sage-transaction-card__name-text">
-          <span class="sage-btn__truncate-text">
-            <%= component.name %>
-          </span>
-        </div>
-      <% end %>
-    </<%= html_tag %>>
-    <div class="sage-transaction-card__amount<%= " sage-transaction-card__amount--#{component.amount_color}" if component.amount_color.present? %>"><%= component.amount %></div>
+    <div className="sage-transaction-card__body-group">
+      <<%= html_tag %> class="sage-transaction-card__name">
+        <% if component.name_href.present? %>
+          <a class="sage-link sage-transaction-card__name-link" href="<%= component.name_href %>">
+            <span class="t-sage--truncate">
+              <%= component.name %>
+            </span>
+          </a>
+        <% else %>
+          <div class="sage-transaction-card__name-text">
+            <span class="sage-btn__truncate-text">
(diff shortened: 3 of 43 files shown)


docs/app/views/examples/components/icon_card/_preview.html.erb
```diff
@@ -33,13 +33,11 @@
 <h3 class="t-sage-heading-6">Icon Card - round</h3>
 <div class="sage-row">
   <% SageTokens::STATUSES.each do | color | %>
-    <div class="sage-col-2">
```
QuintonJason, line 36: This is purely docs related.

docs/lib/sage_rails/app/views/sage_components/_sage_banner.html.erb
```diff
@@ -24,7 +24,7 @@
 
   <% if component.link.present? %>
     <a
-      class="sage-banner__link"
+      class="sage-banner__link sage-btn--subtle"
```
philschanely, line 27: Is the thought finished here? Seems odd to have this mixture of `sage-btn--subtle` on its own?

packages/sage-react/lib/TransactionCard/TransactionCard.jsx
```diff
@@ -47,40 +41,33 @@ export const TransactionCard = ({
...
-              subtle={true}
-            >
-              {relatedProperty}
-            </Button>
-          ) : (
-            <Property className="sage-transaction-card__product" icon={SageTokens.ICONS.TAG}>{relatedProperty}</Property>
-          )}
-          <Property className="sage-transaction-card__time">{transactionTime}</Property>
-        </div>
+        <NameTag className="sage-transaction-card__name">
```
philschanely, line 44: Same here... is there a reason these changes are not carrying forward the `body-group` containers?

docs/lib/sage_rails/app/views/sage_components/_sage_transaction_card.html.erb
```diff
@@ -12,42 +12,40 @@
     <% end %>
   </div>
   <div class="sage-transaction-card__body">
-    <div class="sage-transaction-card__body-group">
-      <<%= html_tag %> class="sage-transaction-card__name">
-        <% if component.name_href.present? %>
-          <a class="sage-transaction-card__name-link" href="<%= component.name_href %>">
+    <<%= html_tag %> class="sage-transaction-card__name">
```
philschanely, line 15: This seems to have reverted a necessary change that used the `body-group` containers and revised `footer` classname below. Was this an accident?

docs/lib/sage_rails/app/views/sage_components/_sage_banner.html.erb
QuintonJason: Yes my thinking is that a mixture is fine if the context is different. That particular has all the styles and interactive effects of a subtle button, but in the content of the banner's link, there are overrides specific to that component. It would be the same necessity if the difference was positioning rather than the text underline.

packages/sage-react/lib/TransactionCard/TransactionCard.jsx
QuintonJason: Good eye. That should not have been removed. Pushed a fix.

joedakroub (approved):
This changes look really great. I don't have any major concerns here. However, I have noticed on some elements, specifically Buttons, that sometimes the focus outline does not always render evenly. It is not always consistent either, so I'm not exactly sure what is causing it, but something I've noticed in the past, so not related to this work.

I've enclosed a video that captures what I've seen:
https://user-images.githubusercontent.com/7331038/124981870-10b94380-e004-11eb-853b-13a2773b54bd.mov

QuintonJason:
> This changes look really great. I don't have any major concerns here. However, I have noticed on some elements, specifically Buttons, that sometimes the focus outline does not always render evenly. It does not consistently render this way either, so I'm not exactly sure what is causing it, but something I've noticed in the past; not related to this work. I initially thought it may be related to the transition, but disabling it still causes the issue. Could be from `calc`? Not sure.
> 
> I've enclosed a video that captures what I've seen:
> https://user-images.githubusercontent.com/7331038/124981870-10b94380-e004-11eb-853b-13a2773b54bd.mov

Ah yes, the infamous subpixel rendering issue! I'll see if I can nab this finally.

monicawheeler (approved):

packages/sage-assets/lib/stylesheets/core/mixins/_sage.scss
```diff
@@ -157,6 +158,7 @@
       transform $outline-animation-speed ease-in-out;
```
monkeypox8, line 158: A longstanding inconsistency that we've had in Safari has been the way that link buttons don't correctly transition the focus outline style. In experimenting with it while testing this branch, I believe I've narrowed it down to the `color` in the `::after` pseudo-element being the offender. If my theory is correct, updating the mixin by adding `color` as a duplicate of `background-color` will solve it:

|  before   |  after (applying `color`)  |
|--------|--------|
|safari-focus-ring|safari-focus-with-color|

Definitely not a blocker, so I'll open a followup if we can't get the fix in here.

monkeypox8 (approved):
👍🏼 LGTM, great work putting all of this together! I do still feel strongly that we should have an `active` state for links & buttons. In the end, that's design's call.

That being said, should we note somewhere that all current browsers (Chrome, FF, Safari) apply these focus styles differently? For example, Chrome and Firefox will display and keep the focus ring after mouseclick/keyup has ended, but Safari will only display it during the click.

<details>
<summary>Chrome mouse click example</summary>

</details>

<details>
<summary>Safari mouse click example</summary>

</details>

docs/lib/sage_rails/app/views/sage_components/_sage_transaction_card.html.erb
QuintonJason: Updates

docs/lib/sage_rails/app/views/sage_components/_sage_banner.html.erb
philschanely: @QuintonJason not blocking, but I do still find this an odd choice. When able, lets talk this out more so i can understand your thought process around this usage. Mine is that using `sage-btn--subtle` apart from in combination with `sage-btn` presents an odd separation form the BEM object's intent and potentially creates a brittle, one-off pairing. IMO if the functionality is similar, I'd isolate it into a generalized mixin to use both places, or try to use the actual `sage-btn in full` (if not, more ideally, the pure `SageButton` component) with a mixin class of` sage-banner__link` so as to preserve the integrity of the `sage-btn`. I am definitely open to be convinced or to understand better what you've done here. LMK.

PR #622, opened 2021-07-12
Pagination Suffix Adjustment

pixelflips:
## Description
Addresses a bug found during the development of new payments page design in kajabi-products.
The design calls for the suffix text to be displayed even if the page count = 1.
Currently, the suffix is only displayed if the count is > 1.

## Screenshots
There will be no visual updates with these changes.

## Testing in `sage-lib`
Verify code updates allow for suffix to be displayed if page count = 1.

1. (**LOW**) Allows pagination suffix to be displayed if page count = 1.
   - [ ] Should not affect any current implementation but a quick sanity check of any page that has pagination should be done.


joedakroub (approved):
Looks good to me Phillip! 👍🏼

monkeypox8 (approved):
👍🏼 🔥 LGTM!

PR #623, opened 2021-07-13
Passing down common props

philschanely:
## Description

This PR adds getter methods to SageComponent for common properties so that they can be accessed within the component template, particularly for the sake of passing them down to child components. The following can now be accessed within a template view

- `css_classes` (previously only compiled into `generated_css_classes`)
- `html_attributes` (previously only compiled into `generated_html_attributes`)
- `spacer` (previously only compiled into `generated_css_classes`)
- `test_id` (previously only compiled into `generated_html_attributes`)

## Testing in `sage-lib`

Validate any components still render these props as expected

(LOW/MEDIUM/HIGH/BREAKING) Internal change to root SageComponent making properties available to templates. No changes expected in components.

### DIFF
--- a/docs/lib/sage_rails/app/sage_components/sage_component.rb
+++ b/docs/lib/sage_rails/app/sage_components/sage_component.rb
@@ -18,7 +18,12 @@ class SageComponent
     @generated_html_attributes ||= ""
   end
 
+  def css_classes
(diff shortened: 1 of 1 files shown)


joedakroub (approved):
These changes look good to me Phil; very nice update! 👍🏼

PR #624, opened 2021-07-13
Panel figure enhancements

philschanely:
## Description

This PR separates the Panel Figure component from the Panel page into its own page to improve documentation and adds the following new features (non-destructive):

- Images can be set up with a specific aspect ratio.
- A background color can be provided for behind the image.
- Images can be padded to reveal a small "frame" of the background color.

## Screenshots

Screen Shot 2021-07-13 at 12 04 35 PM

Screen Shot 2021-07-13 at 12 04 53 PM

Screen Shot 2021-07-13 at 12 05 05 PM

## Testing in `sage-lib`

See new Panel Figure page (Rails only ATM).

## Testing in `kajabi-products`

(LOW) New customization options added to Panel Figure. No impact on current uses.

PR #625, opened 2021-07-14
fix(panel figure): corrects an invalid calc default value

monkeypox8:
## Description
An invalid default 'auto' value in a css `calc` function is causing webpack builds to fail in our CI. This hotfix modifies the css custom property to an integer value rather than a keyword.

There must be a significant difference in configuration between feature branches and `master` because this didn't get flagged until the docs deployment was triggered.

## Screenshots

<details>
<summary>CircleCI</summary>
<a href="https://app.circleci.com/pipelines/github/Kajabi/sage-lib/3023/workflows/ff340172-9a0b-4886-b203-76e789e2308a/jobs/3249" target="_top">View full output</a>

</details>

## Testing in `sage-lib`
- Ensure panel figure displays and functions as expected

- (LOW) New customization options added to Panel Figure. No impact on current uses.

## Related
#624

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_panel.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_panel.scss
@@ -7,7 +7,7 @@ $-panel-figure-padding: sage-spacing(lg);
 $-panel-figure-default-background-color: sage-color(grey, 100);
 
 :root {
-  --sage-figure-aspect-ratio: auto;
+  --sage-figure-aspect-ratio: 1;
   --sage-figure-background-color: #{$-panel-figure-default-background-color};
 }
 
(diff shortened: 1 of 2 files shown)


pixelflips (approved):
👍🏼

QuintonJason (approved):

philschanely:
@monkeypox8 I hope you don't mind I pushed a quick tweak to this as I had these vars in the wrong/prior file and I thought the 16:9 ratio should probably be the default versus 1. Thankfully, nothing gets activated unless `--aspect-ratio` is on, which should mean someone has provided the desired ratio, so it ought not to matter too much.

monkeypox8:
> @monkeypox8 I hope you don't mind I pushed a quick tweak to this as I had these vars in the wrong/prior file and I thought the 16:9 ratio should probably be the default versus 1. Thankfully, nothing gets activated unless `--aspect-ratio` is on, which should mean someone has provided the desired ratio, so it ought not to matter too much.

Not at all! Thanks for making the change. 👍🏼

PR #627, opened 2021-07-14
Make link truncation optional

philschanely:
## Description

This PR patches some potential issues with recent additions of truncation classes to SageLink component.

## Testing in `sage-lib`

The following pages use SageLink and should continue to function as expected:

- Link
- Transaction Card
- Catalog Item
- Preview for Sortable 

## Testing in `kajabi-products`
1. (LOW) Ensures truncation in Links can be set when desired. Affected areas include Transaction Card (not public), and the following:
   - [ ] Catalog Items on Products index page
   - [ ] Catalog Items on Coaching index page
   - [ ] Per prior bug reported, Forms index page links in table should now wrap again as expected

PR #629, opened 2021-07-15
fix(link): adjust template logic for sagelink to correct spec failure

monkeypox8:
## Description
Hotfix to address spec failures on `kajabi-products` ([full list here](https://app.circleci.com/pipelines/github/Kajabi/kajabi-products/53001/workflows/ef5f60e4-4356-4530-a22b-2964d4f341c1/jobs/201874/parallel-runs/0?filterBy=FAILED)) from sage version bump. Modifies conditional to check for presence of optional `label` property before rendering.

Example:
```
1) Form submission offer grants existing user with offer
     # The example failed, but another attempt will be done to rule out flakiness

     Failure/Error: <%= sage_component SageLink, { help_link: true, url: ZendeskHelpLink[:double_optin].to_s, attributes: { rel: "noreferrer noopener", target: "_blank" } } %>

     ActionView::Template::Error:
       undefined method `html_safe' for nil:NilClass
```

### Related
- #627

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
@@ -1,4 +1,5 @@
 <%
+  label = ""
   if component.label.present?
     label = if component.truncate.present?
       %(
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
@@ -1,12 +1,15 @@
 <%
-label = component.label.html_safe
-if component.truncate
-  label = %(
-    <span class="t-sage--truncate">
-      #{component.label}
-    </span>
-  ).html_safe
-end
+  if component.label.present?
+    label = if component.truncate.present?
+      %(
+        <span class="t-sage--truncate">
+          #{component.label}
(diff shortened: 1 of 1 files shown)


pixelflips (approved):

QuintonJason (approved):

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
```diff
@@ -1,12 +1,15 @@
 <%
-label = component.label.html_safe
-if component.truncate
-  label = %(
-    <span class="t-sage--truncate">
-      #{component.label}
-    </span>
-  ).html_safe
-end
+  if component.label.present?
```
philschanely, line 2: It may be a little more foolproof to ensure that `label` is assigned an initial value first such as:
```erb
<%
label = ""
if component.label.present?
  ...
end
%>
```

philschanely (approved):
Glad this solves the failing specs. One suggestion before merging.

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
QuintonJason: Verified that specs are unaffected. Pushed update and will merge after tests pass.

Issue #524, opened 2021-05-20
NEW: Grid items


philschanely:
## Description

Add new component and subcomponents as needed per specs in Sage 3.0.

**Links**
- [Figma specs](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/WIP-Sage-3-%E2%80%94-Admin-Components?node-id=3655%3A22416)

PR #632, opened 2021-07-15
NEW: Media tile

philschanely:
## Description

This PR adds the new Media Tile (previously called Grid Panel in Sage 3.0 docs).

- Tiles have a minimum size matching the small container size. They will therefor wrap indepenently as viewport increases until up to three fit together in a line. Then they will enlarge as the container allows.
- Include a variety of custom content areas including media, title, actions, body, caption and footer. These can be composed through actions or `content_for` slots. See Properties for more.
- Allow for passing additional configs to the media item through `media_configs` according to the Panel Figure specs.

Follow on task will add this component for React as well.

## Screenshots

[Design specs](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/?node-id=6999%3A21709)

Screen Shot 2021-07-16 at 4 12 22 PM

## Testing in `sage-lib`

See new Media Tile page under Components

1. (Low) New Media Tile component has no impact.

## Related

#524


packages/sage-assets/lib/stylesheets/core/functions/_utilities.scss
```diff
@@ -69,6 +68,35 @@
   }
 }
 
+///
```
monicawheeler, line 71: Could you provide a usecase in the comments of this file? Similiar to what we are doing for `rem(value)`? 

[rem function code example](https://github.com/Kajabi/sage-lib/blob/deddd4832b1ff88fcdc2e69dd6cf1a0f3d62e2b1/packages/sage-assets/lib/stylesheets/core/functions/_utilities.scss#L105)

docs/lib/sage_rails/app/views/sage_components/_sage_media_tile.html.erb
```diff
@@ -0,0 +1,85 @@
+<%
+media_aspect_ratio = nil
```
monicawheeler, line 2: Super minor thing on this file, does the spacing need to be corrected for lines 2 through 23?

packages/sage-assets/lib/stylesheets/components/_media_tile.scss
```diff
@@ -0,0 +1,49 @@
+////
+/// Media Tile
+///
+/// @group sage
+////
+
+$-media-tile-max-col-size: relative-column(1, 3, sage-spacing(stage));
+$-media-tile-min-col-size: sage-container(sm);
+
+// .sage-media-tile {}
```
monicawheeler, line 10: Do we need to remove the commented line here? If you feel like we should leave it, perhaps a supporting comment around why it's present?

packages/sage-assets/lib/stylesheets/components/_media_tile.scss
```diff
@@ -0,0 +1,49 @@
...
+
+.sage-media-tile__body {
+  @include clamp(2);
```
monicawheeler, line 13: Is this too prescriptive? Could we ever see a time where more lines are needed? Should a clamping option be built into the component?

docs/app/views/examples/components/media_tile/_preview.html.erb
```diff
@@ -0,0 +1,206 @@
+<%
+common_action_options = [
```
monicawheeler, line 2: I'm having a hard time with this naming convention, specifically in using it in the wild and checking the design system for the usecase. What do you think about renaming `common_action_options` to `dropdown_action_options` or `more_action_options`. And then `common_additional_actions` could be `icon_action_options` or something like that? Something that makes it a bit more descriptive when using the component?

docs/app/views/examples/components/media_tile/_preview.html.erb
```diff
@@ -0,0 +1,206 @@
...
+  items: [
+    {
+      title: "Plain and simple",
+      body: md('Tiles do not have to include an actions area if its not needed.'),
+      footer: common_footer,
+    },
+    {
+      title: "Actions dropdown",
+      body: md('Since an "options" dropdown is common, you can easily pass a list of items to this default slot.'),
+      action_options: common_action_options,
```
monicawheeler, line 142: `action_options` and `actions` feel a bit vague. Is it possible to rename these to something a bit more descriptive, like `dropdown_options` and `icon_actions` ? What do you think?

packages/sage-assets/lib/stylesheets/core/functions/_utilities.scss
philschanely: Yes. On it.

docs/lib/sage_rails/app/views/sage_components/_sage_media_tile.html.erb
philschanely: I believe its correct to have it set to the left edge unless there's a block. Am I missing something?

packages/sage-assets/lib/stylesheets/components/_media_tile.scss
philschanely: Good question. I'll take this to the design folks. I had these instructions from my designer, but its possible we'd want something that fits broader contexts.

docs/app/views/examples/components/media_tile/_preview.html.erb
philschanely: I see your point. I do think some prefix such as `actions` that gives context is helpful. What about these?

- `actions_dropdown_items` for the dropdown items
- `custom_actions` for the additional actions content that can be provided in addition to or instead of the dropdown? (this can be any content desired but may most often be icon buttons)

Is there another prefix/base that makes more sense to you?

docs/lib/sage_rails/app/sage_components/sage_schemas.rb
```diff
@@ -20,6 +20,33 @@ module SageSchemas
...
+    footer: [:optional, NilClass, String],
+    media: [:optional, NilClass, String],
+    media_configs: [:optional, NilClass, {
+      aspect_ratio: [:optional, NilClass, Integer, Float, Set.new(["1x", "2x", "3x"])],
+      is_wistia: [:optional, TrueClass],
+      key_color: [:optional, NilClass, String],
+      padded: [:optional, NilClass, TrueClass],
+    }],
+    title: [:optional, NilClass, String],
+  }
```
joedakroub, line 48: I like that the schemas are separated into their own file. Great idea.

docs/lib/sage_rails/app/views/sage_components/_sage_media_tile.html.erb
```diff
@@ -0,0 +1,85 @@
...
+  <% end %>
+
+  <%= sage_component SagePanelFooter, {
+    align_spread: true,
+    css_classes: "sage-media-tile__footer"
+  } do %>
+    <%= content_for :sage_media_tile_footer if content_for? :sage_media_tile_footer %>
+    <%= component.footer.html_safe if component.footer.present? %>
+  <% end if component.footer.present? or content_for? :sage_media_tile_footer %>
+<% end %>
```
joedakroub, line 85: Nice to see that the number of components we have can be used to compose new ones so easily. 😍

joedakroub (approved):
Some really nice work here Phil. This is a great addition to our component collection. Other than the suggestions Monica made, I don't have much to add. Looking forward to seeing this in use! 👍🏼

docs/app/views/examples/components/media_tile/_preview.html.erb
monicawheeler: I like these. Any opposition to:
`actions_dropdown_items`
`actions_custom`

docs/lib/sage_rails/app/views/sage_components/_sage_media_tile.html.erb
monicawheeler: It's my inclination to see it indented because it appears nested, but I see that it's not :) 👍🏻

docs/lib/sage_rails/app/sage_components/sage_schemas.rb
```diff
@@ -20,6 +20,33 @@ module SageSchemas
...
+
+  MEDIA_TILE = {
+    action_options: [:optional, [[SageSchemas::DROPDOWN_ITEM]]],
+    actions: [:optional, NilClass, String],
+    body: [:optional, NilClass, String],
+    caption: [:optional, NilClass, String],
+    footer: [:optional, NilClass, String],
+    media: [:optional, NilClass, String],
+    media_configs: [:optional, NilClass, {
+      aspect_ratio: [:optional, NilClass, Integer, Float, Set.new(["1x", "2x", "3x"])],
```
joedakroub, line 42: Though there are only 3 options, the `aspect_ratio` values are a little confusing. Not sure if the `x` designation is needed; I imagine it being like a multiplier which isn't really true in this case. Maybe the `x` is not necessary? Just a thought.

philschanely: Fair. The 2x and 3x are multiplicators of the standard content area height in principle (not literally tied together). But def not related with 1x... how about we make that "natural", make "2x" be "medium", and make "3x" be "tall"? 

FWIW here's the reference (which we need to work to bring into the central spec file as its currently in our feature design file):

Screen Shot 2021-07-16 at 11 22 03 AM

philschanely: (Bear in mind we can also just provide a desired custom aspect ratio beyond these presets)

joedakroub: The image is always the same width, correct? Wondering if `short`, `medium` and `tall` make sense in that case?

joedakroub: Given that we can also provide a custom aspect ratio, maybe the `1x`, `2x` and `3x` designations aren't an issue. Definitely would like to hear what others think.

philschanely: On further thought I think I will pull these out of the component. I am not sure they will be useful or relevant beyond the context we're needing them in Manage and we can send in the particular aspect ratios we want there rather than burden the component with this specificity.

packages/sage-assets/lib/stylesheets/components/_media_tile.scss
philschanely: I'm going to pull these for folks to add clamping as they need it rather than enforce it here.

packages/sage-assets/lib/stylesheets/core/functions/_utilities.scss
philschanely: I'm going to pull this function as it is no longer needed here and may not be relevant elsewhere either.

docs/lib/sage_rails/app/views/sage_components/_sage_media_tile.html.erb
monkeypox8: Agreed! A wonderful example of how modular Sage components can be. 🔥

packages/sage-assets/lib/stylesheets/components/_media_tile.scss
```diff
@@ -0,0 +1,66 @@
...
+    @extend %t-sage-body-small;
+  }
+}
+
+.sage-media-tile__content {
+  align-content: start;
+}
+
+.sage-media-tile__footer {
+  align-self: end;
```
monkeypox8, line 42: I may be missing additional context here, but this appears to be unnecessary due to the inherited grid template.

monkeypox8 (approved):
👍🏼 LGTM! A few minor items to consider but `alt` attributes aside, no blockers. Nicely done!

packages/sage-assets/lib/stylesheets/components/_media_tile.scss
philschanely: I'll double-check but I put this so that the footer would always go to the very bottom of the tile in a row; other items are aligning start IIRC.

docs/app/views/examples/components/media_tile/_preview.html.erb
```diff
@@ -0,0 +1,144 @@
...
+    body: %(
+      <p>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+        Etiam rhoncus aliquam odio.
+      </p>
+    ),
+    caption: %(
+      <p>Vivamus dictum rutrum dui, nec placerat ante</p>
+    ),
+    media: %(<img src="//source.unsplash.com/Ssj4uWXcvS4/600x400?1" />),
```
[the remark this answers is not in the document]
philschanely: Def. thanks!

packages/sage-assets/lib/stylesheets/components/_media_tile.scss
```diff
@@ -0,0 +1,66 @@
...
+$-media-tile-min-col-size: sage-container(sm);
+$-media-tile-breakpoints: (
+  1-col-max: rem(1079px),
+  2-col-min: rem(1080px),
+  2-col-max: rem(1443px),
+  3-col-min: rem(1444px),
+);
+
+.sage-media-tile {
+  min-width: $-media-tile-min-col-size;
```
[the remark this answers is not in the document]
philschanely: Yeah, good call.

docs/app/views/examples/components/media_tile/_props.html.erb
```diff
@@ -0,0 +1,73 @@
+
+<tr>
+  <td colspan="4">
+    <strong>MediaTile</strong>
+  </td>
+</tr>
+<tr>
+  <td><%= md('`actions_dropdown_items`') %></td>
+  <td><%= md('Provide a list of dropdown items to be used with an actions dropdown menu in the actions slot next to the title of the tile.
+  This drodpown will appear on the far right of any other actions provided through `actions_custom` property or the `:sage_media_tile_actions_custom` slot.') %></td>
```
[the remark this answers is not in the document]
philschanely: Noep. Drodpown is the corretc splleing. 😊 

... will fix.

packages/sage-assets/lib/stylesheets/components/_media_tile.scss
philschanely: Confirmed this is needed for the desired effect.

PR #639, opened 2021-07-19
[Button] add spinner to loading state

QuintonJason:
## Description

- [x] add a spinner to the loading state of the button
- [ ] ~update react version~ Storybook is jacked up, will address in a follow up

## Screenshots

button-loading-state

## Testing in `sage-lib`

Check the button rails page and click on the loading state example

1. (LOW) Does not exist in the app and will cause on adverse affect


packages/sage-system/lib/button-disable.js
```diff
@@ -0,0 +1,43 @@
...
+  const SELECTOR_BUTTON_DISABLE_ON_SUBMIT = "data-js-sage-disable-on-submit"
+  const BUTTON_LOADING_SPINNER = `<svg class='sage-loader__spinner sage-loader__spinner--loading-button' viewBox='25 25 50 50' aria-hidden='true'><circle class='sage-loader__spinner-path sage-loader__spinner-path--loading-button' cx='50' cy='50' r='20' fill='none' stroke='0072EF' stroke-width='4'></circle></svg>`;
+  const ATTRIBUTE_ARIA_LABEL = 'aria-label';
+
+  // ==================================================
+  // Functions
+  // ==================================================
+
+  function renderDisabledButtonSpinner(evt) {
+    console.log('evt', evt);
```
monkeypox8, line 15: Can remove this log

docs/app/views/examples/components/button/_props.html.erb
```diff
@@ -33,6 +33,12 @@
   <td><%= md('Boolean') %></td>
   <td><%= md('`false`') %></td>
 </tr>
+<tr>
+  <td><%= md('`disabled_on_submit`') %></td>
+  <td><%= md('When the button is in a loading state, this test will be set as the `aria-label`.') %></td>
```
monkeypox8, line 38: Small typo: should this be "text" rather than "test"?

docs/app/views/pages/sandbox.html.erb
```diff
@@ -4,4 +4,28 @@
 
 Use this scratchpad to test out elements and components in context. Just be sure not to commit your changes. ❤️
 ") %>
+<div>
```
monkeypox8, line 7: We'll want to revert the scratchpad here

packages/sage-system/lib/button-disable.js
```diff
@@ -0,0 +1,43 @@
...
+    let currentTarget = evt.currentTarget;
+    let spinnerLabel = currentTarget.getAttribute(SELECTOR_BUTTON_DISABLE_ON_SUBMIT);
+
+    currentTarget.classList.add('sage-btn--is-loading');
+    currentTarget.setAttribute(ATTRIBUTE_ARIA_LABEL, spinnerLabel)
```
monkeypox8, line 20: Should we also set `aria-busy="true"` and `aria-live` when the spinner is displayed? Though I'm not sure if it'd be as useful on a small component that (hopefully) changes state quickly vs a lazyloading section where I'd expect that pattern to appear.

The updated toast component seems like it would benefit, however.

packages/sage-assets/lib/stylesheets/components/_toast.scss
```diff
@@ -115,19 +115,14 @@ $-toast-bottom-spacing: sage-spacing(xs);
...
   margin-right: sage-spacing(xs);
-  animation: rotate $-loader-spinner-speed linear infinite;
 }
 
-.sage-toast__spinner-path {
-  stroke-dasharray: 150,200;
-  stroke-dashoffset: -10;
-  stroke-linecap: round;
+.sage-loader__spinner-path--loading-button {
   stroke: sage-color(white);
```
monkeypox8, line 125: Can this be set to `currentcolor` instead? That way, the SVG will inherit the text color, which should help on lighter backgrounds like the `secondary` button

monkeypox8 (commented):
Added some minor cleanup items to consider, but this looks 🔥!!

monkeypox8 (approved):
whoops, meant to approve and not comment

packages/sage-system/lib/button-disable.js
```diff
@@ -0,0 +1,43 @@
...
+
+    let element = document.createElement('div');
```
philschanely, line 26: Is the div wrapper needed? If so, might a `span` be preferrable given the context of possibly a `button` or a hyperlink?

packages/sage-system/lib/button-disable.js
```diff
@@ -0,0 +1,43 @@
...
+    element.innerHTML = BUTTON_LOADING_SPINNER;
+
+    currentTarget.appendChild(element)
+
+    // return currentTarget.innerHTML = BUTTON_LOADING_SPINNER;
```
philschanely: Code remnant here pertinent to my prior question. 🤔 ?

philschanely (commented):
I am super fan of this initiative. Great start here for sure! 🙌  

I have a few additional thoughts/concerns that may be worth considering before we push this out:

1. In my local testing the spinner appears but the button remains in enabled styling. If the prop is called "disable-on-submit" should the style change to disabled as well? or perhaps this could be `spinner-on-submit`?
2. Related to prior point, I can imagine cases where we may want `disabled_on_submit` and `spinner_on_submit` to be separate concepts. IE immediately show button as disabled, and possibly with a new label ("Submitting...") versus showing the spinner. Any thought towards setting up the two effects?
3. Regardless of revisions above, could we have a way to signal toggling this off/back to initial state? For example, if an asynchronous submission finishes and for some reason does not redirect the page, such as to show validation errors, or for an in-context save button that doesn't redirect after finished. Something like a callback, or a custom event this component watches for to signal it to return to default?

Super excited about this feature as we clearly need it given past precedence with submit functionality in `products`! 🥇

QuintonJason:
> I am super fan of this initiative. Great start here for sure! 🙌
> 
> I have a few additional thoughts/concerns that may be worth considering before we push this out:
> 
> 1. In my local testing the spinner appears but the button remains in enabled styling. If the prop is called "disable-on-submit" should the style change to disabled as well? or perhaps this could be `spinner-on-submit`?
> 2. Related to prior point, I can imagine cases where we may want `disabled_on_submit` and `spinner_on_submit` to be separate concepts. IE immediately show button as disabled, and possibly with a new label ("Submitting...") versus showing the spinner. Any thought towards setting up the two effects?
> 3. Regardless of revisions above, could we have a way to signal toggling this off/back to initial state? For example, if an asynchronous submission finishes and for some reason does not redirect the page, such as to show validation errors, or for an in-context save button that doesn't redirect after finished. Something like a callback, or a custom event this component watches for to signal it to return to default?
> 
> Super excited about this feature as we clearly need it given past precedence with submit functionality in `products`! 🥇

Update regarding your comments: 

1. In the designs, the loading state has the primary blue that exists on the default primary button
2. I updated the references to `spinner_on_submit`. Currently the ("Submitting...") is the `railsui` component. This approach is a different approach that doesn't currently exist. We'd need to add these per instance. The idea would be to get this one merge and pair with @AndrwM regarding implementation with `kajabi-products`, but only with the agenda items view in the build domain.
3. I'll have to pair with @AndrwM on this one as well. But this one may need to be a followup item. The case we are working through are form submit buttons in rails. After the button is clicked, it then becomes disabled with a text swap. After the js has fired, that state remain until a page refresh is triggered. Since our current implementation has the page refresh, we didn't have that need in the initial implementation. Is this ok if we handle in a followup after Andrew sees how the current implementation will work within the app @philschanely?

philschanely:
@QuintonJason fabulous and thanks for these modifications. 1. clear; 2. do you think the data-js- prop should also adjust in line with the `spinner-on-submit` direction? 3. Yes; go ahead and a follow up would be just fine. 🙌

QuintonJason:
> @QuintonJason fabulous and thanks for these modifications. 1. clear; 2. do you think the data-js- prop should also adjust in line with the `spinner-on-submit` direction? 3. Yes; go ahead and a follow up would be just fine. 🙌

@philschanely regarding 2 I've updated remaining instances of `disable` to `spinner`.

philschanely (approved):
❤️  Nice work @QuintonJason

PR #642, opened 2021-07-20
Add content to media tiles

philschanely:
## Description

This quick patch for media tiles adds the ability to compose individual `MediaTile` items as you like inside a `MediaTiles` wrapper.

1. (Low) Quick patch for Media Tiles; no impact


QuintonJason (approved):
Testing with implementation for Build domain

PR #644, opened 2021-07-21
Icon rounded variation

philschanely:
## Description

This PR adds ability to make an Icon that has the `card_color` provided to display with a circular frame.

## Screenshots

Screen Shot 2021-07-21 at 11 44 46 AM

## Testing in `sage-lib`

See Components > Icon and corresponding entry in Storybook.

1. (LOW) Add circular variation to icon with background. No impact on existing uses of SageIcon (Rails) or Icon (React).

### DIFF
--- a/docs/lib/sage_rails/app/sage_components/sage_icon.rb
+++ b/docs/lib/sage_rails/app/sage_components/sage_icon.rb
@@ -1,9 +1,10 @@
 class SageIcon < SageComponent
   set_attribute_schema({
-    color: [:optional, NilClass, SageSchemas::COLOR_SLIDER],
     card_color: [:optional, SageSchemas::STATUSES],
+    circular: [:optional, NilClass, TrueClass],
+    color: [:optional, NilClass, SageSchemas::COLOR_SLIDER],
     icon: SageSchemas::ICON,
     label: [:optional, NilClass, String],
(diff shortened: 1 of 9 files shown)


docs/lib/sage_rails/app/sage_components/sage_icon.rb
```diff
@@ -1,9 +1,10 @@
 class SageIcon < SageComponent
   set_attribute_schema({
-    color: [:optional, NilClass, SageSchemas::COLOR_SLIDER],
     card_color: [:optional, SageSchemas::STATUSES],
+    circular: [:optional, NilClass, TrueClass],
```
pixelflips, line 4: Just a small nit-pick, but in the `SageIconCard` component we refer to this attribute as "round" vs. "circular" - do you think we should keep the naming consistent?

pixelflips (approved):
One non-blocking comment/question, but LGTM! Nice work! 👍🏼

docs/lib/sage_rails/app/sage_components/sage_icon.rb
philschanely: Good question. I had `rounded` at first then realized we do have rounded corners already. Then `circular` felt more specific. Also, we will likely want to refactor and deprecate the IconCard component eventually too. 🤔

PR #648, opened 2021-07-22
Expand where spacing is reset for form fields

philschanely:
## Description

This PR expands the contexts in which default form field (textarea, input, checkbox, switch, radio, and select) have their default margin reset to include with inside `sage-card-grid` and `sage-panel-grid`. The use case is now such that one can add `SageClassnames::GRID_PANEL` or  `SageClassnames::GRID_CARD` as a CSS class on `sage_form_for` and thereby apply the appropriate grid spacing to the container elements:

```
<%= sage_form_for ... html: { class: SageClassnames::GRID_PANEL } ... do %>
  <%= f.input :title ... %>
  <%= f.input :caption ... %>
<% end %>
```

## Testing in `sage-lib`

See default spacing on fields is unaffected outside of these contexts.

1. (LOW) New contexts for reseting spacing on form fields has no impact on existing uses.

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_checkbox.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_checkbox.scss
@@ -44,9 +44,11 @@ $-checkbox-focus-outline-color: sage-color(primary);
     margin-bottom: sage-spacing(card) / 2;
   }
 
+  .sage-panel-grid > &,
   .sage-panel__row > &,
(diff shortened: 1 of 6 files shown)


monkeypox8 (approved):
👍🏼 LGTM, a very welcome addition! I'm curious if and how we might demonstrate example form layouts in the docs site, since `sage_form_for` is an extension of products.

PR #657, opened 2021-07-26
Image tag conversion and image path updates

joedakroub:
## Description
This PR does the following:
- Updates all instances of `` to use `image_tag` for all Rails components
- Updates all image paths in the docs to NOT start with `/assets/`

Image references are currently broken in the production version of Sage docs. 

In order to correct this, we must use the `image_tag` helper which will correctly assign a hash to all images for caching purposes.

In addition, it should be noted that when referencing images in the docs folder, the path MUST NOT includes `/assets/` as the `image_tag` helper knows to look in that directory by default. While including this in the path will work locally, it does not work when deployed.

**Going forward, we should always use `image_tag` for creating inline images for Rails components.**

## Testing in `sage-lib`
Visit the following URLs and ensure that each image appears correctly, has a hash appended to it when inspecting it and that the `alt` attribute appears where appropriate. For some components, if no `alt` attribute is provided, the `aria-hidden: true` attribute appears (`Feature Toggle`, `Hero` and `Modal - Heading`).

- http://0.0.0.0:4000/pages/component/avatar
- http://0.0.0.0:4000/pages/component/billboard
- http://0.0.0.0:4000/pages/component/catalog_item
- http://0.0.0.0:4000/pages/component/choice
- http://0.0.0.0:4000/pages/component/description
- http://0.0.0.0:4000/pages/component/feature_toggle
- http://0.0.0.0:4000/pages/component/hero
- http://0.0.0.0:4000/pages/component/page_heading
- http://0.0.0.0:4000/pages/component/popover
- http://0.0.0.0:4000/pages/component/sortable

This PR touches a number of components and should be tested in the following areas (**there may be others that I am not aware of...**)
1. (**MEDIUM**) The way images are generated has changed, but the output should be the same.
   - [ ] Settings -> Labs
   - [ ] Products -> All Products

## Related
N/A

### DIFF
--- a/docs/app/views/examples/components/avatar/_preview.html.erb
+++ b/docs/app/views/examples/components/avatar/_preview.html.erb
@@ -16,7 +16,7 @@
   <%= sage_component SageAvatar, {
     image: {
       alt: "Court's profile image",
-      src: "/assets/avatar/court.png"
+      src: "avatar/court.png"
     }
   } %>
 </div>
--- a/docs/app/views/examples/components/billboard/_preview.html.erb
+++ b/docs/app/views/examples/components/billboard/_preview.html.erb
@@ -1,7 +1,7 @@
 <div class="sage-row" style="height:500px">
   <div class="sage-col--md-6">
     <%= sage_component SageBillboard, {
-      img: "/assets/billboard/background_1.jpg",
+      img: "billboard/background_1.jpg",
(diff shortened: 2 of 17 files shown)


pixelflips (approved):
LGTM! 👍🏼  - The images are now rendering as expected and with the names (including random string) as expected. Great work!! Thanks for taking care of those Joe!

philschanely (approved):

monicawheeler (approved):
🙌🏻

PR #659, opened 2021-07-27
Panel tile spacing

philschanely:
## Description

Removes unnecessary margin and padding resets from this component.

Found that spacers could not apply due to conflicts these extra settings caused.

## Testing in `sage-lib`

See no change in Components > Panel > Tiles area.

1. (LOW) Removes excess styling from Panel Tiles. No impact on existing uses.


joedakroub (approved):

pixelflips (approved):
LGTM 👍🏼

PR #671, opened 2021-07-30
Dropdown border top and schema cleanup

### LINKED ISSUE #669: [BUG] Broder-top is missing from DropdownItems in Rails
## Detailed description

Rails Dropdown item `border_top` does not appear as expected.

`SageDropdown` in Rails; React version works as expected.

`border_before` is supposed to add a modifier class of `sage-dropdown__item--border-before` on that respective dropdown item that displays a border above the item.

All browsers same

## Steps to reproduce
1. Add a SageDropdown and provide at least one item with `border_before` set to `true`.
2. Observe in browser that no border appears.

## Screenshots

N/A

philschanely:
## Description

Closes #669 

The PR removes the `border_top` property from the `items` in `SageDropdown` as it was not being applied to the items as expected. instead, the `modifiers` allowed for `border-top` to be provided. Took the opportunity to tighten the Schema for `modifiers` to explicitly list the possible string values that can be passed in this array. Also added `css_classes` to the options and tightened the `style` options on items.

Checked this work in the bridge to ensure no regressions since technically nothing changed in the actual affordances other than adding `css_classes` to items.

1. (LOW) Internal updates to schemas in DropdownItems. [This PR](https://github.com/Kajabi/kajabi-products/pull/20137) removed unnecessary properties in use related to this update.

### DIFF
--- a/docs/app/views/examples/components/dropdown/_props.html.erb
+++ b/docs/app/views/examples/components/dropdown/_props.html.erb
@@ -54,7 +54,7 @@
   <td><%= md('```
 Array<{
   attributes: String,
-  css_class: String,
+  css_classes: String,
   icon: String,
   is_heading: Boolean,
   modifiers: [
--- a/docs/app/views/examples/components/dropdown/_props.html.erb
+++ b/docs/app/views/examples/components/dropdown/_props.html.erb
@@ -54,12 +54,14 @@
   <td><%= md('```
 Array<{
   attributes: String,
-  border_before: String,
+  css_class: String,
   icon: String,
   is_heading: Boolean,
-  modifiers: Array,
+  modifiers: [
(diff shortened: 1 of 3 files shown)


docs/app/views/examples/components/dropdown/_props.html.erb
```diff
@@ -54,12 +54,14 @@
   <td><%= md('```
 Array<{
   attributes: String,
-  border_before: String,
+  css_class: String,
```
joedakroub, line 57: Should this be `css_classes`?

joedakroub (approved):
Aside from the on issue mentioned, this looks great. Nice simplification of props. 👍🏼

pixelflips (approved):
LGTM 👍🏼

PR #678, opened 2021-08-03
JBH Choice React Component

FuturaExtraBold:
## Description
Adds new Choice React component to Sage library.

## Screenshots
image
image
image
image

### React
- View [Storybook](http://localhost:4100/?path=/docs/sage-choice--default)
- Check that component UI exists and nothing is broken
- Check docs

(LOW) Adds new Choice component. No effect on existing work in Kajabi Products. Can be used moving forward.

### DIFF
--- a/packages/sage-react/lib/Choice/Choice.jsx
+++ b/packages/sage-react/lib/Choice/Choice.jsx
@@ -119,6 +119,7 @@ Choice.defaultProps = {
   radioConfigs: null,
   subtext: null,
   target: null,
+  text: null,
   type: null,
   verticalAlignIcon: ICON_ALIGNMENTS.DEFAULT,
 };
@@ -141,7 +142,7 @@ Choice.propTypes = {
   }),
   subtext: PropTypes.string,
   target: PropTypes.string,
-  text: PropTypes.string.isRequired,
+  text: PropTypes.string,
   type: PropTypes.oneOf(Object.values(Choice.TYPES)),
   verticalAlignIcon: PropTypes.oneOf(Object.values(Choice.ICON_ALIGNMENTS))
 };
--- a/packages/sage-react/lib/Choice/Choice.jsx
+++ b/packages/sage-react/lib/Choice/Choice.jsx
@@ -38,6 +38,14 @@ export const Choice = ({
     }
   );
 
+  const contentClassNames = classnames(
+    `${baseClass}__content`,
+    {
+      [`${baseClass}__content--custom`]: children,
+      [`${customContentClassName}`]: customContentClassName,
+    }
+  );
+
   const hasRadioConfigs = !!radioConfigs;
   const isLabel = hasRadioConfigs;
   const isLink = ('href' in rest);
@@ -73,7 +81,7 @@ export const Choice = ({
           <img src={graphic} alt="" />
         </span>
       )}
-      <div className={`sage-choice__content ${customContentClassName}`}>
+      <div className={contentClassNames}>
         {text && (
           <em className="sage-choice__text">
             {text}
--- a/packages/sage-react/lib/Choice/Choice.jsx
+++ b/packages/sage-react/lib/Choice/Choice.jsx
@@ -8,6 +8,8 @@ export const Choice = ({
   className,
   icon,
   isActive,
+  label,
+  subtext,
   type,
   ...rest
 }) => {
@@ -27,7 +29,16 @@ export const Choice = ({
   );
   return (
     <div className={classNames} {...rest}>
-      <h1>Choice!</h1>
+      {label && (
+        <em className="sage-choice__text">
+          {label}
+        </em>
+      )}
+      {subtext && (
+        <span className="sage-choice__subtext">
+          {subtext}
+        </span>
+      )}
     </div>
   );
 };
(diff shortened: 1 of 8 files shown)


packages/sage-react/lib/Choice/Choice.jsx
```diff
@@ -0,0 +1,144 @@
...
+  alignCenter: PropTypes.bool,
+  children: PropTypes.node,
+  className: PropTypes.string,
+  customContentClass: PropTypes.string,
+  disabled: PropTypes.bool,
+  graphic: PropTypes.string,
+  icon: PropTypes.oneOf(Object.values(SageTokens.ICONS)),
+  isActive: PropTypes.bool,
+  linkText: PropTypes.string,
+  radioConfigs: PropTypes.shape({
```
philschanely, line 134: This feature feels like more trouble than its worth I'm sure. I do see this error in the console in Storybook:

Screen Shot 2021-08-12 at 3 58 13 PM

I'd recommend we add `onChange` to alleviate this. You can add it to the expected structure for `radioConfigs` in `propTypes`. It might be a benefit to make this and the other values of `radioConfigs` be required while you're there:

```jsx
Choice.propTypes = {
  // ...
  radioConfigs: PropTypes.shape({
    name: PropTypes.string.isRequired,
    id: PropTypes.string.isRequired,
    value: PropTypes.string.isRequired,
    onChange: PropTypes.func.isRequired,
  }),
  // ...
);
```

Then for `defaultProps` you can probably just let `radioConfigs` be `null` by default rather than setting each item to `null`. The schema you provide in `propTypes` takes care of enforcing the desired structure.
```jsx
Choice.defaultProps = {
  // ...
  radioConfigs: null,
  // ...
);
```

This way the object as a whole is optional but when provided all 4 values should be passed in for a successful configuration.

So now for `hasRadioConfigs` you can probably do this:

```jsx
const hasRadioConfigs = !!radioConfigs;
```

and then you can just use a spread to extract the provided settings onto the radio button:

```jsx
<input
  {...radioConfigs}
  checked={isActive}
  type="radio"
/>
```

Finally, (I think! 😅 ) we need a little conditional on the declaration of `htmlFor` inside `attrs`:

```jsx
const attrs = {
  // ...
  ...(hasRadioConfigs && { htmlFor: radioConfigs.id }),
  // ...
};
```

I'll reach out on Slack with some suggestions for adding stories that help demonstrate the component's use cases.

philschanely (commented):
Looking really good. Just a few things to push on a hair more. Thanks for all your work on this!!

packages/sage-react/lib/Choice/Choice.story.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { selectArgs } from '../story-support/helpers';
+import { SageTokens } from '../configs';
+import { Choice } from './Choice';
+import { ICON_ALIGNMENTS, TYPES } from './configs';
```
philschanely, line 5: Instead of importing these let's use the props you added to `Choice`: `Choice.ICON_ALIGNMENTS` and `Choice.TYPES`.

FuturaExtraBold: Nice catch. Done.

packages/sage-react/lib/Choice/Choice.jsx
```diff
@@ -0,0 +1,144 @@
...
+  target: null,
+  type: null,
+  verticalAlignIcon: ICON_ALIGNMENTS.DEFAULT,
+};
+
+Choice.propTypes = {
+  alignCenter: PropTypes.bool,
+  children: PropTypes.node,
+  className: PropTypes.string,
+  customContentClass: PropTypes.string,
```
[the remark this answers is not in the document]
FuturaExtraBold: Done.

packages/sage-react/lib/Choice/Choice.jsx
FuturaExtraBold: Good catch! Done.

packages/sage-react/lib/Choice/Choice.jsx
```diff
@@ -0,0 +1,144 @@
...
+            checked={isActive}
+          />
+        </div>
+      )}
+      {graphic && (
+        <span className="sage-choice__graphic">
+          <img src={graphic} alt="" />
+        </span>
+      )}
+      <div className={`sage-choice__content ${customContentClass}`}>
```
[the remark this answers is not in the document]
FuturaExtraBold: Went with the second suggestion. Nice!

joedakroub (approved):
This is looking really good Ben; very nice work. No concerns here. 👍🏼

philschanely (approved):
🏅  Fabulous work here @FuturaExtraBold! 🙌

packages/sage-react/lib/index.js
```diff
@@ -5,6 +5,7 @@ export { Breadcrumbs } from './Breadcrumbs';
 export { Button } from './Button';
 export { Card } from './Card';
 // export { Chart } from './Chart';
```
QuintonJason, line 7: you still need this commented code?

philschanely: This was from an earlier quick fix... we'll pull it once we fix the issue

FuturaExtraBold: @QuintonJason This is in `develop`. Looks like @philschanely temporarily pulled it.

QuintonJason (approved):

PR #679, opened 2021-08-04
fix(table): fix classname typo in table header processing

philschanely:
## Description

This PR makes a fix to a typo in Tables where headers parsing schema looked for `classname` rather than `className` which means said classes do not get mapped to table headers.

1. (LOW) Fix to apply className to table headers for React table schemas. Not currently in use in app, but needed.


pixelflips (approved):
LGTM!

PR #680, opened 2021-08-04
Next version bump (~4.11.1+)

philschanely:
1. #679 (LOW) Fix to apply className to table headers for React table schemas. Not currently in use in app, but needed.


pixelflips (approved):
LGTM!

PR #698, opened 2021-08-10
Progress Bar min-width fix

joedakroub:
## Description
This PR removes the `min-width` CSS property from the Progress Bar value element as it is displaying small percentage values incorrectly. For example, if the percentage passed in is `0` the bar will always display 5%.

## Screenshots
|  Before  |  After  |
|--------|--------|
|Before|After|

## Testing in `sage-lib`
1. Visit http://0.0.0.0:4000/pages/component/progress_bar and ensure that the percentage is still displaying correctly.

1. (**LOW**) Ensure that percentages are displaying correctly in:
   - [ ] Podcasts Importing
   - [ ] Products - Cloning a product


philschanely (approved):
No concerns here

PR #704, opened 2021-08-11
SageLabel: remove duplicate quotes around `interactive_type` label

### LINKED ISSUE #702: [BUG] SageDropdown submits form on SageLabel button click
## Detailed description

When used inside a form, clicking a `SageDropdown` label button will submit the form instead of opening the dropdown menu. This seems to be limited to `SageLabel` as the button; other dropdown variants do not appear to be affected.

h/t @connorgallopo 

`SageDropdown` combined with `SageLabel`

The `type="button"` on the rendered `SageLabel` has two sets of quotes, most likely causing the attribute to be invalid and defaulting to `type=submit` of the `button` element

All browsers are affected

## Steps to reproduce
1. 

## Screenshots
sage-dropdown-label

monkeypox8:
## Description

Clicking a SageDropdown label button within a form will submit the form instead of displaying the dropdown. The `type` attribute is being rendered in two sets of quotes, causing it to be invalid or unrecognized by browsers. As a result, the base `button` type of `submit` takes over as the default.

This bug is limited to `SageLabel` when used as the button; other dropdown variants are not affected.

## Screenshots

|  Before  |  After  |
|--------|--------|
|dropdown-label-button-before|dropdown-label-button-after|

## Testing in `sage-lib`

1. Navigate to the [dropdown](http://localhost:4000/pages/component/dropdown) example in the Sage docs
2. Inspect the "SageLabel Button, with SageLabel Values, and aligned right" example
3. Confirm that the element has attribute `type="button"` as expected

1. (**LOW**) Applies corrected button `type` attribute to `SageLabel` when used within a dropdown
   - [ ] Offer upsell (uses ajax and should not be affected; requires feature flag `comm_offer_detail`)
   - [ ] Offer edit - publish status (uses custom markup for dropdown and should not be affected; requires feature flag `comm_offer_detail`)

## Related

- closes #702

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_label.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_label.html.erb
@@ -16,7 +16,7 @@
 >
   <<%= label_content_tag %>
     class="sage-label__value"
-    <%= 'type="button"' if component.interactive_type %>
+    <%= "type=button" if component.interactive_type %>
     <%= component.generated_html_attributes.html_safe %>
   />
     <span class="sage-label__value-text">


connorgallopo (approved):
LGTM

philschanely (approved):

PR #705, opened 2021-08-11
Fix menu icon used in OptionsDropdown

philschanely:
## Description

Recent Icon change removed DOT_MENU in lieu of DOT_MENU_HORIZONTAL but this one component used DOT_MENU... updated in this patch.

1. (LOW) Fixes options menus that would disappear without this icon. Verify options dropdown is visible in:
   - [ ] People table rows at far right


monicawheeler (approved):
👍

joedakroub (approved):
This looks good. 👍🏼

PR #699, opened 2021-08-11
refactor(rails sagecomponent class): Trims the HTML whitespace rendered by SageComponent class

AndrwM:
## Description
This improvement trims the additional html whitespace that is rendered in ERB views. Allows us to format our SageRails html partials however we like, to maximize maintainability/readability, while not introducing janky hard-to-read source code in the web inspector.

When first standing up the SageComponent class we weren't sure the best way to accomplish this. IIRC has been on the improvements wishlist for a while, happy open up this fix!

## Screenshots
|  Before  |  After  |
|--------|--------|
|image|image|

## Testing in `sage-lib`
Ensure nothing is blowing up

## Testing in `kajabi-products`
1. (**LOW**) SageComponent-rendered html white-space changes
   - [ ] If the specs are green this should be fine

## Related
none

PR #706, opened 2021-08-11
fix(rails sagecomponent class): avoid stripping whitespace if contains pre html tags (Fixes #699)

AndrwM:
🙋 **Update:**
Updated this PR with a solution that I think can satisfy the bug that was introduced within #699.
This PR will strip whitespaces for all components that do not contain `"</pre>"` within their returned html. Rather than attempting to use a specific regex to parse this we're relying on Ruby's string method `include?`. See the description in diff for more info!

https://apidock.com/ruby/String/include%3f

Re-requesting review!

## Description
~This change caused whitespace to be clear within `<pre>` tags. I spent the better part of the afternoon testing out a regex solution with limited success. For the time being I'll go ahead and revert. :sadpanda:~

~Ultimately, I'm not sure parsing _every_ Sage components' output with a regex is the right solution. Regexes can be delicate because of the ambiguity of their syntax and the number of various HTML-output variations. Nokigiri may be a better tool to reach for to "post-process" the html coming out of the partial.~

Conversation ref: https://kajabi.slack.com/archives/C01A424HY8Y/p1628709560057600

## Screenshots
n/a

## Testing in `sage-lib`
None, reversion

None, reversion

## Related
#699

### DIFF
--- a/docs/lib/sage_rails/app/sage_components/sage_component.rb
+++ b/docs/lib/sage_rails/app/sage_components/sage_component.rb
@@ -31,10 +31,16 @@ class SageComponent
     context.render(
       partial: "sage_components/#{template_path}",
       locals: { component: self }
-    )
-    .tap { cleanup_section_vars }
-    .squish
-    .html_safe
+    ).tap do |html|
+      cleanup_section_vars
+
+      # We strip additional whitespace out of the partial's html
+      # to allow for cleaner rendering in the browser.
+      # In cases where <pre></pre> html tags exist in the returned html,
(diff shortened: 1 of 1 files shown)


monkeypox8 (approved):
👍🏼 but 😢

philschanely (approved):
Seconding @monkeypox8 😢

philschanely (approved):
🎉

monicawheeler (approved):
Looks great!

joedakroub (approved):

PR #709, opened 2021-08-12
Sage dynamic select input - select2

kajabijamell:
## Description
Adds a select2 input type for sage forms

API Options:
```   
url: String,
label: String,
name: String,
id: String,
has_error: [:optional, TrueClass],
message: [:optional, String],
required: [:optional, TrueClass],
// sets default selected option for the input
default_value: [:optional, String, Integer],
default_text: [:optional, String],
// enables pagination for 25 rows at a time
paginate: [:optional, TrueClass],
// allows user to search through retrieved results
search: [:optional, TrueClass],
// displays an x button for the user to clear the selected value
clear: [:optional, TrueClass],
placeholder: [:optional, String],
// toggles legacy select2 input preset options would be "sage" or "bootstrap"
theme: [:optional, String],
```

### Component States

**Empty state**
image

**Selected state**
image

**Input open w/ search**
image

**Input open w/o search**
image

**input w/ error**
image

## Testing in `sage-lib`

**_IMPORTANT_**

This input has dependencies that are only available in Kajabi Products. There's no way to test this from `sage-lib`.

Testing is straightforward. You will need a URL generated from rails to deliver the select options via `AJAX` request. See component sample code [here](http://localhost:4000/pages/component/dynamic_select). 

Examples of possible URLs can be found in the `select_options` controller found here in Kajabi Products `app/controllers/admin/select_options_controller.rb`.

Once you've found a URL to fetch data, you can then play with the different API attributes to reviews the input in different states. Screenshots are provided above as well.

This PR is additive, meaning it will not affect ant current implementation of the select2 inputs. It may be wise to double-check some existing select2 inputs just to make sure there's no overlap in styles or js.  

1. (**LOW**) Add dynamic select inputs to the sage-lib and does not affect any current components or code structures.


packages/sage-system/lib/dynamicSelect.js
```diff
@@ -0,0 +1,115 @@
+Sage.dynamicSelect = (function() {
```
philschanely, line 1: 😍

packages/sage-assets/lib/stylesheets/components/_dynamic_select.scss
```diff
@@ -0,0 +1,95 @@
...
+$-dynamic-select-selected-tag-border-radius: sage-border(radius-x-large);
+$-dynamic-select-selected-tag-padding: 0 sage-spacing(xs);
+$-dynamic-select-selected-tag-height: rem(30);
+$-dynamic-select-clear-font-weight: sage-font-weight(regular);
+$-dynamic-select-open-arrow: rotate(180deg) scaleX(-1);
+
+// empty state //
+.sage-dynamic-select {
+  position: relative;
+  &__data {
```
philschanely, line 20: Could we please move this out onto its own? We tend to avoid this type of use of `&` in Sage. 

```scss
.sage-dynamic-select {
  // ...
} 

.sage-dynamic-select__data {
  // ...
}

File: packages/sage-assets/lib/stylesheets/components/_dynamic_select.scss [hunk 3]
```diff
@@ -0,0 +1,95 @@
...
+    bottom: sage-spacing(xs);
+    left: 50%;
+  }
+  .select2-container--sage {
```
philschanely, line 24: Does this need to be nested for sufficient specificity, or could this be pulled out to avoid the extra nesting depth?

packages/sage-assets/lib/stylesheets/components/_dynamic_select.scss
```diff
@@ -0,0 +1,95 @@
...
+// selected state //
+.sage-dynamic-select.sage-select--value-selected .select2-container--sage {
+  .sage-select__arrow::before {
+    top: rem(6);
+  }
+  .select2-selection--single.sage-select__field {
+    height: $-dynamic-select-selected-height;
+    padding: $-dynamic-select-selected-padding;
+    border-color: $-dynamic-select-border-color;
+    .select2-selection__rendered {
```
philschanely, line 80: Same here... if not needed, can this be pulled out a layer or more to avoid the deep nesting?

philschanely (commented):
Looking good here, @jamell-kajabi ! 

Regarding nesting depth, if its necessary (which in vendor situations like this would be understandable!) you can add `/* stylelint-disable selector-max-compound-selectors */` before the set of styles that violate the rule and then `/* stylelint-enable selector-max-compound-selectors */` afterwards.

One thing that would be great to add is some documentation around this and how it can be used. This is typically adding a folder and the set of files you can see for others inside `docs/app/views/examples/components` and then an entry for this component in `docs/app/helpers/components_helper.rb`. Then it will appear in the Components sidebar. 

However, I understand following the other examples may not be 100% possible due to the asynchronous nature. It would be great to add the required files, document the `props` from your `sage_dynamic_select.rb` file similar to other components. But then if you can't easily show actual uses in the `preview` file, consider at least providing a code sample like what is shown on the `Toast` component.

I know this is going to be work well invested since this component is used so much in the app now. Thanks SO MUCh for doing this!! 🙇‍♂️

packages/sage-assets/lib/stylesheets/components/_dynamic_select.scss
philschanely: For clarity, the goal is to avoid using `&` for segments of a particular element selector, but OK to use it for a whole element. So:

✅ 
```scss
.sage-dynamic-select {
  // ...
} 

.block__element1 {
  // stuff
  .block--variant & {
    // stuff
  }
}
```

philschanely: And with the BEM naming we'd typically not nest an element selector inside a block selector so that the styles contained in a selector are ideally applied to that actual thing. If specificity or context is needed you bascially flip it so instead of:

```scss
.block {
  &--variant {
    &__element {
      // stuff
    }
  }
}
```

You do:

```scss
.block__element {
  // stuff

  .block--variant & {
    // stuff
  }
}
```

It is more typing (increased chance for typos 😬 ) but helps ensure each selector block is explicitly applying to that thing, thus preventing style leaks. Everything is also in one place for that particular thing too. Hope this makes sense. 👍  but if not please let me know and we can sync up. 🙌

packages/sage-system/lib/dynamicSelect.js
kajabijamell: Thanks, @philschanely, all great points!

packages/sage-assets/lib/stylesheets/components/_dynamic_select.scss
kajabijamell: done :)

packages/sage-assets/lib/stylesheets/components/_dynamic_select.scss
kajabijamell: done 👍

packages/sage-assets/lib/stylesheets/components/_dynamic_select.scss
kajabijamell: +1

kajabijamell:
Thank you for your review @philschanely! I was able to refactor some nesting which definitely helps with readability. I did have som questions regarding adding the input to our docs site, would love to sync with you around this.

docs/lib/sage_rails/app/views/sage_components/_sage_dynamic_select.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  <select
+      class="sage-dynamic-select__data"
+      <%= component.generated_html_attributes.html_safe %>
+      name="<%= component.name %>"
+      id="<%= component.id %>"
+      tabindex="-1"
+      aria-hidden="true"
+      data-js-dynamic-select="true"
+      data-url=<%= component.url %>
+      data-default-value="<%= component.default_value if component.default_value.present? %>"
```
monkeypox8, line 18: Should these data-attributes should be conditionally rendered in their entirely rather than outputting an empty/null value?

For example, could we do something more like:
```
<%= "data-default-text=#{component.data-default-text}" if component.data-default-text.present? %>
```

[select2's docs](https://select2.org/configuration/data-attributes) mention that setting a data-attribute will override the value set in the JS config, and I'm not sure if it's the case with an empty data-attribute. What I'm wondering is if that would cause the empty data-attributes to override view-specific logic if used in a shared partial, for example.

monkeypox8 (commented):
🔥 This is looking great, Jamell! Will be a substantial improvement to our select2 implementation in the app.

In addition to the nesting items that Phil suggested, my comments below are primarily on some cleanup and hopefully clearing up my understanding on derived values.

packages/sage-assets/lib/stylesheets/components/_dynamic_select.scss
```diff
@@ -0,0 +1,95 @@
+$-dynamic-select-default-height: rem(44);
+$-dynamic-select-default-padding: rem(4.4) sage-spacing(sm) 0;
```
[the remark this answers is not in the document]
kajabijamell: I had to make slight adjustments in padding for the input value to fit when the label is placed at the top.

packages/sage-assets/lib/stylesheets/components/_form_select.scss
```diff
@@ -15,7 +15,7 @@ $-select-color-default: map-get($sage-field-colors, default);
 $-select-color-error: map-get($sage-field-colors, error);
 $-select-color-label-background: map-get($sage-field-colors, label-background);
 $-select-color-success: map-get($sage-field-colors, success);
-$-select-filled-top-padding: rem(6);
+$-select-filled-top-padding: rem(4.4);
```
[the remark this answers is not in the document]
kajabijamell: The padding is so small that its pixel value is in the decimals.

docs/lib/sage_rails/app/views/sage_components/_sage_dynamic_select.html.erb
kajabijamell: This is a great point! I'll be sure to implement this change.

packages/sage-assets/lib/stylesheets/components/_dynamic_select.scss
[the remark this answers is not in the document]
philschanely: Could it perhaps round to either `4` or `5` perhaps?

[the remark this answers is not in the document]
kajabijamell: yes, i was able to update the values to whole numbers.

docs/app/views/examples/components/dynamic_select/_preview.html.erb
```diff
@@ -0,0 +1,42 @@
...
+  message: "Error message goes here"
+  url: "Endpoint for API",
+  default_value: "Prefill select input with default value",
+  default_text: "Prefill select input with default title",
+  theme: "Choose Sage or Bootstrap for legacy input",
+  search: "Allow user to search through retrieved results",
+  clear: "Allow user to clear input value"
+
+  // Example Implementation
+  sage_component SageDynamicSelect, {
```
monkeypox8, line 27: As @philschanely had mentioned earlier, ideally we'd have a rendered preview of the component in action, but this works as a substitute.

As a follow-up, I wonder if we could reference the external JS dependencies (jquery and select2) from a CDN just for this preview page. That'd give us a working preview but prevent the dependencies from being packaged up.

docs/app/views/examples/components/dynamic_select/_preview.html.erb
```diff
@@ -0,0 +1,42 @@
+<h3>Working with the<code>Sage Dynamic Select Input (Select2)</code></h3>
+<p>
+  The dynamic select input is a sagified version of the select2 input that resides in <code>Kajabi Products.</code> The input fuctionality leverages the Javascript library select2 and includes several options for tailoring the input to the UI. The input can retrieve options from any arbitrary API that will return validly structured JSON.<br/> <br/> Basic usage and complete documentation ca be found
+  <a href="https://select2.org/getting-started/basic-usage" target="_blank"><code>here</code></a>.
+</p>
+
+<h4>Error Handling</h4>
+<p>Much like other sage inputs, the dynamic select includes options to render error messages from the server. By using the <code>has_error</code> attribute, one can trigger the erros states and print an error message that gets placed below the input.</p>
```
monkeypox8, line 8: Tiny typo here "error states"

docs/app/views/examples/components/dynamic_select/_preview.html.erb
```diff
@@ -0,0 +1,42 @@
+<h3>Working with the<code>Sage Dynamic Select Input (Select2)</code></h3>
+<p>
+  The dynamic select input is a sagified version of the select2 input that resides in <code>Kajabi Products.</code> The input fuctionality leverages the Javascript library select2 and includes several options for tailoring the input to the UI. The input can retrieve options from any arbitrary API that will return validly structured JSON.<br/> <br/> Basic usage and complete documentation ca be found
```
monkeypox8, line 3: 😍 Beautifully written!

One suggestion, we can replace the `<br>` tags with a new `<p>`

packages/sage-assets/lib/stylesheets/components/_dynamic_select.scss
```diff
@@ -0,0 +1,97 @@
+$-dynamic-select-default-height: rem(45);
+$-dynamic-select-default-padding: rem(4) sage-spacing(sm) 0;
+$-dynamic-select-border-color: sage-color(gray, 400);
+$-dynamic-select-placeholder-color: sage-color(gray, 400);
+$-dynamic-select-color-success: map-get($sage-field-colors, success);
+$-dynamic-select-border-box-shadow-size: map-get($sage-field-configs, box-shadow-size);
+$-dynamic-select-selected-height: rem(54);
```
monkeypox8, line 7: We don't have an existing spec in Figma for this, so I won't consider it a blocker but do want to note that the height for the component is much taller (3.375rem) than the standard `SageFormSelect` field (2.25rem) with an option selected. This shouldn't be an issue for standalone fields, but in cases where inputs are horizontally aligned, the difference will be noticeable:

|  SageFormSelect   |  SageDynamicSelect  |
|--------|--------|
|selected-single|selected-select2|

monkeypox8 (approved):
👍🏼 LGTM, a few minor edit tweaks to the preview copy but no major blockers that I can see. After testing with the bridge, I think we may need some additional design input re: the height change for selected values as a follow-up. Great work, this will be a useful addition!

packages/sage-system/lib/dynamicSelect.js
```diff
@@ -0,0 +1,110 @@
...
+        processResults: function (data, params) {
+          var results = data.results ? data.results : data;
+          formattedData = $.map(results, function (obj) {
+            obj.id = obj[this.opts.valueAttribute];
+            return obj;
+          }.bind(this));
+
+          params.page = params.page || 1;
+          $(data).select2.data = formattedData;
+          var showMore = data.totalCount ? (params.page * 25) < data.totalCount : true;
```
joedakroub, line 54: Is the `25` representing the page size? If so, is this something we should parameterize in the `opts` object?

joedakroub (approved):
This is looking really nice Jamell! Thanks so much for putting the time in to convert this to Sage. Will make an awesome addition to our collection. I left one comment, not a blocker. 👍🏼

docs/app/views/examples/components/dynamic_select/_preview.html.erb
philschanely: Note as well that the `md()` helper may be useful to simplify as well.

docs/app/views/examples/components/dynamic_select/_preview.html.erb
philschanely: We may be able to add them as peer dependencies in the repo. @voodooGQ would this make sense?

philschanely (approved):
Looking in great shape, @jamell-kajabi ! Thanks for your hard work on this!

PR #711, opened 2021-08-13
feat(button group): add wrap option

### LINKED ISSUE #690: [BUG] Button group overflow/wrap on small screens
## Detailed description

With small screen devices, button groups do not wrap their content and extend display past container bounds. This seems to occur most frequently with `SageCard` and `SagePanel` grid parents.

`SageButtonGroup`

Expected: buttons wrap for display within containers and do not produce a scrollbar on small-screen devices

Possible solutions to explore:
- Apply `flex` display on button group only when breakpoint is > `medium` or `large`
- Default to `flex-wrap`, reverting on `medium` or `large` breakpoints
- Refactor button right margin using `grid-gap` (now that Safari support is available)
- Determine if `SageCard` and `SagePanel` require the use of grid at small screen breakpoints. Can we default to `block`?

All browsers affected (see screenshots)

## Steps to reproduce
1. View https://sage-design-system.kajabi.com/pages/component/button in any browser
2. Decrease the browser size to < 400px or enable mobile emulation 
3. Verify that button group display does not wrap, and causes the layout to scroll

## Screenshots

<details>
<summary>In app</summary>

</details>

<details>
<summary>Chrome</summary>

</details>

<details>
<summary>Firefox</summary>

</details>

<details>
<summary>Safari</summary>

</details>

## Related
- https://github.com/Kajabi/kajabi-products/pull/19840#pullrequestreview-725695892

monkeypox8:
## Description

Adds optional `wrap` param to `SageButtonGroup`, allowing content to wrap to container widths.

- Sets *all* `SageButtonGroup` containers to `flex-wrap: wrap` (whether the param is enabled or not) to prevent mobile screen widths from horizontally scrolling. For breakpoints larger than `sm-min`, wrapping remains disabled by default
- Replaces use of `margin` with `gap` (as of Safari v14.1/14.7, `gap` is now supported in flex containers for all modern browsers) 
- `margin` has also been replaced with `gap` for `SageChoice` inside a `SageButtonGroup` 

## Screenshots

|  | Before  |  After  |
|---|-----|--------|
| sage example|before|after|
|in app|before|after|

## Testing in `sage-lib`
1. View the [Sage button example page](http://localhost:4000/pages/component/button) and scroll to the "Button group: Allow wrap" section
2. Click the "Toggle button group wrap" button to activate wrapping on all button group examples
3. Decrease the browser window width, and confirm that buttons wrap as expected
4. Click the "Toggle button group wrap" button once more to deactivate wrapping
5. Decrease the browser window width, and confirm that buttons extend past the confines of their container. Note that below the `sm-min` breakpoint width, wrapping is still enabled regardless of the setting (see note in description)

1. (**MED/HIGH**) Adds option for button groups to wrap to multiple lines. Default behavior on mobile views has been changed to allow wrapping (other breakpoints are unaffected), so existing uses should be checked to ensure content still flows as expected. Examples include:
   - [ ] Website Pages and Website Landing Pages: Edit page view - Publish status
   - [ ] Coupons: Delete coupon confirmation modal
   - [ ] Offers: Offer export modal
   - [ ] Email campaigns: Broadcast send type selection

## Related
- closes #690


packages/sage-assets/lib/stylesheets/components/_button.scss
```diff
@@ -483,6 +468,23 @@ $-btn-loading-min-height: rem(36px);
 .sage-btn-group {
   display: flex;
   align-items: center;
+  flex-wrap: wrap;
```
monkeypox8, line 471: Currently defaulting `SageButtonGroup` to wrap content in order to fix cases of horizontal overflow on mobile. The base style disables this at the earliest (`sm-min`) breakpoint, but I'm hoping to get feedback on whether this is what we want. Would it be preferred to keep the default `nowrap` as before, with the option to override? In certain cases, such as a modal with lengthy button text, wrapping may not be desired.

docs/lib/sage_rails/app/sage_components/sage_button.rb
```diff
@@ -4,12 +4,12 @@ class SageButton < SageComponent
     attributes: [:optional, NilClass, Hash],
     custom_content_class: [:optional, NilClass, String],
     disabled: [:optional, NilClass, TrueClass],
-    spinner_on_submit: [:optional, NilClass, String],
     full_width: [:optional, NilClass, TrueClass],
     icon: [:optional, { name: String, style: Set.new(["left", "right", "only"]) }],
     no_shadow: [:optional, NilClass, TrueClass],
     raised: [:optional, NilClass, TrueClass],
     small: [:optional, NilClass, TrueClass],
+    spinner_on_submit: [:optional, NilClass, String],
```
monkeypox8, line 12: Moved to correct alpha order

packages/sage-assets/lib/stylesheets/components/_button.scss
joedakroub: I agree that in the scenario you describe with a modal with lengthy button text, having the option to override the wrapping could be desirable. Not sure how often this would happen however.

docs/app/views/examples/components/button/_preview.html.erb
```diff
@@ -227,14 +227,36 @@ demo_configs = [
   } %>
 <% end %>
 
+<%= sage_component SageCardBlock, {} do %>
+  <%= sage_component SageLabel, {
+    color: "draft",
+    value: "Button group: Allow wrap",
+  } %>
+<% end %>
+<p>When enabled, button groups are no longer restricted to a single row, allowing contents to wrap to multiple lines. <strong>Note: all button groups regardless of this attribute will wrap at breakpoints below `sm-min` to prevent horizontal scrolling on small screen devices.</strong></p>
```
philschanely, line 236: 💡  remember to make the most of our handy `<%= md(..., use_sage_type: true) %>` helper :)

packages/sage-assets/lib/stylesheets/components/_button.scss
philschanely: Likewise I see the use case, but I wonder if we should flip it and have `nowrap` if most of the time we'd want the wrapping to occur?

packages/sage-assets/lib/stylesheets/components/_button.scss
```diff
@@ -483,6 +468,23 @@ $-btn-loading-min-height: rem(36px);
...
+
+  @media screen and (min-width: sage-breakpoint(sm-min)) {
+    flex-wrap: nowrap;
+  }
+}
+
+@each $-key, $-value in $sage-spacings {
+  .sage-btn-group--gap-#{$-key} {
+    gap: sage-spacing($-key);
```
philschanely, line 480: 😍

packages/sage-assets/lib/stylesheets/components/_button.scss
monkeypox8: @philschanely funny you should say that! I'd originally set up this PR with that expectation; the default for `SageButtonGroup` was to wrap with a param provided for `noWrap`. However, during testing I realized that doing so could potentially cause unexpected results in layout positioning for a significant number of views (97 uses in the application at last count).

In theory, this shouldn't affect anything other than a need for additional vertical height. But due to [a recent issue](https://github.com/Kajabi/kajabi-products/pull/20280) where buttons outside of the viewport were causing specs to fail, I wonder if other specs may run into the same problem.

joedakroub (approved):
Looks great Kevin. Nice work! 👍🏼

pixelflips (approved):
LGTM! Working as expected and has a really nice setup. Great work @monkeypox8 👍🏼

PR #713, opened 2021-08-13
Patch dropdown trigger leading

philschanely:
## Description

This PR makes a quick patch to dropdown trigger styles that ensure consistent size and alignment for the button trigger.

## Testing in `sage-lib`

See no changes in Components > Dropdown

## Testing in `kajabi-products`

1. (LOW) Fix to line height in dropdown button trigger fixes tiny alignment issue.

## Related

#489

PR #714, opened 2021-08-13
JBH Alert Updates

### LINKED ISSUE #498: Update Alert
## Description

Update component to match latest specs in Sage 3.0. The following were noted during initial audit:

- Fix icon alignment
- Check color combinations

**Dependencies**:
- [Color updates](https://github.com/Kajabi/sage-lib/issues/490)

**Links**
- [Figma specs](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/WIP-Sage-3-%E2%80%94-Admin-Components?node-id=3671%3A16691)
- [Existing component/s](https://sage-design-system.kajabi.com/pages/component/alert)

FuturaExtraBold:
## Screenshots

|  Before  |  After  |
|--------|--------|
|||

## Related
Closes #498


packages/sage-assets/lib/stylesheets/components/_alert.scss
```diff
@@ -8,12 +8,14 @@
 $-alert-colors: (
   info: primary,
   published: sage,
+  success: sage,
```
FuturaExtraBold, line 11: The React component uses SUCCESS, the Rails component uses PUBLISHED.

packages/sage-assets/lib/stylesheets/components/_alert.scss
```diff
@@ -8,12 +8,14 @@
...
   warning: yellow,
   danger: red,
 );
 $-alert-icon-colors: (
   info: primary,
   published: sage,
+  success: sage,
```
FuturaExtraBold, line 18: The React component uses SUCCESS, the Rails component uses PUBLISHED.

packages/sage-assets/lib/stylesheets/components/_alert.scss
```diff
@@ -48,6 +50,10 @@ $-alert-icon-colors: (
       text-decoration: underline;
     }
 
+    .sage-alert__icon::before {
+      line-height: sage-line-height(medium);
```
FuturaExtraBold, line 54: This lines up the icon with the alert title, as they now have the same line-height.

packages/sage-assets/lib/stylesheets/components/_alert.scss
```diff
@@ -116,6 +122,7 @@ $-alert-icon-colors: (
     .sage-alert--#{$name} & > a {
       @include sage-focus-outline($outline-offset-inline: 6px, $outline-offset-block: 0);
       @include sage-focus-outline--update-color(sage-color($color, 400));
+      font-weight: sage-font-weight(medium);
```
FuturaExtraBold, line 125: Changes the font-weight from 400 to 500.

packages/sage-assets/lib/stylesheets/components/_alert.scss
philschanely: FWIW I'd vote in favor of consolidating to `success` if there was any effort to trim down in the future.

FuturaExtraBold: Agreed. Unsure of the reach this would have to existing Rails alerts. I would suggest a follow up to audit.

packages/sage-assets/lib/stylesheets/components/_alert.scss
philschanely, line 54: This may not work consistently from a responsive standpoint... was that considered here? May either need a type spec or the second param set to `false`. I [just put through a PR](https://github.com/Kajabi/sage-lib/pull/713) with an updated description for this Sass function (should be on `develop` now).

philschanely (approved):
One small tweak. Also do you think [this ticket](https://github.com/Kajabi/sage-lib/issues/626t is relevant to pull into this effort?

packages/sage-assets/lib/stylesheets/components/_alert.scss
FuturaExtraBold: @philschanely I would keep the Alert button issue as separate for now. We're gonna focus on completing Sage 3.0 this cycle. Just rebased against `develop`. Trying the second param now...

packages/sage-assets/lib/stylesheets/components/_alert.scss
philschanely: I know what you mean... it appears to align. In Inspector though you can see that `medium` here tries to add a `var()` that is undefined. There are no `var()` for the line-heights such as `medium` and `large` but only for the type specs that are responsive. So its actually falling back to a different value in the cascade. 

Screen Shot 2021-08-23 at 3 01 06 PM

If you switch to a type spec such as `body` or use `sage-line-height(medium, false)` you should see a valid value; not sure which one you need for correct size but my hunch is it should match the type spec for the type its next to. 🤔

packages/sage-assets/lib/stylesheets/components/_alert.scss
philschanely: I know what you mean... it appears to align. In Inspector though you can see that `medium` here tries to add a `var()` that is undefined. There are no `var()` for the line-heights such as `medium` and `large` but only for the type specs that are responsive. So its actually falling back to a different value in the cascade. 

Screen Shot 2021-08-23 at 3 01 06 PM

If you switch to a type spec such as `body` or use `sage-line-height(medium, false)` you should see a valid value; not sure which one you need for correct size but my hunch is it should match the type spec for the type its next to. 🤔

FuturaExtraBold: @philschanely I was using the incorrect variable name. However, I'm still having some difficulty with the line-height here. For desktop, I want `sm` (1.5), for mobile, I want `md` (1.75). Using a type spec via `@extend` is throwing off some other properties for the icon. (If this is how you were suggesting I use it). Any suggestions?

`md` on mobile:
image

`sm` on desktop:
image

FuturaExtraBold: Using `line-height: var(--sage-line-height-h6);` works, but I've not seen this used in any other component stylesheet.

PR #717, opened 2021-08-16
[Button] loading state focus ring alignment

QuintonJason:
## Description

- [x] updated the focus ring alignment for agenda items

## Screenshots

|  Before  |  After  |
|--------|--------|
|loadingbuttonagendaitemsbefore|loadingbuttonagendaitems|

## Testing in `sage-lib`

It is unseeable when the component is used in isolation.

1. (**LOW**) Update alignment of the focus ring when an agenda items is saved.
   - [ ] Coaching programs client view -> edit agenda item

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_button.scss
@@ -323,7 +323,7 @@ $-btn-loading-min-height: rem(36px);
   pointer-events: none;
 
   &.sage-btn--icon-only-check::before {
-    content: '';
+    content: "";
   }
 
   svg {
--- a/packages/sage-assets/lib/stylesheets/components/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_button.scss
@@ -320,8 +320,11 @@ $-btn-loading-min-height: rem(36px);
 }
 
 .sage-btn--is-loading {
-  min-height: $-btn-loading-min-height;
   pointer-events: none;
+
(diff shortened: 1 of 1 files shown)


joedakroub (approved):
Looking good! Nice update @QuintonJason. 👍🏼

packages/sage-assets/lib/stylesheets/components/_button.scss
```diff
@@ -320,9 +320,12 @@ $-btn-loading-min-height: rem(36px);
 }
 
 .sage-btn--is-loading {
-  min-height: $-btn-loading-min-height;
   pointer-events: none;
 
+  &.sage-btn--icon-only-check::before {
```
monkeypox8, line 325: As a follow-up, I can see a case where we'd want to extend this pattern to other icons as well. I'd be in favor of using a class or data-attribute so we're not tied to this specific icon name.

monkeypox8 (approved):
👍🏼 LGTM, I wasn't able to see the "loading" animation when editing a coaching agenda item, but I also had several feature flags missing when testing with the bridge. In any case, the "check" icon was quickly removed from display.

QuintonJason:
> 👍🏼 LGTM, I wasn't able to see the "loading" animation when editing a coaching agenda item, but I also had several feature flags missing when testing with the bridge. In any case, the "check" icon was quickly removed from display.

Will update: on the `kajabi-products` side when calling a button, add the followup prop `spinner_on_submit: "my aria label",` that activates it

PR #688, opened 2021-08-06
Jcm main sections

cmcfadzean:
## Description
In order to make the documentation clear for developers this PR removes the Foundations and Content Sections. 
This is the first step to cleaning up the site and adding some new features.

Issue #605, opened 2021-06-29
Remove `attributes` from components that use it


philschanely:
## Description

Once #604 is complete, all Rails components that use `attributes` can be refactored to remove this property.

## Related
Blocked by #604

philschanely:
Moved to Jira https://kajabi.atlassian.net/browse/UXD-68

PR #733, opened 2021-08-20
SageLink updates

joedakroub:
## Description
This PR updates the Sage Link rails implementation:
1. Uses `link_to` under-the-hood to generate links
2. Now has a `component.content` slot for adding custom/nested content
3. General refactoring and cleanup

## Screenshots
|  Before  |  After  |
|--------|--------|
|Before|After|

## Testing in `sage-lib`
1. Visit http://0.0.0.0:4000/pages/component/link and ensure that the screen matches the `After` screenshot
2. Ensure all links are rendering properly in the DOM

1. (**HIGH**) Any use of the `SageLink` Rails component will be affected.
   - [ ] Account Password - /admin/settings/password
   - [ ] Contact Info - /admin/settings/contact_info
   - [ ] Account Users - /admin/settings/users


docs/app/views/examples/components/link/_preview.html.erb
```diff
@@ -27,6 +27,14 @@
       "data-js-tooltip-position": "right",
     },
   } %>
+  <%= sage_component SageLink, {
+    url: "#",
+    external: false,
+    help_link: false,
+    show_label: true,
+  } do %>
+    <%= tag.span "Internal link with #{tag.strong "custom content"}".html_safe, class: "custom-content" %>
```
joedakroub, line 36: Example of providing custom/nested content.

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
```diff
@@ -1,85 +1,38 @@
...
-        </span>
-      ).html_safe
-    else
-      component.label.html_safe
-    end
+css_classes = ["sage-link"]
+html_attributes = {}
+  .merge(component.attributes || {})
+  .merge(component.html_attributes || {})
+label = component.content
```
joedakroub, line 6: CSS classes get stored in an array so that we can easily add classes bases on conditions.

`attributes` and `html_attributes` get merged into one hash that can be added to based on conditions.

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
```diff
@@ -1,85 +1,38 @@
...
+elsif !component.external && component.launch
+  css_classes.append("sage-link--launch")
+elsif component.external && component.launch
+  html_attributes[:rel] = "noopener noreferrer"
+  html_attributes[:target] = "_blank"
+elsif component.external && !component.launch
+  css_classes.append("sage-link--no-launch")
+  html_attributes[:rel] = "noopener noreferrer"
+  html_attributes[:target] = "_blank"
+end
```
joedakroub, line 31: Here we added classes and set properties of `html_attributes` to be used later in the `link_to` call.

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
```diff
@@ -1,85 +1,38 @@
...
+
+html_attributes[:class] = css_classes
+  .push(component.generated_css_classes)
+  .join(" ")
+  .squish
 %>
```
joedakroub, line 37: Add the `generated_css_classes` last, join the array into a space-separated string and trim.

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
```diff
@@ -1,85 +1,38 @@
...
-    rel="noopener noreferrer"
-    <% component.attributes.each do |key, value| %>
-      <%= "#{key}=\"#{value}\"".html_safe %>
-    <% end if component.attributes&.is_a?(Hash) %>
-    <%= component.generated_html_attributes.html_safe %>
-  >
-    <%= label %>
-  </a>
-<% end %>
+<%= link_to label, component.url, html_attributes %>
```
joedakroub, line 38: Call `link_to` with the label, url and attributes.

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
```diff
@@ -1,85 +1,38 @@
...
-    end
+css_classes = ["sage-link"]
+html_attributes = {}
+  .merge(component.attributes || {})
+  .merge(component.html_attributes || {})
+label = component.content
+
+if component.label.present?
+  label = if component.truncate.present?
+    tag.span component.label, class: "t-sage--truncate"
```
joedakroub, line 10: Using the `tag` helper to generate a `span` element.

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
```diff
@@ -1,85 +1,38 @@
...
+  label.html_safe
+end
+
+if component.help_link && component.show_label
+  css_classes.append("sage-link--help")
+elsif component.help_link && !component.show_label
+  css_classes.append("sage-link--help-icon-only")
+  label = tag.span label, class: "visually-hidden"
+elsif !component.external && component.launch
+  css_classes.append("sage-link--launch")
```
pixelflips, line 23: Not sure if this update is actually the cause, but I thought I remembered our launch links had a `launch` icon? I see the styles in the _link.scss file, but it doesn't seem to render. 

I see it no longer appears on the live docs site either. If we no longer need the icon, it might be worth it to remove the styles.

joedakroub: Good catch Phillip! I'll take a look into that. Strange that the live docs are displaying this as well. I know we used to have launch icons here.

anechol:
Very awesome update!!

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
joedakroub: Looks like the launch links need a `span.t-sage--truncate` wrapper. I added logic in to address this. Launch icon is back. 👍🏼

pixelflips (approved):
LGTM! 👍🏼

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
```diff
@@ -1,85 +1,39 @@
...
+elsif !component.external && !component.launch
+elsif !component.external && component.launch
+  css_classes.append("sage-link--launch")
+elsif component.external && component.launch
+  label = tag.span component.label, class: "t-sage--truncate"
+  html_attributes[:rel] = "noopener noreferrer"
+  html_attributes[:target] = "_blank"
+elsif component.external && !component.launch
+  css_classes.append("sage-link--no-launch")
+  html_attributes[:rel] = "noopener noreferrer"
```
monkeypox8, line 30: This brings up something that I've been pondering for a while now after seeing similar use thoroughout Sage: should we continue to apply both of these attributes for "launch" links in components?

If I understand the usage of the referral header correctly, there would seem to be cases where we'd actually want it to be accounted for. Zendesk help links are a good example. But if we do feel that we would like to keep the existing `noreferrer` behavior, we should be able to drop the use of `noopener`, since that keyword is implied with `noreferrer`.

monkeypox8 (commented):
I'm seeing a couple of broken views running the bridge, and judging by their error lists, I suspect it may be related to the `html_attributes` and `attributes` merge in the component. Anyone else able to verify?

<details>
<summary>Products list</summary>

</details>

<details>
<summary>Offer edit</summary>

</details>

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
joedakroub: I am able to replicate this error as well. Looking into it now. I removed `noopener` from both; good point. 👍🏼

joedakroub: I was able to correct the broken views mentioned above. The error was due to the use of the `tag` helper. I switched to using the `content_tag` helper and the issues were fixed.

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
philschanely: It might be nice to take this chance to trim to using `html_attributes` instead of `attributes`. We have a ticket out about this effort more broadly, but IMO its a bug in our system that we allow both on some components. I def appreciate merging them here! Do you know if/how this method handles duplicate entries for a given key?

See #605

philschanely:
Related: #688

philschanely (approved):
Fabulous work here. Excited about the outcome.

joedakroub:
@philschanely The `merge` function will overwrite the props of the previous hash with the next. So if `attributes` contains an `href=123` and `html_attributes` contains `href=456`, it will take the last value.

docs/lib/sage_rails/app/views/sage_components/_sage_link.html.erb
joedakroub: I am definitely open to this. I just checked the app and the usage of `attributes` on SageLink is about 10 instances.

joedakroub:
@philschanely I removed the `attributes` prop and updated the example to use `html_attributes`. Will create a PR in `kajabi-products` to address the `SageLink` instances. 👍🏼

monkeypox8 (approved):
👍🏼 LGTM! The offer edit page (and other views with the shared automations partial) are currently failing on this branch due to the use of `attributes`, but it's working great locally with the related patch + bridge. Nice work, this will be a valuable addition!

PR #736, opened 2021-08-20
docs(status): improivng Status Page

cmcfadzean:
## Description
Updating the site strucutre to better map for developers. 
New Structure: 
Foundations
Components
Patterns
Interactions ( Coming Soon )
Helpers

This also replaces the bottom of footer links with a link to a support page with the satisfaction survey. 
Brings back Tokens and Code Status to better help developers navigate what is and is not available.
## Testing in `sage-lib`
Booting up the lib should show the new structure.


docs/app/views/application/_sidebar.html.erb
```diff
@@ -9,103 +9,106 @@
...
-            text: "Colors",
-            link: pages_design_path(:color),
-            active: current_page?(pages_design_path(:color)),
-          },
+        { link: pages_patterns_path(:panels_cards), text: "PATTERNS", items: [
           {
-            text: "Typography",
-            link: pages_design_path(:typography),
-            active: current_page?(pages_design_path(:typography)),
+            text: "Organization",
```
philschanely, line 73: Is this a duplicate of "Layout" on purpose?

docs/app/views/application/_status_item.html.erb
```diff
@@ -1,8 +1,28 @@
 <tr id="<%= item[:title] %>">
-  <td><%= item[:title].titleize %></td>
-  <td><span class="status-key__icon status-key__icon--<%= item[:scss] %>"><span class="visually-hidden"><%= item[:scss].titleize %></span></span></td>
-  <td><span class="status-key__icon status-key__icon--<%= item[:a11y] %>"><span class="visually-hidden"><%= item[:a11y].titleize %></span></span></td>
-  <td><span class="status-key__icon status-key__icon--<%= item[:docs] %>"><span class="visually-hidden"><%= item[:docs].titleize %></span></span></td>
-  <td><span class="status-key__icon status-key__icon--<%= item[:rails] %>"><span class="visually-hidden"><%= item[:rails].titleize %></span></span></td>
-  <td><span class="status-key__icon status-key__icon--<%= item[:react] %>"><span class="visually-hidden"><%= item[:react].titleize %></span></span></td>
+  <td><%= item[:title] %></td>
```
philschanely, line 2: I remember we hid these for two reasons (at least that I recall) and still believe should be addressed before we reveal them again:

1. We didn't have clear understanding of what each category meant.
2. We didn't have a clear plan for when and how these would be updated or changed.

docs/app/views/pages/token.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= content_for :heading do %>
 <%= md(%(
-# DEPRECATED PAGE: Tokens
+# Tokens
```
philschanely, line 3: I am not in favor of adding this back in in this current form. These tokens are largely irrelevant to developers and the Sassdocs site is a much stronger integrated documentation source for this. What is the thinking here?

philschanely (commented):
Excited to continue streamlining but not so sure about some relics we're resurrecting here. Looking forward to your thoughts. (see comments)

docs/app/views/application/_sidebar.html.erb
philschanely: I see now that Layout is hidden from the nav now... any need for this section to remain here?

docs/app/views/pages/token.html.erb
cmcfadzean: This was a request from I believe @anechol saying it would be nice to have easy access to the system tokens. While I mainly agree that most developers would not interact with this list on a regular basis, I do think that it is pretty common including on system documentation and can be a handy reference for understanding what the core values of the system are even if you don't plan on utilizing them in code.

docs/app/views/application/_status_item.html.erb
cmcfadzean: A common question we get about the system is around the state of components and their ability to be used. I would suggest we absolutely address the ambiguity and make keeping this list updated a key part of our process. Ideally, at a glance, this can give a developer the confidence that a component they are going to be needing is ready to be pulled off the shelf. It also can give us a high-level overview of the state of the system at any given time.

philschanely (approved):
Thanks for working through concerns with me. Looking good! 🎉

monicawheeler (approved):

PR #732, opened 2021-08-19
docs(master to main): update the repo's reference to main branch as main

monicawheeler:
Relates to #729 

Update terminology to reference `main` rather than `master`

PR #730, opened 2021-08-19
Issue Template Update: Suggestion Template

monicawheeler:
## Description

Add a Suggestion Issue template.

Update the readme to include a clear path to adding issues from the README

PR #683, opened 2021-08-05
Ensure expandable card is hidden by default

philschanely:
Closes #682 

## Description

Expandable card flashes content when page loads. This PR tightens how the card is hidden to ensure it stays securely hidden by default until `aria-expanded="true"` is toggled on via JS:

- remove `aria-expanded="false"` filter to aid default appearance
- add `overflow: hidden` to aid hiding when `aria-hidden="true"` is not activated.

## Testing in `sage-lib`

Validate that Components > ExpandableCard still appears hidden by default until actiavated.

## Testing in `kajabi-products`

1. (LOW) Fixes a bug in Expandable Card to ensure it does not flash on page load before hiding. Note: Undo this patch after this version bump: https://github.com/Kajabi/kajabi-products/pull/20205

PR #738, opened 2021-08-20
Next version bump (~v4.13.0)

monkeypox8:
1.  #711 (**MED/HIGH**) Adds option for button groups to wrap to multiple lines. Default behavior on mobile views has been changed to allow wrapping (other breakpoints are unaffected), so existing uses should be checked to ensure content still flows as expected. Examples include:
    - [ ] Website Pages and Website Landing Pages: Edit page view - Publish status
    - [ ] Coupons: Delete coupon confirmation modal
    - [ ] Offers: Offer export modal
    - [ ] Email campaigns: Broadcast send type selection
2. #683 (LOW) Fixes a bug in Expandable Card to ensure it does not flash on page load before hiding. Note: Undo this patch after this version bump: https://github.com/Kajabi/kajabi-products/pull/20205
3. #713 (LOW) Fix to line height in dropdown button trigger fixes tiny alignment issue.
4. #739  (LOW) Styling adjustment to prevent the button from stretching when input group has message/helper text. No impact is expected but a quick sanity check should be performed.
      - [ ] Settings/Account Details
5. #709  (LOW) Adds dynamic select input
6. #732 (LOW) Docs only update; references to `master` have been removed
7. #737 (LOW) Updates progress bar height and add new "success" variant.
8. #741 (LOW) Internal updates to schema for CardRow component. No impact on usages expected.
9. #730 (LOW) Docs only update.


docs/lib/sage_rails/app/sage_components/sage_component.rb
```diff
@@ -31,7 +31,16 @@ def render
     context.render(
       partial: "sage_components/#{template_path}",
       locals: { component: self }
-    ).tap { cleanup_section_vars }
+    ).tap do |html|
```
monkeypox8, line 34: This change was reverted in `main` and will need to be reapplied here prior to the version bump merge.

anechol (approved):
LGTM

FuturaExtraBold (commented):
👍

PR #739, opened 2021-08-21
Input Group Message Bug

### LINKED ISSUE #672: [BUG] input group - presence of a message causes the action button to be stretched
## Detailed description

When a message exists on the form input in an input group, the sibling action button is stretching to the height of both the input and the message instead of just the input

InputGroup

The presence of a message in the input group should not stretch the action button.

Chrome, Safari, Edge

## Steps to reproduce
1. Visit the sage input group page. Ensure that the preview contains a `mesage_text`
2. Notice that the button is stretched

## Screenshots
Screen Shot 2021-07-30 at 3 05 44 PM

pixelflips:
## Description
Closes #672

Adjusts styling to prevent button stretching when input within an input group includes message/helper text.

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2021-08-20 at 4 28 00 PM|Screen Shot 2021-08-20 at 4 28 38 PM|

## Testing in `sage-lib`

- Add `message_text:` attribute to a `SageFormInput` within `docs/app/views/examples/components/input_group/_preview.html.erb`
- Visit http://localhost:4000/pages/component/input_group
- Verify InputGroup button no longer stretches and the `message_text` appears as expected.

1. (**LOW**) Styling adjustment to prevent the button from stretching when input group has message/helper text. No impact is expected but a quick sanity check should be performed.
   - [ ] Settings/Account Details

## Related
#672

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_button.scss
@@ -234,7 +234,7 @@ $-btn-loading-min-height: rem(36px);
   .sage-input-group & {
     position: absolute;
     top: rem(2px);
-    right: calc(0.125rem / 2);
+    right: rem(1px);
     background-color: sage-color(white);
     box-shadow: rem(-1px) 0 0 0 sage-color(grey, 300);
 
--- a/packages/sage-assets/lib/stylesheets/components/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_button.scss
@@ -233,10 +233,8 @@ $-btn-loading-min-height: rem(36px);
 
   .sage-input-group & {
     position: absolute;
-    transform: translateY(-50%);
-    top: 50%;
-    right: calc(#{rem(2px)} / 2);
-    height: calc(100% - #{rem(2px)});
+    top: rem(2px);
+    right: calc(0.125rem / 2);
     background-color: sage-color(white);
     box-shadow: rem(-1px) 0 0 0 sage-color(grey, 300);
 


packages/sage-assets/lib/stylesheets/components/_button.scss
```diff
@@ -233,10 +233,8 @@ $-btn-loading-min-height: rem(36px);
 
   .sage-input-group & {
     position: absolute;
-    transform: translateY(-50%);
-    top: 50%;
-    right: calc(#{rem(2px)} / 2);
-    height: calc(100% - #{rem(2px)});
+    top: rem(2px);
+    right: calc(0.125rem / 2);
```
joedakroub, line 237: Should the `rem` function be used here as well? Not sure on when to use it and when not to.

joedakroub (approved):
Great catch! Looking great Phillip! Left one comment, not a blocker. 👍🏼

packages/sage-assets/lib/stylesheets/components/_button.scss
philschanely: IMO always use it for places where rem/px would be used. Here it may just need to be wrapped with `#{}`:

```
calc(#{rem(2px)} / 2)
```

Although I might ask if/why we need the `calc` here at all? Couldn't we just do `rem(1px)`?

philschanely (approved):

packages/sage-assets/lib/stylesheets/components/_button.scss
pixelflips: @joedakroub nice catch... that was a bad copy/paste from dev tools while trying out adjustments. See @philschanely comment below which totally makes sense and I will use it to update.

pixelflips: @philschanely you're correct, the `calc` is a bit of overkill here. Will update. Thanks for catching!

PR #747, opened 2021-08-25
Icon alignment beside type

### LINKED ISSUE #742: [Improvement] Sage Icon Alignment option for beside particular type specs
## Description

Icons set beside type in the app sometimes appear misaligned. Consider adding a `beside_type` property that allows a type spec to be provided like so:

```
<%= sage_component SageIcon, { icon: "check-circle", beside_type: "body" } %>
```

The result would be to use the given type spec to set a height and/or line-height for the icon that matches that of the responsive type it is paired beside.

philschanely:
## Description

This PR adds a `beside_type` property to `SageIcon` that allows for cleaner alignment when Icon is used in inline/flex, or grid layouts.

## Screenshots

|   |  Screenshot  |
|--------|--------|
| Standard icon size | Screen Shot 2021-08-24 at 12 34 09 PM |
| Larger icon size sample | Screen Shot 2021-08-24 at 12 38 22 PM |

Documentation

Screen Shot 2021-08-25 at 10 30 40 AM

## Testing in `sage-lib`

See new section in Icon documentation

1. (LOW) Adds an alignment option for Icons when there are placed beside text. No impact on existing uses.

## Related

Closes #742


docs/lib/sage_rails/app/sage_components/sage_tokens.rb
```diff
@@ -295,4 +295,20 @@ module SageTokens
   SPACER_SIZES = [:xs, :sm, :md, :lg, :xl, "2xs", "xs", "sm", "md", "lg", "xl", "2xl", "stage", "panel", "card", "stack"]
 
   STATUSES = ["danger", "draft", "info", "locked", "published", "warning"]
+
+  # All type specs. See `type-specs_scss` to keep in sync.
```
monicawheeler, line 299: Does this mean that any changes to the type specs that we need to update this file? Should we create a reference from the type specs back to here? Or does this note cover that scenario?

docs/lib/sage_rails/app/sage_components/sage_icon.rb
```diff
@@ -1,5 +1,6 @@
 class SageIcon < SageComponent
   set_attribute_schema({
+    beside_type: [:optional, NilClass, Set.new(SageTokens::RESPONSIVE_TYPE_SPECS)],
```
monicawheeler, line 3: The naming of "beside" is giving me pause. Could we use something like "adjacent_type"?

philschanely: Can you clarify further? Since we're addressing for inline placement, I'm not sure adjacent makes anything clearer?

docs/lib/sage_rails/app/sage_components/sage_tokens.rb
philschanely: This is an issue we should have a more systematic answer for IMO. Not sure what to do at this point. I do see a typo here though in the file name I can fix. Any other ideas? We could pull this one for now though we do need to keep the `RESPONSIVE_`... one below in some for for this component's validation. That one is less tied to the specs 🤔

philschanely: I suppose what I mean above is that when we have a token that is used or duplicated or implemented somewhere that creates a liability for any changes. I definitely agree that duplication should be minimal if existent at all. However, when it is need such as for pairing like this, I don't think we yet have a better solution than that when something changes we need to audit its uses to gauge the impact of the change. 🤔

docs/lib/sage_rails/app/sage_components/sage_icon.rb
monicawheeler: "adjacent" means "next to or adjoining" and is more clear in its meaning than "beside" because "beside" can also mean "apart from" and that can be confusing. Let's pair up if this still doesn't clarify!

docs/lib/sage_rails/app/sage_components/sage_tokens.rb
monicawheeler: That's fair and I understand the need for requiring some duplication and referencing to other areas. I want to be sure that from both sides of the change, there are comments to indicate where to make the appropriate addition. Also, let's pair if I need to clarify!

PR #764, opened 2021-09-01
[Sage-760]: Updates path issue to fix Container page

anechol:
## Description

Updates path issue to fix Container page

## Screenshots
N/A

## Testing in `sage-lib`

- Navigate to Container component props page
- Check that page renders
- Click "Layout Containers"
- Check that page renders

## Testing in `kajabi-products`

N/A. Sage Docs-related.

## Related

#760

PR #766, opened 2021-09-01
SAGE-734 - Doc site search

### LINKED ISSUE #760: [BUG] Container component does not load in Sage Docs
## Detailed description

Accessing the [Container](https://sage-design-system.kajabi.com/pages/component/container) component in the Sage Docs, does not load and throws an error. Local and in production

## Steps to reproduce

1. Navigate to [Container](https://sage-design-system.kajabi.com/pages/component/container)
2. Observe the error

## Screenshots

Screen Shot 2021-09-01 at 10 10 39 AM

cmcfadzean:
## Description

Adds simple search to the doc site; searches component names.

### Search button in header
Screen Shot 2022-01-07 at 2 28 12 PM

### Search page
Screen Shot 2022-01-07 at 2 28 35 PM

### Search results
Screen Shot 2022-01-07 at 2 28 56 PM

## Testing in `sage-lib`
- [ ] Check to see that the search returns the components you would expect. 

1. (**N/A**) This is a Sage change only. No testing on `kajabi-products`.


monicawheeler (approved):
Looks great!

anechol:
Oh, hey I didn't know you were working on the Container page. I just [merged in a fix](https://github.com/Kajabi/sage-lib/pull/764) for that, but if removing the link makes more sense, then that's cool!

pixelflips (approved):
LGTM! One small thing is that the search isn't available on the homepage which would be awesome since that's the page I usually navigate to first. Would save a click if I could hop right to the search from there as well. Definitely not a show-stopper.

pixelflips:
> LGTM! One small thing is that the search isn't available on the homepage...

Doh! Just noticed it is on the homepage, just not in the header where I first went looking. Again not a show-stopper.

monicawheeler:
@pixelflips - I agree that we should have a home navbar search option but adding one is not a simple plug-and-go. I am going to leave the search just within the site, not on the home page navigation for now. Thank you for the review!

PR #785, opened 2021-09-07
Modal trigger with `remote_url`

### LINKED ISSUE #677: [Improvement] Modal trigger should specify remote_url
## Description

Currently we allow for modals to be opened on clicking a trigger that has the `data-js-modaltrigger` attribute set to point to the `id` for the corresponding Modal to fire up. We also allow for `Modal` to provide a `data-js-remote-url` that we look for as a Modal is being opened in order to optionally load content from a remote source into the Modal.

It seems it would be helpful for some situations if the trigger could be configured to provide the desired `remote_url` to load on click. This way a given Modal could be used to load and unload a particular set of asynchronous content.

In summary:

- provide an affordance on the trigger that provides a `url` from which to load the modal contents. This should override if the Modal has a remote_url provided.
- provide an affordance—also on the trigger—that can indicate whether or not the modal should remove the contents when closed, even if the Modal does not have this setting on itself.

## Design

N/A

## Additional Information

At present, a similar affordance exists in the application using the `data-remote: true` setting on a trigger and the `href` of the link as the content to load in the modal.

In our app right now, a few places use the `remote_url` for display content in a Modal for a series of items in a list. The result is that a series of empty Modal shells are added, and, when the corresponding item in the list activates the modal, its content loads and then unloads on close. This is not a bad solution since we start with empty containers rather than loading all the contents from the server. however, if the hyperlink could send the remote_url to the Modal to load the content, only one modal would be needed.

See Themes > Theme Store and files such as `app/views/admin/themes/_theme_modal_tile.html.erb`

philschanely:
## Description

This PR adds affordance on Rails `SageModal` to allow a trigger to provide the URL that is used to load asynchronous content. See the updated documentation in the Modal preview page.

Side note: this also adds a new controller for the Docs site for loading asynchronous content useful for other areas of documentation. Thanks to @AndrwM 

## Testing in `sage-lib`

See Components > Modal

1. (LOW) New feature for Modal's JS that allows a remote url to be provided by a trigger through the `data-js-remote-url` attribute. No expected impact on current use of Modal.

## Related

Closes #677


pixelflips (approved):
LGTM! 👍🏼  Nice work on the documentation too!!

PR #786, opened 2021-09-08
Next version bump v4.14.0

FuturaExtraBold:
1. #785  (LOW) New feature for Modal's JS that allows a remote url to be provided by a trigger through the `data-js-remote-url` attribute. No expected impact on current use of Modal.


pixelflips (approved):
LGTM!

PR #816, opened 2021-09-22
Remove `stash` and `stash pop` from `branches:update` script

monicawheeler:
## Description

`stash` and `stash pop` could cause unintended side effects for local dev environments like popping stashed changes. This PR is to remove those commands so that the script only updates branches and begins the rebase.

## Testing in `sage-lib`
N/A - package.json script update 

1. (**LOW**) Nothing for `kajabi-products`; `package.json` update


anechol (approved):
🚀

PR #824, opened 2021-09-23
[Sage-538] Floating Label Alignment Fix

### LINKED ISSUE #538: [BUG] Input "floating" labels out of vertical alignment
## Description
"floating" input labels are appearing ~2-3px below the expected vertical alignment with border. 

|  Expected   |  Actual  |
|--------|--------|
|||

This appears to be caused by the `transform` offset using an incorrect calculation due to the recent typography updates. Removing the custom property assignment for `var(--font-size-body)` displays the correct positioning.

transform

anechol:
## Screenshots

|  Before  |  After  |
|--------|--------|
|||

## Related

Closes #538


monicawheeler (approved):

packages/sage-assets/lib/stylesheets/core/mixins/_sage.scss
```diff
@@ -461,7 +461,7 @@
     transform: translateY($textarea-label-padding - $textarea-padding);
   }
   @else {
-    transform: translateY(-($input-text-height + $input-padding * 2)); // IE doesn't support calc in transforms so we provide a fallback
+    transform: translateY(-50%) translateY(-($input-text-height + $input-padding)); // IE doesn't support calc in transforms so we provide a fallback
```
monkeypox8, line 464: Since we no longer support IE, we should be safe to remove the fallback entirely

monkeypox8 (approved):
👍🏼 LGTM, love the simple solution for the negative custom prop!

packages/sage-assets/lib/stylesheets/core/mixins/_sage.scss
anechol: That's what I thought, but I wasn't entirely sure. I'll just remove the fallback 👌🏽

PR #575, opened 2021-06-14
[Sage-493] Sage 3 - Button Updates

monicawheeler:
Closes #493 

_FYI: this work is going in as a separate version bump (note the merging branch is `main`)_

## Description

Updates styling to Sage 3.0 specs:

* Adds a variation that styles the `secondary` button like the designed tertiary button ([for later consideration](https://github.com/Kajabi/sage-lib/issues/844))
* Aligns focus rings, `color` and `background-color` to designs
* Subtle buttons: Remove small variation in documentation example in favor of disabled
* Add clearer headings for each section of the button types
* Overall documentation improvements

### Buttons

|  Before  |  After  |
|--------|--------|
| Screen Shot 2021-09-28 at 9 06 11 AM | Screen Shot 2021-09-30 at 10 58 38 AM  |
---

### Additional options
|  Before  |  After  |
|--------|--------|
| Screen Shot 2021-09-28 at 11 06 37 AM | Screen Shot 2021-09-28 at 11 06 47 AM |

## Testing in `sage-lib`
1. Navigate to http://localhost:4000/pages/component/button
2. Observe new updates and verify in alignment with specs

## Testing in `kajabi-products`
1. (**BREAKING**) Button updates to the usage of `secondary` can potentially alter the UI throughout the app
   - [ ] Buttons are everywhere

PR #849, opened 2021-10-01
[Sage 493] Quick Fix for subtle button padding 

monicawheeler:
_Note: merging directly into `main` to patch up the 4.18.0 bump_

## Description

With the Sage 3.0 button conversion, `sage-btn--subtle` padding that was introduced for the `hover` effect, which adjusted numerous subtle buttons in the UI to the left by 8px. This task:

* Removes the added `padding` from the [button work](https://github.com/Kajabi/sage-lib/pull/575/files#diff-8b8601a12655a3ec410c2968329becfdd300a5f88fdf21883467af2e7c3d8767R415) 
* Adjusts the `::after` outline mixin in order to hack into the `::after` to accomplish `hover` as designed

### Issue being corrected
some of the many alignment issues

### AFTER
Screen Shot 2021-10-01 at 10 38 14 AM

## Testing in `sage-lib`
1. Navigate to http://localhost:4000/pages/component/button#element-button
2. Verify the "Subtle Buttons" are no longer padded left and right and the hover/focus states are correct

A quick fix which goes into version bump 4.18.0; check for visual regressions in all subtle buttons

### DIFF
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -6,9 +6,9 @@ ruby "2.4.9"
 
 gem 'rails', '5.2.6'
 
-gem 'sass-rails', '~> 5.0'
+gem 'sass-rails', '~> 5.0', '>= 5.0.7'
 gem 'uglifier', '>= 1.3.0'
-gem 'webpacker', '~> 4.x'
+gem 'webpacker', '~> 4.3', '>= 4.3.0'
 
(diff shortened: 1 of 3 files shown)


anechol (approved):
🚀

FuturaExtraBold (approved):
Nice catch!

PR #850, opened 2021-10-01
[SAGE-791] Copy Button Submit Bug

### LINKED ISSUE #791: [BUG] SageCopyButton unintentionally submits forms
## Detailed description
`SageCopyButton` does not currently specify a `type` attribute. This results in the [default for `<button>` being applied](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type), meaning that when used within a form, `SageCopyButton` will submit the form.

Suggested workaround for now is to apply `type: "button"` as an `html_attribute`, but most developers are not likely to be aware of the element's default `type` setting. Documentation around use of the `html_attribute` attribute is also missing, making it even more difficult to know about available options.

Recommendation: apply `type="button"` as the default. We may need to discuss whether we provide an option to override, but it seems highly unlikely that we'd want to submit a form via a copy action.

## Screenshots
<details>
<summary>Copy button submitting form (note the "Saving…" state of the primary CTA)</summary>

</details>

## Related 
- https://github.com/Kajabi/kajabi-products/pull/20626

anechol:
## Description

Fixes a bug where the Copy Button component submits a form when clicked.

## Screenshots

No screenshots. Added attribute `type="button"` to `<button>` tag in Rails component.

## Testing in `sage-lib`
- Check that Rails component now has `type` attribute.

(**LOW**) Fixes a bug where the Copy Button component submits a form when clicked.

## Related

Closes #791 
Related to [COMM-1108](https://github.com/Kajabi/kajabi-products/pull/20626)

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_copy_button.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_copy_button.html.erb
@@ -1,6 +1,7 @@
 <button
   class="sage-copy-btn <%= "sage-copy-btn--fill-container" if component.fill_container %> <%= component.generated_css_classes %>"
   data-js-copy-button="<%= component.value.html_safe %>"
+  type="button"
   <%= component.generated_html_attributes.html_safe %>
 >
   <%= sage_component SageCopyText, { value: component.value, semibold: component.semibold.present? && component.semibold } %>


monkeypox8 (approved):
👍🏼 🔥 LGTM, thanks for getting this one in!

pixelflips (approved):
LGTM! 👍🏼

PR #851, opened 2021-10-01
[Sage-531] Modal Body Scrolling

### LINKED ISSUE #531: Look into scrolling inside modal body
## Description

Per design request, look into an approach to setting a max heigh on modal body to allow scrolling such that a modal does not go below the fold.

anechol:
## Description

Updates modal styling to add a max-height based on the viewport and allow scrolling within.

## Screenshots

https://user-images.githubusercontent.com/14791307/135695862-19c16905-3719-45cd-b5d9-355fe09cc219.mov

## Testing in `sage-lib`

WIP

WIP

## Related

Closes #531


monicawheeler (approved):
🙌 LGTM

monkeypox8 (commented):
Looking great! We'll also have to lock down the horizontal axis, since the `grid` and `flex` containers will cause an overflow:

overflow-x

packages/sage-assets/lib/stylesheets/components/_modal.scss
```diff
@@ -54,8 +54,10 @@ $-modal-header-image-size: rem(28px);
 .sage-modal__container {
   visibility: hidden;
   z-index: sage-z-index(modal);
+  overflow-y: scroll;
```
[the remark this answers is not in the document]
anechol: Ahh good catch. Changing that now.

packages/sage-assets/lib/stylesheets/components/_modal.scss
```diff
@@ -54,8 +54,11 @@ $-modal-header-image-size: rem(28px);
 .sage-modal__container {
   visibility: hidden;
   z-index: sage-z-index(modal);
+  overflow: hidden;
+  overflow-y: scroll;
   width: calc(100vw - #{sage-spacing(md)});
   max-width: sage-container(md);
+  max-height: calc(100vh - #{sage-spacing(xl)});
```
monkeypox8, line 61: With the breakpoints below, we add `6vh` and `8vh` margins respectively, forcing the modal below the viewport once the overflow applies. Consider applying those two values (possibly as component variables) to the `max-height` calculation. So something like:

```
  max-height: calc(100vh - (#{sage-spacing(xl) + #{$-modal-margin-lg}));
```

monkeypox8 (approved):
👍🏼 LGTM! Noted one additional item to consider; not a blocker. Nice work!

packages/sage-assets/lib/stylesheets/components/_modal.scss
monkeypox8: I suppose the variable could also  include `sage-spacing(xl)` to simplify the `calc`

anechol: Working through this a bit, I've added these as variables, but I'm just going to use the modal margin in the calc as that's enough spacing to not look awkward compared to the parent container's padding and won't allow for page scrolling that obscures the modal itself on smaller mobile sizes.

packages/sage-assets/lib/stylesheets/components/_modal.scss
```diff
@@ -7,15 +7,17 @@
 
 $-modal-padding-x: sage-spacing(md);
 $-modal-padding-y: sage-spacing(md);
+$-modal-margin-lg: 6vh;
```
philschanely, line 10: Any particular reason these are in `vh` rather than a spacer value of some sort? Were there any specs that clarify how these were determined?

philschanely (approved):

packages/sage-assets/lib/stylesheets/components/_modal.scss
philschanely: @anechol I now see this is something @monkeypox8 asked about. I'll delete resolve this. Sorry!

PR #847, opened 2021-09-30
fix(actions workflow): update lerna publish step in release deploy workflow

monicawheeler:
Resolves #822 

## Description

Versioning has not been happening since converting to GitHub Actions. This issue aims to resolve that by adjusting permissions on the `yarn version` step in the CI. 

## Screenshots
n/a

## Testing in `sage-lib`
n/a

## Testing in `kajabi-products`
1. (**LOW**) Sage related GitHub Actions change only; no `kajabi-products` testing necessary

PR #852, opened 2021-10-04
fix(lerna): revert lerna publish changes

monicawheeler:
`Lerna Publish` step is failing in the [workflow](https://github.com/Kajabi/sage-lib/runs/3792428391?check_suite_focus=true)

This PR reverts `Lerna Publish` change from #847 

This is a direct to `main` update


monicawheeler:
FYI - @voodooGQ

anechol (approved):

PR #853, opened 2021-10-04
[Sage-848] Avatar Alignment Bug

### LINKED ISSUE #848: [BUG] Avatar alignment needs centered
## Detailed description

The button custom content does not align in the center. An `align-items: center` to the `sage-btn__custom-content` class resolves this. 

## Additional notes
Since button text should not wrap, alignment vertically center makes sense for this custom content.

## Steps to reproduce
1. Navigate to http://localhost:4100/?path=/story/sage-button--custom-content
2. View the alignment of the button custom content is not centered

## Screenshots
avatar-alignment

anechol:
## Description

Adds flex styling to align custom button content.

## Screenshots

|  Before  |  After  |
|--------|--------|
|||
Screen Shot 2021-10-04 at 12 03 51 PM|Screen Shot 2021-10-04 at 12 03 16 PM

## Testing in `sage-lib`

- Navigate to [Button](http://localhost:4100/?path=/story/sage-button--custom-content)
- Check that the alignment is correct

(**LOW**) Adds flex styling to align custom button content.

## Related

Closes #848

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_button.scss
@@ -519,11 +519,6 @@ $-btn-loading-min-height: rem(36px);
   }
 }
 
-.sage-btn__custom-content {
-  display: flex;
-  align-items: center;
-}
-
 // Button groups allow for several buttons together to be spaced appropriately
 .sage-btn-group {
   display: flex;
--- a/packages/sage-assets/lib/stylesheets/components/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_button.scss
@@ -519,6 +519,11 @@ $-btn-loading-min-height: rem(36px);
   }
 }
 
+.sage-btn__custom-content {
+  display: flex;
+  align-items: center;
+}
+
 // Button groups allow for several buttons together to be spaced appropriately
 .sage-btn-group {
   display: flex;
--- a/packages/sage-react/lib/Button/Button.story.jsx
+++ b/packages/sage-react/lib/Button/Button.story.jsx
@@ -47,6 +47,7 @@ CustomContent.decorators = [
         <style>
           {`.demo-custom-class {
             display: flex;
+            align-items: center;
             gap: 8px;
             padding: 8px 0;
           }`}


FuturaExtraBold (approved):
🔥

monicawheeler (approved):

anechol:
Thanks for the callout @philschanely. I didn't find many instances of custom content being used, but thinking about this further, the fact that it's custom content means that theoretically _anything_ could be added. I'm wondering then if it's worth moving forward with this as there's an option to just use a custom class with custom styles if this slot is used. Thoughts?

monicawheeler:
Aligning the button contents to be centered, even if it's custom and could contain any component, seems like a good default. The reason being that since button text ideally should be a single line and not wrap.

anechol:
I've removed the default button styling as I've realized that's a mistake and added the alignment fix to the custom class in Storybook.

FuturaExtraBold (approved):
🔥

anechol:
Hey @FuturaExtraBold do you have any further thoughts on the above comments? 🤔

philschanely:
@anechol and cc: @monicawheeler I hear you on the reasoning for the new setting and I agree its a good default.

My hope is that if we are adding or modifying styling we ensure firsthand that the changes do not adversely affect any existing uses. If they do, we should evaluate the need and communicate the change in advance to the appropriate owners or even put forward the fix ourselves with the version bump (this has been our active practice to date). It is my strong opinion that we will decrease the buy-in and trust of the system if we create problems either without being aware of them or knowingly and without communicating or offering solutions and leave it to others to figure out and fix.

philschanely:
@anechol I see you've revised the PR. That's fine. I just want to circle back to clarify that I don't disagree with the concept of the original PR—it could very well be a helpful default on this feature. I just mean to double-check that we're also investigating changes like this to ensure we're either surely adding net positive or also anticipating and proactively addressing any issues they create. Thanks for your work on this!! 🙇‍♂️

anechol:
Thanks for the input @philschanely! Given the original issue, I think it's fine to just fix the present issue in Storybook here and circle back with another issue later if we truly want to set some default for custom content. This minor change won't touch `kp`.

PR #875, opened 2021-10-08
[Sage-802] Expandable Card Alignment Prop

### LINKED ISSUE #802: [Improvement] Component: Expandable Card - Add icon position
## Description

Provide an option for expandable cards to position the icon to the right of the trigger. This is needed within Coaching and can potentially be valuable in other areas of the app.

## Design
[Coaching designs](https://www.figma.com/file/zTHwmRx2GpM1Omq2HYbfdb/Coaching-%2F-Create-%26-Manage?node-id=5924%3A48961)

## Additional Information

[Expandable card component on Sage](https://sage-design-system.kajabi.com/pages/component/expandable_card)

### Proposed addition

| Existing Implementation | Additional Option |
| ----------- | ----------- |
| Screen Shot 2021-09-17 at 2 23 54 PM | Screen Shot 2021-09-17 at 2 23 04 PM|

anechol:
## Description

Adds a new prop to the Expandable Card component to allow arrow alignment to the right.

## Screenshots

Screen Shot 2021-10-11 at 2 17 54 PM

### Rails
- View [Rails component](http://localhost:4000/pages/component/expandable_card)
- Check that new component props works as expected.

### React
- View [Storybook](http://localhost:4100/?path=/docs/sage-expandablecard--default)
- Check that new component props works as expected.

(**LOW**) New prop added for Expandable Card. Does not affect existing functionality.

## Related

Closes #802

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_expandable_card.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_expandable_card.scss
@@ -70,6 +70,7 @@ $-expandable-card-padding-xs: sage-spacing(xs);
       transform: translate3d(-50%, -50%, 0);
     }
   }
+
   .sage-btn__truncate-text {
     order: -1;
   }
--- a/packages/sage-assets/lib/stylesheets/components/_expandable_card.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_expandable_card.scss
@@ -59,3 +59,14 @@ $-expandable-card-padding-xs: sage-spacing(xs);
     transform: rotateZ(90deg);
   }
 }
+
+.sage-expandable-card--align-right {
+  .sage-btn {
+    display: flex;
+    justify-content: space-between;
+    width: 100%;
+  }
+  .sage-btn__truncate-text {
+    order: -1;
+  }
+}
--- a/packages/sage-assets/lib/stylesheets/components/_expandable_card.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_expandable_card.scss
@@ -59,19 +59,4 @@ $-expandable-card-padding-xs: sage-spacing(xs);
     transform: rotateZ(90deg);
   }
 }
-
-.sage-expandable-card--align-right {
-  .sage-btn {
-    display: flex;
-    justify-content: space-between;
-    width: 100%;
-
-    &::after {
-      transform: translate3d(-50%, -50%, 0);
-    }
-  }
-
-  .sage-btn__truncate-text {
-    order: -1;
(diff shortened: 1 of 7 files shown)


monicawheeler (changes requested):
Looks great! One additional item to include: the documentation for `align_right` within `docs/app/views/examples/components/expandable_card/_props.html.erb`

packages/sage-assets/lib/stylesheets/components/_expandable_card.scss
```diff
@@ -59,3 +59,19 @@ $-expandable-card-padding-xs: sage-spacing(xs);
     transform: rotateZ(90deg);
   }
 }
+
+.sage-expandable-card--align-right {
+  .sage-btn {
```
philschanely, line 64: FWIW our practice so far has been to place styles like this in the selected element's file: button in this case, and to therefor do this kind of selection:

```scss
.sage-btn {
  // other existing button styles here
  
.sage-expandable-card--align-right & {
    // styles here
  }
}
```

The hope is to keep all styles affecting a given component within that component, and thus have a standard when there's places like this that could go either way. You could still leave a comment in this file regarding that aspect if that seems good to you.

packages/sage-assets/lib/stylesheets/components/_expandable_card.scss
```diff
@@ -59,3 +59,19 @@ $-expandable-card-padding-xs: sage-spacing(xs);
...
+    display: flex;
+    justify-content: space-between;
+    width: 100%;
+
+    &::after {
+      transform: translate3d(-50%, -50%, 0);
+    }
+  }
+
+  .sage-btn__truncate-text {
```
philschanely, line 74: Same deal here, in that this could move over to be alongside other styles for `sage-btn__truncate-text` in `button.scss`.

philschanely (commented):
Looking good here. A few comments regarding nested SCSS practice. Also, I agree with your technique for the styling but wonder if there's a clearer label for the property. `align_right` leaves it a little unclear to me what will be aligned or affected. So something like `button_arrow_align_right` makes it clearer. It also relates to the button filling the space available so i could see `button_fill_width` as another idea to consider. What do you think?

Lastly, a quick audit of this component's use might also indicate whether or not this could switch to being the default setting.

anechol:
@monkeypox8 Thanks for catching that. I've removed the margin.

A more uniform naming convention across Sage would be beneficial in my opinion, especially if it sounds more like well-known CSS values. Could definitely be something to consider.

PR #893, opened 2021-10-15
Switch: Sync up features between Rails and React

philschanely:
## Description

This PR syncs up configurations between Rails and React for Switches. No existing properties were changed but future work should further align property names.

React lacked the following:

- `disabled`
- `hasBorder` for adding a padded border around the switch
- `hideText` for hiding the label text
- `required`
- `togglePositiion` for moving the switch to the right of the label

Rails lacked the following:

- `in_list` for rendering `li` as container tag rather than `div` (default)

No visual changes.

## Testing in `sage-lib`

- See Docs > Components > Switch for new preview options
- See Storybook > Switch to test new features

1. (LOW) Switch in React and Rails now have full set of similar features. No impact on existing components in use.


monkeypox8 (approved):
👍🏼 LGTM! Tested briefly with the bridge, no concerns here.

PR #894, opened 2021-10-15
Chart: Stabilizes React component 

philschanely:
## Description

Prior to this PR the Chart component was developed using Recharts. However, this library presented conflicts with the version used in `kajabi-products` that lead to the js failing to compile. It was commented out here for further investigation.

This PR adds a temporary fix that removes the library from direct use in the component allowing instead for it to be passed in when used in the desired context. 

Future consideration should investigate syncing Recharts versions with `kp` and using peer dependencies. However, this may also be helpful as a start towards allowing multiple charting frameworks to be used depending on a product team's preferences. In that case different configurations would need to be integrated and forked from the existing Recharts assumptions.

## Testing in `sage-lib`

See Storybook > Charts for demonstration of these features.

1. (LOW) Recharts-based Chart component is re-integrated. No expected changes as this is a new feature. However, earlier attempts to implement this component caused JS to fail to compile in app, so load a page such as People and confirm all loads and behaves as expected.

### DIFF
--- a/packages/sage-react/lib/Chart/Chart.jsx
+++ b/packages/sage-react/lib/Chart/Chart.jsx
@@ -8,6 +8,7 @@ import { SageTokens } from '../configs';
 
 export const Chart = ({
   configs,
+  framework,
   data,
   noDataText,
   loading,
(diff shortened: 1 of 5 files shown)


monkeypox8 (approved):
👍🏼 LGTM. One minor SVG-related error appears while running the bridge (screen from People list view), but I'm not sure if it's a result of anything from this branch. Page renders and functions as expected.

PR #899, opened 2021-10-15
[Sage-890] Copy Button Variation Documentation

### LINKED ISSUE #890: [Enhancement] Button: Document "copy" functionality and example
Common use of standard button is to use it to copy a static value; this needs to be shown on Button page.

See Copy Button component, which is currently set up more for form field contexts. It nonetheless uses the `data-js-copybutton` property that can be put on other components.

- [ ] Add a preview/documentation example of using this attribute on a standard button such as to achieve the contextual use shown in the mock below.

Screen Shot 2021-10-14 at 10 42 10 AM

Note that in this mock it is acceptable and possible for the Button to be just the icon-only or also encompass the email value. Regardless, the `data-js-...` attribute can be implemented. Both variations may be useful to demonstrate.

anechol:
## Description

Adds an example in the Sage docs for a Copy Button variation.

## Screenshots

WIP

## Testing in `sage-lib`

WIP

(**LOW**) Adds an example in the Sage docs for a Copy Button variation. No effect on Kajabi Products.

## Related

Closes #890


docs/app/views/examples/components/button/_preview.html.erb
```diff
@@ -272,6 +272,23 @@ demo_configs_subtle = [
   } %>
 </div>
 
+<h3>Copy Button</h3>
```
pixelflips, line 275: We also have a copy button documented in http://localhost:4000/pages/component/copy_text. Any thoughts on how we may be able to reduce the confusion?

PR #901, opened 2021-10-18
fix: add `content` as a possible default component property

philschanely:
## Description

Adds `content` as a generic attribute available for components.

## Testing in `sage-lib`

N/A fallback attribute

1. (**LOW/MEDIUM/HIGH/BREAKING**) Adds a new default attribute for all sage components that also patches a recent issue where `content` was not outputting.


prsimp (approved):
Looks good to me 👍

PR #905, opened 2021-10-19
MOCK: Plan tier panel

joedakroub:
## Description
This PR is a POC mock for the upcoming Plan Tier Panel.

## Screenshots
|[Design Spec]()|Possible now (See sandbox page)|
|-----|----|
|image|Possible|

## Items to discuss:
- [ ] DESIGN: Spacing is using the `SagePanelStack` which is different from the Figma design
- [x] DESIGN: Spacing between the check icon and feature is different due to the default gap in the `SagePanelRow`
- [x] DESIGN: Checkmark icon style does not match what we can achieve with `SageIcon`
- [x] GENERAL: `SageButton` does not allow for centered text (not sure if this is a regression or if it never centered text) so the `Select` button text is not centered
- [x] GENERAL: There are no width constraints set on the panel. It is assumed that it will be placed inside a Container.

### DIFF
--- a/docs/app/views/pages/sandbox.html.erb
+++ b/docs/app/views/pages/sandbox.html.erb
@@ -1,3 +1,154 @@
+<%
+plans = [
+  {
+    name: "Beginner",
+    price: 69,
+    duration: "mo",
+    billing: "Billed annually",
+    description: "Try out the core features",
+    current: true,
+    features: [
+      "1 Products",
+      "250 Members",
+      "1k marketing emails",
+      "10K Contacts",
+      "1 Pipeline",
+      "50 landing pages",
+      "1 Admin user",
+      "1 Site",
+      "Kajabi branding",
+      "0% Transaction fee",
+      "Chat bot support",
+      "Live webinars",
+      "Standard analytics",
+    ]
+  },
+
+  {
+    name: "Basic",
+    price: 149,
+    duration: "mo",
+    billing: "Billed annually",
+    description: "All the basics for starting your new business",
+    current: false,
+    features: [
+      "3 Products",
+      "1k Members",
+      "Unlimited marketing emails",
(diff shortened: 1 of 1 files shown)


docs/app/views/pages/sandbox.html.erb
```diff
@@ -1,3 +1,154 @@
...
+        </p>
+      <% end %>
+      <%= sage_component SagePanelBlock, { type_block: true } do %>
+        <%= sage_component SageButton, {
+          disabled: plan[:current],
+          full_width: true,
+          style: "primary",
+          value: plan[:current] ? "Current plan" : "Select",
+        } %>
+        <hr>
```
joedakroub, line 136: I think the `hr` works here to thematically break up the content, but am open to another option.

docs/app/views/pages/sandbox.html.erb
```diff
@@ -1,3 +1,156 @@
...
+            style: "primary",
+            value: plan[:current] ? "Current plan" : "Select",
+          } %>
+          <hr>
+          <%= sage_component SagePanelStack, {} do %>
+            <h3 class="<%= "#{SageClassnames::TYPE::HEADING_6} #{SageClassnames::SPACERS::XS_BOTTOM}" %>">Features</h3>
+
+            <% plan[:features].each do | feature | %>
+              <%= sage_component SagePanelRow, { grid_template: "et" } do %>
+                <%= sage_component SageIcon, { icon: "check-circle", color: "sage", size: "lg" } %>
```
philschanely, line 143: I wonder if the background options (esp the circular ones) would work here? May need a spacer/padding modifier though.

monicawheeler:
Closing. Per Joe, these are either integrated or being handled elsewhere 👍

PR #967, opened 2021-10-27
[SAGE-886] Button update to spec

### LINKED ISSUE #886: [Improvement] Button: Icon size and spacing adjustments
- [ ] Buttons need further sync with latest Figma spec
  - [ ] check size in mockups and adjust padding
  - [ ] padding may also size down in icon-only variations to ensure these appear square with border and/or hover effects
  - [ ] double-check icon sizes in normal and subtle buttons. Ensure any fix works consistently with responsive type size changes

[Figma spec](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/?node-id=6999%3A21637)

Icon-only:

Screen Shot 2021-10-14 at 10 16 42 AM

QuintonJason:
## Description

- [x] update sage 3 button to spec, namely padding between icon only subtle vs non-subtle variations
- [x] resolve hover interaction outline flash

#### Standard button
|  Before  |  After  |
|--------|--------|
|standardbefore|standardafter|

#### Subtle Button
|  Before  |  After  |
|--------|--------|
|subtlebefore|subtleafter|

## Testing in `sage-lib`

Visit Rails and React button pages and interact with the `icon-only` variants and verify they match the spec

1. (**MEDIUM**) Updates spacing in standard and subtle icon only buttons throughout.

## Related

Closes #886

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_button.scss
@@ -439,8 +439,6 @@ $-btn-loading-min-height: rem(36px);
       color: map-get($-styles, default);
       background-color: transparent;
 
-      // @include sage-focus-outline--update-color(map-get($-styles, focus-outline));
-
       // TODO: refactor the subtle button hover state, after Sage 3.0 conversion
       &::after {
         border-color: transparent;
--- a/packages/sage-assets/lib/stylesheets/components/_button.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_button.scss
@@ -484,14 +484,14 @@ $-btn-loading-min-height: rem(36px);
         }
 
         &:hover::after {
-          opacity: 1;
           border-color: transparent;
+          opacity: 1;
         }
 
         &:focus::after {
-          border-color: map-get($-styles, focus-outline);
           width: $-btn-icon-only-focus-size;
           height: $-btn-icon-only-focus-size;
+          border-color: map-get($-styles, focus-outline);
 
         }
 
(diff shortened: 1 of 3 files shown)


philschanely (approved):
LGTM

packages/sage-assets/lib/stylesheets/components/_button.scss
```diff
@@ -437,8 +439,6 @@ $-btn-loading-min-height: rem(36px);
       color: map-get($-styles, default);
       background-color: transparent;
 
-      @include sage-focus-outline--update-color(map-get($-styles, focus-outline));
-
       // TODO: refactor the subtle button hover state, after Sage 3.0 conversion
```
monkeypox8, line 442: Are the changes here sufficient to remove this comment?

packages/sage-assets/lib/stylesheets/components/_button.scss
```diff
@@ -448,10 +448,9 @@ $-btn-loading-min-height: rem(36px);
         color: map-get($-styles, hover);
 
         &::after {
```
monkeypox8, line 450: Non-blocking item: the focus state for `subtle` buttons feels a little choppy. I'm wondering if we can cut the `transform` timing to say, `0.1s`. As it currently stands, my guess is the `0.2s` duration is too long for the relative transform, drawing out the transition's appearance. I tried to add a screen capture, but the limited frame rate doesn't seem to show the effect the same way.

monkeypox8 (approved):
👍🏼 LGTM, nicely done!

packages/sage-assets/lib/stylesheets/components/_button.scss
QuintonJason: #1011 will handle the followup

packages/sage-assets/lib/stylesheets/components/_button.scss
QuintonJason: will add to the followup once verified

Issue #935, opened 2021-10-22
[Enhancement] PanelRow/PanelListItem/CardRow/CardListItem/SortableItemCustom: Grid template adjustments


philschanely:
- [x] Make `grid_template` optional on the following Rails and React components:
  - [x] PanelRow
  - [x] PanelListItem
  - [x] CardRow
  - [x] CardListItem
  - [x] SortableItemCustom
- [x] Ensure the layout without a grid template provided is a pleasing default for all these.
  - Default is a CSS grid set to `justify-content: space-between` which may work fine. Consider whether switching to `justify-content: start` might be a better default?
  - This is likely safe to modify as long as `grid_template` functionality remains same when set since all current implementations of these components have had the `grid_template` enforced up to this point.
- [x] Add new `gap` override option on these as well.
  - Existing default gaps should persist to ensure the correct gaps are applied based on context (32 for stage, 24 for panel, 16 for card).
  - However, within these, other gaps should also be possible to specify from existing spacer tokens and override the default above.

PR #974, opened 2021-11-01
[SAGE-935] Panel, Card, Sortable Optional Grid Template & Gap

pixelflips:
## Description
Makes `grid-template` optional and adds optional `gap` attribute to the following components:

-  PanelRow
-  PanelListItem
-  CardRow
-  CardListItem
-  SortableItemCustom

Note: the current default gaps will not change but additional sizes can now be passed in as needed.

## Screenshots
There are no visual changes related to these adjustments.

## Testing in `sage-lib`

PANEL:

- Navigate to [Panel](http://localhost:4000/pages/component/panel) ([Storybook](http://localhost:4100/?path=/docs/sage-panel--default))
- Verify no layout has changed.
- Try removing grid template or adding gap attribute to `panelRow` or `panelListItem`

CARD:

- Navigate to [Card](http://localhost:4000/pages/component/card) ([Storybook](http://localhost:4100/?path=/docs/sage-card--default))
- Verify no layout has changed.
- Try removing grid template or adding gap attribute to `cardlRow` or `cardListItem`

SORTABLE:

- Navigate to [Sortable](http://localhost:4000/pages/component/sortable)
- Verify no layout has changed.
- Try removing grid template or adding gap attribute to `SortableItemCustom`

1. (**MEDIUM**) Make grid-template optional and adds an optional gap attribute to components.
   - [ ] A sanity check in KP should be performed to verify no layout issues appear.

## Related
#935

### DIFF
--- a/docs/app/views/examples/components/card/_props.html.erb
+++ b/docs/app/views/examples/components/card/_props.html.erb
@@ -25,3 +25,33 @@
   <td><%= md('Optionally: `"sm"`') %></td>
   <td><%= md('`--`') %></td>
 </tr>
+<tr>
+  <td colspan="4"><%= md('**Card List Item**') %></td>
+</tr>
+<tr>
+  <td><%= md('`grid_template`') %></td>
+  <td><%= md('Sets the grid template to be used in the row of the component. See the [Grid Templates](http://sage-design-system.kajabi.com/pages/layout/grid_templates) to decide which pattern.') %></td>
+  <td><%= md('String') %></td>
+  <td><%= md('`nil`') %></td>
+</tr>
+<tr>
+  <td><%= md('`gap`') %></td>
+  <td><%= md('Adjusts the size of the gap between items in a list item.') %></td>
+  <td><%= md('Optionally: [`:xs` | `:sm` | `:md` | `:lg`]') %></td>
+  <td><%= md('`nil`') %></td>
+</tr>
+<tr>
+  <td colspan="4"><%= md('**Card Row**') %></td>
+</tr>
+<tr>
+  <td><%= md('`grid_template`') %></td>
+  <td><%= md('Sets the grid template to be used in the row of the component. See the [Grid Templates](http://sage-design-system.kajabi.com/pages/layout/grid_templates) to decide which pattern.') %></td>
+  <td><%= md('String') %></td>
+  <td><%= md('`nil`') %></td>
+</tr>
+<tr>
+  <td><%= md('`gap`') %></td>
+  <td><%= md('Adjusts the size of the gap between items in a card row.') %></td>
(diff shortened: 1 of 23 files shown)


pixelflips:
@QuintonJason not sure how much or if any of this might help/overlap with any of your sortable work, but just pinging you here just in case.

docs/lib/sage_rails/app/sage_components/sage_card_row.rb
```diff
@@ -1,6 +1,7 @@
 class SageCardRow < SageComponent
   set_attribute_schema({
-    grid_template: SageSchemas::GRID_TEMPLATE,
+    grid_template: [:optional, NilClass, SageSchemas::GRID_TEMPLATE],
+    gap: [:optional, NilClass, Set.new([:xs, :sm, :md, :lg])],
```
joedakroub, line 4: For my own understanding, is the `NilClass` needed for the `gap` property?

pixelflips: @joedakroub from my understanding we need it because Sortable consumes the SageCardRow. @philschanely had a much better explanation so hopefully, he can chime in as well.

joedakroub (approved):
I tested this locally and all looks well from what I can see. I added one minor comment, not a blocker. This looks great Phillip! Nice work. 👍🏼

monkeypox8 (approved):
👍 LGTM! Really helpful changes here. Great work!

docs/app/views/examples/components/card/_props.html.erb
```diff
@@ -25,3 +25,33 @@
   <td><%= md('Optionally: `"sm"`') %></td>
   <td><%= md('`--`') %></td>
 </tr>
+<tr>
+  <td colspan="4"><%= md('**Card List Item**') %></td>
+</tr>
+<tr>
+  <td><%= md('`grid_template`') %></td>
+  <td><%= md('Sets the grid template to be used in the row of the component. See the [Grid Templates](http://sage-design-system.kajabi.com/pages/layout/grid_templates) to decide which pattern.') %></td>
```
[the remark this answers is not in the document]
pixelflips: @monkeypox8 great catch! I'll adjust those and it does look like we can use relative URLs just fine.

docs/lib/sage_rails/app/sage_components/sage_card_row.rb
philschanely: I have found this is helpful for components that may be nested within others and where such values are populated by that parent component. In those cases if the value might possibly be not available, the child component would error out since the `nil` value would not be allowed. I have therefore taken to making `NilClass` a possible value on any optional properties since then.

PR #980, opened 2021-11-03
Avatar: Add image layering with initials

### LINKED ISSUE #975: [Improvement] Avatar: Add layering of image and initials
## Description

Revise Avatar so that if an image is provided it merely layers on top of initials, (assuming they are provided as well).

This is needed for how the Gravatar API works where an image is always provided so that a fallback could be present. Yet, that fallback/default image is not ideal. However, the API does allow for a transparent PNG to be the fallback, in which case we'd like to see the initials SVG showing through.

philschanely:
## Description

This PR adds the ability to layer initials and an image. This allows for both lazy loading of the initials while the image loads and for a fallback to be visible if the image does not load or is transparent (as in the use case for working with the Gravatar API).

## Testing in `sage-lib`

See Avatar in main site and Storybook. 

1. (LOW) Avatar updated to allow initials to layer behind an image for lazy loading and fallback. No adverse impact on existing uses expected. See the following:
   - [ ] People > table shows avatar initials as before
   - [ ] App header > shows avatar with initials or image as before

## Related

Closes #975

### DIFF
--- a/docs/app/views/examples/components/avatar/_preview.html.erb
+++ b/docs/app/views/examples/components/avatar/_preview.html.erb
@@ -64,3 +64,23 @@ NOTE: This also allows for a fallback approach with image APIs such as Gravatar 
 In such cases a transparent PNG image can be provided as the default image so that when this default image is layered on top of the initials,
 the initials are seen rather than a generic "anonymous user" image.
 ') %>
+
+<p>Invalid Gravatar with blank fallback to initials:</p>
+<%= sage_component SageAvatar, {
+  initials: "PS",
+  image: { src: "https://s.gravatar.com/avatar/000?default=blank" }
+} %>
+
+<p>Valid Gravatar with lazy loading:</p>
+<%= sage_component SageAvatar, {
+  initials: "PS",
+  image: { src: "https://www.gravatar.com/avatar/205e460b479e2e5b48aec07710c08d50?" }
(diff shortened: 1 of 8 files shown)


docs/app/views/examples/components/avatar/_preview.html.erb
```diff
@@ -53,3 +53,14 @@
...
+<h3>Layering and lazy loading</h3>
+<%= md('
+By default, Avatars with both an `image` and `initials` provided will stack with the image on top of the initials.
+The result is a lazy loading effect where initials appear immediately and an image will appear later on slower or delayed connetions.
+If this effect is not desired when an `image` is used, then also set `lazy_load_initials: false`.
+
+NOTE: This also allows for a fallback approach with image APIs such as Gravatar that return a default image when no user image is available.
+In such cases a transparent PNG image can be provided as the default image so that when this default image is layered on top of the initials,
+the initials are seen rather than a generic "anonymous user" image.
+') %>
```
QuintonJason, line 66: suggestion to an example of one

QuintonJason (approved):
Let a non-blocking comment. Looks and works great!

anechol (approved):
🚀

monkeypox8 (approved):
Looks like I just missed the cut, but LGTM!

PR #982, opened 2021-11-03
[Improvement] Dropdown: make options dropdown able to change subtle trigger

philschanely:
## Description

In React, the OptionsDropdown was forcing `triggerButtonSubtle` on the nested Dropdown to `true` but due to recent button changes this is desirable to be changeable on the OptionsDropdown. It now defaults to `true` but can be set otherwise when desired.

## Testing in `sage-lib`

See Storybook > Dropdown > Option menu

1. (LOW) React OptionsDropdown now has ability to set whether contained trigger is `subtle: true` (default) or not. No impact on existing default configurations.

### DIFF
--- a/packages/sage-react/lib/Dropdown/OptionsDropdown.jsx
+++ b/packages/sage-react/lib/Dropdown/OptionsDropdown.jsx
@@ -14,6 +14,7 @@ export const OptionsDropdown = ({
   panelMaxWidth,
   panelSize,
   options,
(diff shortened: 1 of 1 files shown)


monkeypox8 (approved):
👍🏼 LGTM! Just confirming that the `grid_template` optional modification is being carried over from the previous updates and not being re-introduced here.

pixelflips (approved):
LGTM! 👍🏼

PR #986, opened 2021-11-03
Layout Stacking Order and Assistant Updates

cmcfadzean:
## Description
The new design direction moves away from a top bar that spans the full top of the page and back to our current structure with a full-height sidebar. This change reflects this direction and gives us the ability to begin the transition in our super admin and also ideally in our admin layout. 

## Screenshots

|  Before  |  After  |
|--------|--------|
Screen Shot 2021-11-03 at 11 52 21 AM
|
Screen Shot 2021-11-03 at 11 04 56 AM

## Testing in `sage-lib`
The logo should direct back to the index and the layout should scroll correctly. Also ensuring that the menu toggles on mobile works as expected. 

This change will require an update to the super admin to ensure that this does not break the existing layout. 
Checking the status of the super admin is key before a bump with this change is landed


docs/app/views/layouts/application.html.erb
```diff
@@ -6,14 +6,11 @@
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/devicons/devicon@master/devicon.min.css">
   </head>
   <%= sage_component(SageBody, { css_classes: "sage-docs" }) do %>
-    <%= render "application/content_skip_link",
```
philschanely, line 9: We should probably keep this aspect intact for accessibility. Perhaps it can be provided through an optional slot in the Branding or added manually into the local sidebar here in Docs?

philschanely (approved):
Looks good to me. 👍  One small thing is to just bring this layout to the other main layouts; sandbox at least needed it.

ju-Skinner:
Closing...we currently have a redesign of the doc site in the works

PR #992, opened 2021-11-04
Update proptypes for Card row

philschanely:
## Description

This PR patches a few PropTypes issues with the CardRow for a quieter console.

1. (LOW) Patch to proptypes issue in CardRow. No impact on existing uss.

### DIFF
--- a/packages/sage-react/lib/Card/CardRow.jsx
+++ b/packages/sage-react/lib/Card/CardRow.jsx
@@ -32,18 +32,21 @@ export const CardRow = ({
   );
 };
 
+CardRow.GAP_OPTIONS = CARD_ROW_GAP_OPTIONS;
+CardRow.ALIGNMENT_OPTIONS = CARD_ROW_ALIGNMENT_OPTIONS;
+
 CardRow.defaultProps = {
   children: null,
   className: '',
   gridTemplate: null,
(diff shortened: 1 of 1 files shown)


pixelflips (approved):
LGTM! 👍🏼

packages/sage-react/lib/Card/CardRow.jsx
```diff
@@ -32,18 +32,21 @@ export const CardRow = ({
   );
 };
 
+CardRow.GAP_OPTIONS = CARD_ROW_GAP_OPTIONS;
```
monkeypox8, line 35: I was looking at the config and both `START` and `TOP` settings are assigned to `start`, which got me curious. Any idea if this was by intention?

monkeypox8 (approved):
👍🏼 LGTM! Left one non-blocking question in the config.

packages/sage-react/lib/Card/CardRow.jsx
philschanely: Shoot I just missed this before merging. Will look closer but yes I do believe it is. Idea is to do both a "CSS" thinker's approach and a more developer-focused. AKA if a dev wants to align do they think of that as "top" or the css property of "start"? Providing both here.

PR #993, opened 2021-11-04
React select patches

### LINKED ISSUE #990: [BUG] Select: React Select forces label as first item in option list and has incorrect error color
## Detailed description

- [x] React Select forces the `label` provided as the first `option` in the option list. We should provide other options for when this is not desired.
- [x] React Select also does not enforce proper color on error message when in error state.

## Steps to reproduce
1. In Storybook add a `label` to Select and note that the label appears as the first option without any way to adjust this.
2. Also add a message and set `hasError` true to see that the coloring is not as expected.

## Screenshots

Field showing label and value selected
Screen Shot 2021-11-04 at 1 31 52 PM

Field with options enabled (note the label appears as first option)
Screen Shot 2021-11-04 at 1 31 47 PM

Incorrect error state coloring

Screen Shot 2021-11-04 at 2 31 40 PM

philschanely:
## Description

- [x] React Select forces the `label` provided as the first `option` in the option list. We should provide other options for when this is not desired.
- [x] React Select also does not enforce proper color on error message when in error state.

## Screenshots

|  Before  |  After  |
|--------|--------|
| Screen Shot 2021-11-04 at 1 31 52 PM | Screen Shot 2021-11-04 at 4 15 01 PM |
| Screen Shot 2021-11-04 at 1 31 47 PM | Screen Shot 2021-11-04 at 4 15 05 PM |
| Screen Shot 2021-11-04 at 2 31 40 PM | Screen Shot 2021-11-04 at 4 15 22 PM |

## Testing in `sage-lib`

See Storybook > Select

1. (**LOW**) React Select component is updated with fix to show error colors and label as expected. No adverse effect expect on existing uses.

## Related

Closes #990

### DIFF
--- a/packages/sage-react/lib/Select/Select.jsx
+++ b/packages/sage-react/lib/Select/Select.jsx
@@ -19,6 +19,7 @@ export const Select = ({
     'sage-select',
     className,
     {
+      'sage-form-field--error': hasError,
       'sage-select--error': hasError,
       'sage-select--value-selected': value,
     }
--- a/packages/sage-react/lib/Select/Select.jsx
+++ b/packages/sage-react/lib/Select/Select.jsx
@@ -9,6 +9,7 @@ export const Select = ({
   hasError,
   id,
   label,
+  includeLabelInOptions,
   message,
   onChange,
   options,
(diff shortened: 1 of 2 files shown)


QuintonJason (approved):
Tested with and without an option selected with and without the error state. Looks and works great!

joedakroub (approved):
This is a nice update Phil. Nice work. 👍🏼

PR #1010, opened 2021-11-10
fix(panel): updates to gap

QuintonJason:
## Description

- [x] update `gap` to fire for sage panel rows

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2021-11-09 at 7 43 12 PM|Screen Shot 2021-11-09 at 7 46 04 PM|

## Testing in `sage-lib`

In the rails or react panel page, add a `gap` to the row and verify that it's working

1. (**LOW**) Doesn't affect Panel Rows currently as this is a new addition.

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_panel.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_panel.scss
@@ -114,7 +114,7 @@
 }
 
 @each $-key, $-value in $sage-spacings {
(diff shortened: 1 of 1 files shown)


philschanely (approved):

pixelflips (approved):
Great catch!! Thanks for updating! LGTM 👍🏼

PR #1017, opened 2021-11-10
Reorganize for patterns

philschanely:
## Description

This PR is a PoC for how we could start to be more intentional about prescribing common patterns. It involves a little reorganizing (still in process) in order to make room for the `Patterns` area to be a discrete list of such templates for easy copy-and-paste.

- A pattern is a template-like configuration of components that are commonly used together.
- Examples of patterns here show a preview of the pattern and a code sample that can be easily copied and pasted into code and then populated with the desired content and adjusted as needed.

## Screenshots

Screen Shot 2021-11-10 at 3 39 15 PM

## Testing in `sage-lib`

See Patterns section. First two links should be reconsidered for a better location in the navigation structure. Here we suggest that we could have patterns for Common Layouts, Forms, and Actions.

1. (**LOW**) Docs-only addition of a Patterns section.


docs/app/views/patterns/_example.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      id: "code-example-tabs-#{example[:partial]}}",
+      items: [
+        {
+          text: "Preview",
+          target: "preview-#{example[:partial]}",
+          active: true
+        },
+        {
+          text: "Code",
+          target: "code-#{example[:partial]}",
```
monicawheeler, line 15: ❤️

monicawheeler (approved):

monkeypox8 (approved):
😍 🔥 Love it! Numerous devs have asked for patterns to follow in various circumstances, so this will go a long way towards helping that cause.

Left one comment about setting the width on `sage-tabs-container`; primarily a precaution since I wasn't sure when this PR will merge to KP.

philschanely:
Closing for now as the system has changed enough since this was put in that we need a fresh start if/when we rekindle the direction.

PR #1026, opened 2021-11-12
feat(icon copy): add feature to copy icon code to clipboard

monicawheeler:
## Screenshots 

|  Before - no copy behavior  |  After - `title` attribute  |  After - toast triggered  |
|--------|--------|--------|
|Screen Shot 2021-11-12 at 3 48 49 PM | Screen Shot 2021-11-12 at 3 49 31 PM | Screen Shot 2021-11-12 at 3 49 35 PM |

## Gif of interaction

copy-icon

## Example of code directly pasted from interaction
`<%= sage_component SageIcon, { icon: "handle-2-vertical" } %>`

## Testing in `sage-lib`
1. Navigate to http://localhost:4000/pages/foundations/icon
2. Scroll to "All Icons"
3. Click on the icon image
4. Paste copied `SageIcon` anywhere you need!

1. (**N/A**) Doc site only updates.


docs/app/views/pages/icon.html.erb
```diff
@@ -256,9 +256,15 @@ Marketing uses the Megaphone icon. We do not use the megaphone icon for other tr
   <h2>All Icons</h2>
   <%= sage_component SageGridRow, {} do %>
     <% SageTokens::ICONS.each do |icon| %>
-      <%= sage_component SageGridCol, { breakpoint: "md", size: "4" } do %>
-        <div class="docs-icon-block">
-          <i class="sage-icon-<%= icon %>-2xl" aria-hidden="true"></i>
+      <%= sage_component SageGridCol, { breakpoint: "md", size: "3" } do %>
+        <div
+          class="docs-icon-block"
+          data-js-copy-button="<&#37;= sage_component SageIcon, { icon: output correct icon here } &#37;>"
```
monicawheeler, line 262: @monkeypox8 - this is the line of code that I was having trouble with. I am unable to get the name of the icon to output for the value in `icon: icon`. As you can see `<%= icon %>` outputs perfectly for line 263, so I am guessing it is something in the component string because of the mixed content in the `data-js-copy-button` attribute. If you have a moment or two to review and provide suggestions, that would be great! Thank you!

monkeypox8: Oh, you were already 99.9% of the way there. Just need to add the same `<%= icon %>` in the output of the mixed string:
```
data-js-copy-button="<&#37;= sage_component SageIcon, { icon: <%= icon %> } &#37;>"
```

docs/app/views/pages/icon.html.erb
```diff
@@ -256,9 +256,15 @@ Marketing uses the Megaphone icon. We do not use the megaphone icon for other tr
   <h2>All Icons</h2>
   <%= sage_component SageGridRow, {} do %>
     <% SageTokens::ICONS.each do |icon| %>
-      <%= sage_component SageGridCol, { breakpoint: "md", size: "4" } do %>
-        <div class="docs-icon-block">
-          <i class="sage-icon-<%= icon %>-2xl" aria-hidden="true"></i>
+      <%= sage_component SageGridCol, { breakpoint: "md", size: "3" } do %>
+        <div
+          class="docs-icon-block"
+          data-js-copy-button="<&#37;= sage_component SageIcon, { icon: &ldquo;<%= icon %>&rdquo; } &#37;>"
```
monicawheeler, line 262: Are these character entities for left and right quotes to correct ones? `&#34;` might be valid too? thoughts?

Referenced: https://dev.w3.org/html5/html-author/charref

joedakroub: These left/right quotes are fancy quotes. I think we can just use `&quot;` around the icon name and that should suffice.

joedakroub (approved):
This is great Monica; super helpful! Nicely done. 👍🏼

monkeypox8 (approved):
👍🏼 LGTM, a very helpful improvement! As @joedakroub suggested, swapping out the character with the generic `&quot;` seems to be the best path forward.

pixelflips (approved):
Nothing to add, but this is 🔥 !! Great work and thanks. Super useful!

PR #1030, opened 2021-11-15
[Sage-966] Modal - Updates Docs with Sage Tokens

### LINKED ISSUE #966: [Improvement] Documentation - update modal docs
## Description

updated the modal rails docs to use helper token instead of hard-coded `t-sage-` classes

anechol:
## Description

Updates Modal docs to use Sage tokens instead of hard-coded `t-sage` classes.

## Screenshots

N/A

## Testing in `sage-lib`

Check that there are no usages of hard-coded Sage classes on the Preview page.

(**LOW**) Updates Modal docs to use Sage tokens instead of hard-coded `t-sage` classes. No effect on Kajabi Products

## Related

Closes #966

### DIFF
--- a/docs/app/views/examples/components/modal/_preview.html.erb
+++ b/docs/app/views/examples/components/modal/_preview.html.erb
@@ -89,12 +89,12 @@
         } %>
       <% end %>
 
-      <p class="t-sage-body">
+      <p class="<%= SageClassnames::TYPE::BODY %>">
         Lorem ipsum dolor sit amet, consectetur adipiscing elit,
         sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
       </p>
 
-      <p class="t-sage-body">
+      <p class="<%= SageClassnames::TYPE::BODY %>">
         Lorem ipsum dolor sit amet, consectetur adipiscing elit,
         sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
       </p>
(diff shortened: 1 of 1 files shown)


pixelflips (approved):
LGTM! 👍🏼

docs/app/views/examples/components/modal/_preview.html.erb
```diff
@@ -89,12 +89,12 @@
         } %>
       <% end %>
 
-      <p class="t-sage-body">
+      <p class="<%= SageClassnames::TYPE::BODY %>">
         Lorem ipsum dolor sit amet, consectetur adipiscing elit,
```
monkeypox8, line 93: Not a blocker, but we may want to consider making this placeholder string a variable since it's being output in 10+ other examples.

monkeypox8 (approved):
👍🏼 LGTM! One item to consider, not a blocker.

Also, unrelated to the work in this PR, but the first row of modal trigger buttons is breaking out of the page container. A short term fix coul be to add `wrap: true` onto the button group, though mobile still overflows the container bounds. Longer term, we'll likely want to reorganize our examples on this page.

monicawheeler (approved):

docs/app/views/examples/components/modal/_preview.html.erb
anechol: Thanks for this! As far as the button wrapping issue, I've fixed that for now, but yes, the mobile wrapping issue is still there.

PR #1032, opened 2021-11-16
[Sage-942] Responsive PanelTile

### LINKED ISSUE #942: [Bug] PanelTiles and CardTiles: Add responsive considerations
SagePanelTile and SageCardTile both are set up to allow tiles in a 2-up, 3-up, or 4-up layout but as yet have no responsiveness considered.

- [x] Add breakpoints that change 4-up to 2-up below desktop and single column below tablet.
- [x] Add breakpoints that change 3-up to 2-up below desktop and single column below tablet.
- [x] Add breakpoints that change 2-up to single column below tablet.

Note, however, that these tiles can be contained inside Panels or Cards that themselves are narrow. We may therefore also want to consider either or both of:
- Find a way to provide breakpoints for where transitions can happen at custom breakpoints. This could provide the most fine-grained control but require configuration.
- Assuming that such a card or panel is inside a SageContainer that employs one of our standard container sizes, experiment with adding contextual breakpoints for where such tiles begin to be crowded. For example:
  - Given a `lg` container with a Panel and a 4-up tile set inside, when should the 4-up switch to 2-up?
  - Given a `sm` container with a Panel and a 2-up tile set inside, when should the 2-up switch to single column?
 
   This might provide better out-of-the-box functionality but still may not fit common uses of the component.

FuturaExtraBold:
## Description
Adds responsiveness for PanelTile. All instances of 4-up, 3-up, and 2-up change to 2-up on a tablet, and 1-up on mobile.

## Related
Closes #942 

## Screenshots
|  Desktop  |  Tablet  |  Mobile  |
|--------|--------|--------|
|image|image|image|

### Rails
- View [Rails component](http://localhost:4000/pages/component/panel)
- Check that the responsive layout works consistently no matter the number of tile columns.

### React
- View [Storybook](http://localhost:4100/?path=/story/sage-panel--panel-tiles)
- Check that the responsive layout works consistently no matter the number of tile columns.

1. (**LOW**) This simply adjusts the number of panel tiles on different screens. No updates or retrofitting is necessary.

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_panel.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_panel.scss
@@ -165,35 +165,26 @@
 }
 
 .sage-panel__tiles {
-  display: grid;
-  grid-gap: sage-spacing();
-  list-style: none;
-}
+  --sage-panel-tiles-gap: #{sage-spacing()};
 
-@for $i from 2 through 4 {
-  .sage-panel__tiles--#{$i}-up-condensed {
-    grid-gap: sage-spacing(xs);
+  display: grid;
+  grid-gap: var(--sage-panel-tiles-gap);
+  grid-template-columns: repeat(var(--sage-panel-tiles-columns), calc((100% - (var(--sage-panel-tiles-gap) * (var(--sage-panel-tiles-columns) - 1))) / var(--sage-panel-tiles-columns)));
+  list-style: none;
+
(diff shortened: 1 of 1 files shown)


anechol (approved):
LGTM

monicawheeler (approved):
🙌 Looks great!

philschanely (approved):

PR #1037, opened 2021-11-17
[Sage-885] Description - Design Spec Updates

### LINKED ISSUE #885: [Enhancement] Description: Update to new specs
- [x] ~~**BLOCKER: Need Design to integrate in Figma Specs for Single Source of Truth** Contact @xtinesun for status~~ New specs located [here](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/Sage-3-for-Admin?node-id=7780%3A20536)
- [x] Lets definitely take this opportunity to remove the direct title, data, and link attributes in lieu of these being on `items` only.
- [x] The current default for the layout is aligned to the Stacked layout in the new mocks. This should be adjusted to align to the new Default layout versus a property to enable the `stacked` layout. I'd suggest a `layout` property that can accept `stacked` as a value. This leaves us a little more open to future potential variations. Then the current property `inline_spread` can be removed.
- [x] In lieu of the current `link` property for each item let's use a new `action` property more. This should map to an optional Button within the component where some fixed aspects are set up (primary, subtle) where other settings can be provided as a hash here such as the `value`, `attributes`, and if/how the icon should appear. See the recent updates to `Alert` where we do something similar for `primary_action`. The default configuration if an `action` is added should be the icon only version.
- [x] Update type specs for all variations to match prescribed by design spec. Along these lines the current `allcaps_titles` can be removed as it is no longer needed with the new specs.
- [x] New spacing and border around/between "rows"
- [x] Default bottom spacing should be removed when within `SagePanel` context (if not altogether... may need to audit use/impact)
- [x] The layout for the individual rows should use CSS Grid such as `grid-template: var(--sage-description-title-width, auto) 1fr` if there are no actions present in the `items` or `grid-template: var(--sage-description-title-width) 1fr var(--sage-description-action-width, auto)`. Then some new props on the component should allow the developer to provide values that set specific values for these two CSS custom props. `auto` may be sufficient, but if specific values are needed consider `50%` for `title` and for the `action` a width that assumed a default action will be just a subtle button with the arrow icon (icon only).
- [x] ~~Description is not built at all in React yet. A quick stub of it is in #878 that can be pulled over here and synced with the new design direction.~~ added the component by accident when the PR above was closed 😳 

## Design Guidance

See comments in sidebar in Figma spec.

Notes on current design spec:

- We should not be concerned by the label group rows that show LabelGroup as a `data` value. This should be handled either as a `data` value or as a separate element in the app context.
- Along the lines of the above, there is no real need at this time for the "no type" column variation depicted in the mocks. This was only ever intended for the label group context.
- There is no responsive relationship between stacked and inline; they are totally separate configurations.

anechol:
## Description

Updates the Description to new design specs and props.

[Figma](https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/Sage-3-for-Admin?node-id=7780%3A20536)

## Screenshots

Screen Shot 2021-12-01 at 10 40 54 AM
Screen Shot 2021-12-01 at 10 41 04 AM
Screen Shot 2021-12-01 at 10 41 16 AM
Screen Shot 2021-12-01 at 10 41 23 AM

### Rails
- View [Rails component](http://localhost:4000/pages/component/description)
- Check that component renders as expected and new props function as intended.

### React
- View [Storybook](http://localhost:4100/?path=/docs/sage-avatar--with-badge)
- Check that new prop functions as expected 

(MEDIUM) Updates the Description to new design specs and props. Some props have been deprecated. Patch Kajabi/kajabi-products#21609 should merge in with version bump.

## Related

Closes #885


docs/app/views/examples/components/description/_preview.html.erb
```diff
@@ -1,26 +1,56 @@
...
+      title: "Name",
+      data: "John Doe",
+    }
+  ],
 } %>
 
-<h2>Multiple items</h2>
+<h3>Mulitple Items - Stacked</h3>
 <%= sage_component SageDescription, {
+  primary_action: {
```
philschanely, line 35: Is this still relevant?

docs/app/views/examples/components/description/_preview.html.erb
```diff
@@ -1,26 +1,56 @@
...
+    value: "Button",
+  },
+  layout: "stacked",
   items: [
     {
       title: "Name",
       data: "John Doe",
+      primary_action: {
```
philschanely, line 43: This could likely be just `action` as I'm not sure about any need for `secondary_action`, etc.

docs/app/views/examples/components/description/_props.html.erb
```diff
@@ -39,14 +20,24 @@ Array<{
...
+  icon_only: Boolean,
+  value: String,
+  test_id: String,
+}
+```')
+    %>
+  </td>
   <td><%= md('`nil`') %></td>
-</tr>
+</tr>
```
monicawheeler, line 43: Minor note: end of line missing here.

docs/lib/sage_rails/app/sage_components/sage_description.rb
```diff
@@ -2,16 +2,20 @@ class SageDescription < SageComponent
   # TODO: simplify to items only (remove data, link, title from root)
```
monicawheeler, line 2: Is this comment still necessary?

docs/app/views/examples/components/description/_props.html.erb
```diff
@@ -39,14 +20,24 @@ Array<{
...
 </tr>
 <tr>
-  <td><%= md('`title`') %></td>
-  <td><%= md('The "title" or term for the element.') %></td>
-  <td><%= md('`String`') %></td>
+  <td><%= md('`primary_action`') %></td>
+  <td><%= md('Creates a custom primary action button in the last column of the row.') %></td>
+  <td>
+    <%= md('```
+primary_action: {
```
philschanely, line 33: I see this as part of the items (where I think it belongs) but not on the root schema; should this be moved?

packages/sage-assets/lib/stylesheets/components/_description.scss
```diff
@@ -19,45 +19,47 @@
...
-  justify-content: space-between;
-}
-
 .sage-description__term-group {
-  margin-bottom: sage-spacing(2xs);
+  display: grid;
+  grid-template-columns: var(--sage-description-title-width, 50%) 1fr var(--sage-description-action-width, auto);
+  gap: sage-spacing(sm);
+  width: 100%;
+  padding: sage-spacing(xs) 0;
```
monicawheeler, line 27: It looks like the design shows padding at 24px top and bottom. Is this something that might need to be customizable?

docs/app/views/examples/components/description/_props.html.erb
```diff
@@ -1,32 +1,13 @@
...
-  <td><%= md('`inline_spread`') %></td>
-  <td><%= md('If `true` the title/data pairs will render in line with each other and spread apart, with the `title` left-aligned and the `data` right-aligned.') %></td>
-  <td><%= md('`Boolean`') %></td>
-  <td><%= md('`false`') %></td>
-</tr>
 <tr>
   <td><%= md('`items`') %></td>
   <td><%= md('
     An array of items to create a set of `title`: `description` pairs.
-    If this is provided, standalone `title` and `data` properties are ignored in favor of rendering this set of items.
```
philschanely, line 29: `items` should the full schema that matches the items, so `primary_action` (or just `action`?) should move here in some fashion.

docs/lib/sage_rails/app/views/sage_components/_sage_description.html.erb
```diff
@@ -1,47 +1,48 @@
...
   class="
     sage-description
-    <%= "sage-description--inline-spread" if component.inline_spread.present? && component.inline_spread %>
-    <%= "sage-description--normalcase-titles" if normalcase_titles %>
+    <%= "sage-description--stacked" if component.layout == "stacked" %>
     <%= component.generated_css_classes %>
   "
+  <% if component.action_width || component.title_width %>
+    style="
+      --sage-description-action-width: <%= component.action_width || "auto" %>;
```
philschanely, line 10: We may want either to set default fixed values for these or make them conditional altogether to resort to the value set in the SCSS.

docs/lib/sage_rails/app/views/sage_components/_sage_description.html.erb
```diff
@@ -1,47 +1,48 @@
...
-              <dd class="sage-description__data"><%= item[:data] %></dd>
-            <% end %>
-          <% else %>
+  <% component.items.each do | item | %>
+    <div class="sage-description__term-group">
+      <% if item[:title] %>
+        <dt class="sage-description__title"><%= item[:title] %></dt>
+      <% end %>
+      <% if item[:data] %>
+        <% if item[:link] %>
```
philschanely, line 22: I think we can actually remove `link` and only use the `action` instead. (acknowledging this may require refactoring in `kp` but I think we're moving towards the link always being at the end of the row not on the value itself).

packages/sage-assets/lib/stylesheets/components/_description.scss
```diff
@@ -19,45 +19,47 @@
...
+    border-bottom: sage-border();
   }
 
-  .sage-description--inline-spread & {
-    display: flex;
-    justify-content: space-between;
-    width: 100%;
+  .sage-description--stacked & {
+    grid-template-areas:
+      "title title button"
```
philschanely, line 37: 😍

monicawheeler:
Hi Ashley! Looking great so far. I added a couple of notes and also these can be added to the `components_helper.rb` file.

```
react_component_slug: "sage-description--default",
figma_embed: "https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2F9Km09NjlZHYWsMP7EGT8tI%2FSage-3-for-Admin%3Fnode-id%3D19501%253A25347",
```

packages/sage-assets/lib/stylesheets/components/_description.scss
```diff
@@ -19,45 +19,47 @@
...
-.sage-description--inline-spread {
-  display: flex;
-  flex-flow: row wrap;
-  justify-content: space-between;
-}
-
 .sage-description__term-group {
-  margin-bottom: sage-spacing(2xs);
+  display: grid;
+  grid-template-columns: var(--sage-description-title-width, 50%) 1fr var(--sage-description-action-width, auto);
```
philschanely, line 24: If we have no actions in any of the items we may end up with a null/empty column at the end but dang grid will still show the gap there. So we may need to have a modifier that is added if, after inspecting the provided `items` we find there are no `action`s to render. In such case we'd probably leave off this end column or user grid areas to display `data` in both middle and right column if that's simpler.

packages/sage-react/lib/Description/Description.jsx
```diff
@@ -56,44 +75,73 @@ export const Description = ({
...
 
-    return renderItem({ title, data, link });
+    return renderItem({ title, data, link, primaryAction });
+  };
+
+  const setCustomProps = () => {
+    const props = {};
+
+    if (actionWidth) {
+      props['--sage-description-action-width'] = actionWidth || 'auto';
```
philschanely, line 85: Love this approach. In this setup we shouldn't need the `auto` fallback.

philschanely (commented):
Off to a great start. Some changes suggested here.

anechol:
@monicawheeler @philschanely Thanks for the feedback! I'll get these changes in soon.

packages/sage-assets/lib/stylesheets/components/_description.scss
anechol: Looks like the padding was updated while I was working on this haha.

packages/sage-assets/lib/stylesheets/components/_description.scss
anechol: Should be fixed now!

anechol:
Updates:
- Removes `link` prop
- Prop documentation
- `action` prop
- Grid styling when no action is present
- CSS custom prop logic for `title` and `action`

packages/sage-assets/lib/stylesheets/components/_description.scss
philschanely: What if we always use the grid-template areas such that you don't have to change the columns, but just the span of areas? So this form would be `title data button`, the `--no-action` would be `title data data`, and the stacked form could be just as you have it?

docs/lib/sage_rails/app/views/sage_components/_sage_description.html.erb
```diff
@@ -1,47 +1,47 @@
...
-        <%= link_to component.link do %>
-          <dd class="sage-description__data"><%= component.data %></dd>
-        <% end %>
-      <% else %>
-        <dd class="sage-description__data"><%= component.data %></dd>
+  <% component.items.each do | item | %>
+    <div 
+      class="
+        sage-description__term-group
+        <%= "sage-description__term-group--no-action" if !item[:action] %>
```
philschanely, line 20: This is fine except I think design wants the gap column on all rows if even just one of them has an action. So consider adding this logic at the top:

```rb
has_actions = false
component.items.each do | item |
  if item[:actions]
    has_actions = true
  end
end
```

and then changing to `... if !has_actions` here.

packages/sage-react/lib/Description/Description.jsx
```diff
@@ -2,98 +2,141 @@ import React from 'react';
...
+        </div>
+      )}
     </>
   );
 
   const renderItems = () => {
     if (items && items.length > 0) {
       return items.map((item) => (
-        <div key={item.id || uuid()} className="sage-description__term-group">
+        <div
```
philschanely, line 64: Similar to comments in ruby file, lets add this logic at top of this function:

```jsx
let hasActions = false;
items.forEach(item => {
  if item.action {
    hasActions = true;
  }
});
```

This this can be `${!hasActions ? ...}`

anechol:
@philschanely Appreciate you taking this over in my absence. These changes look very good!

Issue #944, opened 2021-10-23
[Enhancement] Tokens: PoC for a classnames utility


philschanely:
Seeing more use of SageClassnames and SageTokens in the monolith there may be some benefit to creating some shorthand helpers that make providing these classnames with fewer characters. Consider:

```erb
<p class="<%= sage_classnames(type_spec: "body-sm", color: "charcoal-200", align: "center") %>">
  Lorem ipsum dolor sit...
</p>
```

would output:

```html
<p class="t-sage-body-sm t-sage--color-charcoal-200 t-sage--align-center">
  Lorem ipsum dolor sit...
</p>
```

Should consider all existing classname tokens as possible params and enforce the values that map to valid classes.

philschanely:
Moved to Jira https://kajabi.atlassian.net/browse/UXD-58

PR #1038, opened 2021-11-18
[Sage-944] Classnames Utility Proof of Concept

FuturaExtraBold:
## Description
Adds a classnames helper that converts an object to classes for styling.

## Related
#944


docs/app/views/examples/components/classnames_test/_preview.html.erb
```diff
@@ -0,0 +1,13 @@
+<h3>Default</h3>
+<% cn = sage_classnames({
+  align: "center",
+  base_classes: "alpha bravo charlie",
```
philschanely, line 4: Not opposed to this but technically one could put these in outside of our function. Might consider keeping ours focused just on what Sage offers?

philschanely:
@FuturaExtraBold this is off to a great start. Is the new component just for testing now? I think this could be documented and demonstrated under the Helpers menu instead of Components.

docs/app/views/examples/components/classnames_test/_preview.html.erb
FuturaExtraBold: Yes. I will remove the `base_classes`.

FuturaExtraBold:
> @FuturaExtraBold this is off to a great start. Is the new component just for testing now? I think this could be documented and demonstrated under the Helpers menu instead of Components.

@philschanely Yes. This is just for testing as I work. It should be one helper file when work is complete.

PR #1046, opened 2021-11-19
[SAGE-1042] Switch value output fix

### LINKED ISSUE #1042: [BUG] Sage Switch does not output `value` property as expected
## Detailed description
When a Sage Switch instance includes the `value` property, the rendered output is actually the `id`, and not the `value`. While this may have been intentional, the result is that any provided `value` will always be overridden by the `id`:

sage-switch

This occurs for both the default Sage Switch and its `standalone` variant, and can be seen on the [documentation site](https://sage-design-system.kajabi.com/pages/component/switch). While each example has a unique `value`, (ex. `sage-switch-1` and `sage-switch-1-value`), the rendered output only displays the `id`.

## Steps to reproduce
1. View the [Sage Switch documentation](https://sage-design-system.kajabi.com/pages/component/switch)
2. Click the "Code" tab, and expand the "Sage Component" markup
3. Verify that the `id` and `value` of one of the examples uses separate strings 
4. Scroll down and expand the "Generated HTML Code" markup 
5. Locate your example from step 3, and confirm that the `value` matches the `id` output

monkeypox8:
## Description
Corrects a bug in the Rails version of Sage Switch where `id` strings were output in place of `value` (see #1042 for full details).

- `value` is now optional for the Rails component. In current implementations, `value` was set as a duplicate of `id` in the majority of cases, suggesting that it was only being provided to bypass the component requirement
- When omitted, `value` falls back to `id` as a way to duplicate existing behavior 

Also addresses two unrelated typos on the documentation site.

## Screenshots
No visual changes

## Testing in `sage-lib`
1. View the Sage Switch documentation
2. Click the "Code" tab, and expand the "Sage Component" markup
3. Verify that the `id` and `value` of one of the examples have unique strings
4. Scroll down and expand the "Generated HTML Code" markup, or switch back to the "Preview" tab
5. Locate your example from step 3, and confirm that the `value` and `id` output as expected

1. (**MED**) Updates Sage Switch to output optional `value` property with fallback to `id`. Verify that instances save correctly and state is maintained on refresh. This includes uses of the shared `kajabi-products` partial `app/views/admin/shared/_sage_switch.html.erb`
   - [ ] Affiliates -> Affiliate commissions
   - [ ] Offers -> Offer Additional Settings -> (multiple items) Cart Abandonment, Affiliate Commissions (requires `comm_offer_detail`)
   - [ ] Landing pages -> Page Meta Fields
   - [ ] Website pages -> Form Fields

## Related
- Closes #1042

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_switch.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_switch.html.erb
@@ -1,15 +1,15 @@
 <% if component.standalone %>
-  <input 
-    type="<%= component.type %>" 
-    id="<%= component.id %>" 
+  <input
+    type="<%= component.type %>"
+    id="<%= component.id %>"
     class="
       sage-switch
       sage-switch--standalone
       <%= component.has_error ? "sage-switch--error" : "" %>
       <%= component.generated_css_classes %>
-    " 
-    name="<%= component.name %>" 
-    value="<%= component.id %>"
+    "
+    name="<%= component.name %>"
+    value="<%= component.value.present?  ? component.value : component.id %>"
     <%= "checked" if component.checked %>
     <%= "disabled" if component.disabled %>
     <%= "required" if component.required %>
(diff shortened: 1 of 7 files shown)


docs/lib/sage_rails/app/views/sage_components/_sage_switch.html.erb
```diff
@@ -28,12 +28,12 @@ tag = component.in_list ? 'li' : 'div'
...
-      id="<%= component.id %>" 
-      class="sage-switch__input" 
-      name="<%= component.name %>" 
-      value="<%= component.id %>"
+    <input
+      type="<%= component.type %>"
+      id="<%= component.id %>"
+      class="sage-switch__input"
+      name="<%= component.name %>"
+      value="<%= component.value.present?  ? component.value : component.id %>"
```
monkeypox8, line 36: I was also considering leaving the `value` as a required attribute, and instead adding an `omit_value` option for cases where that would be useful (ex. having a default `<name>=on` in the form data rather than `<name>=<value string>` when checked)

pixelflips (approved):
👍🏼  LGTM! Checked both docs and with the bridge for kp. Nice work!

monicawheeler (approved):
👍 - looks good! Verified in `kajabi-products` with bridge and states are being saved.

QuintonJason (approved):

PR #1048, opened 2021-11-22
fix(panel heading): fix spacing for title within panel heading

### LINKED ISSUE #1047: [BUG] SagePanelHeader: Space regression under title
## Detailed description

Too much space under title after update to PanelHeader

## Steps to reproduce
1. Navigate to an Offer, http://www.kajabi.test:3000/admin/offers/2/edit
2. Observe Offer Status and Offer Details title have too much space below

## Screenshots
Screen Shot 2021-11-22 at 11 17 21 AM

monicawheeler:
## Description

Spacing on `sage-panel__title` without `sage-panel__subtext` is causing too much spacing for heading.

## Fix

Move spacing to be contextual when subtext is present.

## Screenshots

|  Before  |  After  |
|--------|--------|
| before | after |

## Testing in `sage-lib`
1. Navigate to http://localhost:4000/pages/component/panel
2. Verify spacing is correct on title and subtext

1. (**MEDIUM**) Spacing in SagePanelHeader between title and subtext needs to be correct
   - [ ] Upsell offers panel - v3 Upsell offer 

## Related
Closes #1047

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_panel.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_panel.scss
@@ -88,8 +88,11 @@
 
 .sage-panel__title {
   @extend %t-sage-heading-5;
-  margin-bottom: sage-spacing(xs);
   color: sage-color(charcoal, 500);
+
+  & + .sage-panel__subtext {
+    margin-top: sage-spacing(xs);
+  }
 }
 
 .sage-panel__subtext {


packages/sage-assets/lib/stylesheets/components/_panel.scss
```diff
@@ -88,8 +88,11 @@
 
 .sage-panel__title {
   @extend %t-sage-heading-5;
-  margin-bottom: sage-spacing(xs);
```
monkeypox8, line 91: A minor side effect is with any Panel Titles that skip the subtext, we lose the additional 8px spacing between content and header. For example, on the Panel documentation page between "Basic Panel" and the example. I believe this is fairly rare in app use, however.

monkeypox8 (approved):
👍🏼

packages/sage-assets/lib/stylesheets/components/_panel.scss
monicawheeler: I agree that we lose that spacing but the inner panel items are spaced with `gap` currently, so the impact should be minimal, as you noted. Thank you for noting that!

PR #1055, opened 2021-11-29
fix(workflow): update-release workflow token switch

monicawheeler:
## Description

`update-release` workflow is not triggering. Trying to swap tokens around from release-deploy.yml to update-release.yml so the first workflow is not triggering off the default key. 

_Note: if this is successful, I will get Smitty to update the token name in the repo._

## Testing in `sage-lib`

1. Workflow view in https://github.com/Kajabi/sage-lib/actions

1. (**N/A**) Workflow changes for `sage-lib` only. No `kajabi-products` testing needed.

### DIFF
--- a/.github/workflows/release-deploy.yml
+++ b/.github/workflows/release-deploy.yml
@@ -38,7 +38,7 @@ jobs:
 
       # Setup Auth token to push to github packages
       - name: Set NPM Config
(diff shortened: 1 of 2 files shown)


philschanely (approved):
🤷‍♂️  👍

monkeypox8 (commented):
Interesting! Curious if it'd be useful to have at least one lead with full access.

PR #1075, opened 2021-12-03
[Sage-1041] Docs - Color Tokens

### LINKED ISSUE #1041: [Improvement] Docs: Add Color tokens to colors page
## Description

Colors page currently shows the scss token functions but the latest Rails and React tokens are probably more useful. Add the following for each color:

- [ ] Classname tokens (outputs `t-sage--color-X`)
   - Rails: `SageClassnames::TYPE_COLORS::X` (SAGE, SAGE_100, SAGE_200, etc.)
   - React: `SageClassnames.TYPE_COLORS.X`
- [ ] Hex token (outputs `#xxxxxx`)
   - Rails: `SageTokens::COLOR_PALETTE::X` (SAGE, SAGE_100, SAGE_200, etc.)
   - React: `SageTokens.COLOR_PALETTE.X`

anechol:
## Description

Adds color tokens as copy buttons to the Color page in Sage documentation.

## Screenshots

|  Before  |  After  |
|--------|--------|
Screen Shot 2021-12-08 at 3 24 31 PM|Screen Shot 2021-12-14 at 3 52 50 PM

## Testing in `sage-lib`

- Check that the new color information makes sense for the docs given the [original issue](https://github.com/Kajabi/sage-lib/issues/1041) filed
- Check that copy button properly copies the token

(**LOW**) Sage docs updates. No effect on Kajabi Products work.

## Related

Closes #1041

### DIFF
--- a/docs/app/views/pages/_color_values.html.erb
+++ b/docs/app/views/pages/_color_values.html.erb
@@ -1,5 +1,5 @@
 <div class="colors">
-  <div class="colors__block color-<%= color %>-300">sage-color(<%= color %>)</div>
+  <div class="colors__block color-<%= color %>-300"><%= color.capitalize %> 300</div>
 </div>
 <div class="colors">
   <% (1..5).each do |i| %>
--- a/docs/app/views/pages/_color_values.html.erb
+++ b/docs/app/views/pages/_color_values.html.erb
@@ -3,6 +3,19 @@
 </div>
 <div class="colors">
   <% (1..5).each do |i| %>
-    <div class="colors__block color-<%= color %>-<%= i %>00">sage-color(<%= color %>, <%= i %>00)</div>
+    <div class="colors__block color-<%= color %>-<%= i %>00">
+      <div class="colors__tokens">
+        <div class="colors__classname-token">
+          <h3>Classname Tokens</h3>
+          <p>Rails: <code>SageClassnames::TYPE_COLORS::<%= color.upcase %>_<%= i %>00</code></p>
+          <p>React: <code>SageClassnames.TYPE_COLORS.<%= color.upcase %>_<%= i %>00</code></p>
+        </div>
+        <div class="colors__hex-token"> 
+          <h3>Hex Tokens</h3>
+          <p>Rails: <code>SageTokens::COLOR_PALETTE::<%= color.upcase %>_<%= i %>00</code></p>
+          <p>React: <code>SageTokens.COLOR_PALETTE::<%= color.upcase %>_<%= i %>00</code></p>
+        </div>
+      </div>
+    </div>
   <% end %>
 </div>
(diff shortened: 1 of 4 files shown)


anechol:
Would love to get some feedback on this direction for displaying this new information.

monicawheeler:
What do you think about removing the explicit token names for Rails and React and just using `SAGE, SAGE_100, SAGE_200`, and so in the left column? This is the approach I am taking in the Typography documentation within the SageType PR.

anechol:
@monicawheeler Thanks for the suggestion. Overall, I'm wondering just how much information is necessary for the page, and also very conflicted. If there's a situation where the developer needs the entire variable in Rails, but has only ever worked in React, would that approach give them enough information? How do I draw the line between being unnecessarily verbose and giving just enough information? And between having a nice design and potentially being an eyesore? 🤔

monicawheeler:
@anechol - that is a good consideration. Perhaps it could be helpful to describe the tokens above the colors blocks in paragraph form. This could keep the color blocks cleaner and could look something like: 

> ## Rails Conventions
>_To use classname tokens follow this convention: 
>`SageClassnames::TYPE_COLORS::<token_name>`_
>
>_To use hex tokens follow this convention: 
>`SageTokens::COLOR_PALETTE::<token_name>`_
>
> ## React Conventions
>_To use classname tokens follow this convention:
>`SageClassnames.TYPE_COLORS.<token_name>`_
>
>_To use hex tokens follow this convention: 
>`SageTokens.COLOR_PALETTE.<token_name>`_

anechol:
@monicawheeler I like this idea! Gonna try this out and see how it looks.

anechol:
Here's a concept of what I have as a second option based on Monica's suggestion above:
Screen Shot 2021-12-09 at 1 15 57 PM

Screen Shot 2021-12-09 at 1 16 17 PM

One concern I have about this layout is how much work should we have users do with regards to copying the convention? Is having them type it out good enough, or would it make more sense for them to just copy and paste the exact color they need? Additionally, how would we feel about making each color copyable?

Any feedback is very welcome and appreciated.

philschanely:
@anechol consider `SAGE_400` in the sliders here which would mirror the token names closer?

anechol:
@cmcfadzean Personally, I'd prefer defaulting to classnames as that would prevent the developer from making changes to any page-specific CSS files. But the hex tokens have the benefit of being used within Sage components that take strings as props for colors, like icons or avatars.

joedakroub (approved):
This looks terrific! Really nice improvement; very convenient. Really nice work Ashley! 👍🏼

pixelflips (approved):
LGTM! 👍🏼  Really nice work!

monicawheeler (approved):

PR #1077, opened 2021-12-06
[SAGE-1059]: Referrer bug fix and cleanup

pixelflips:
## Description

- Updates incorrect attribute as described in #1059 
- In addition, this PR cleans up and removes the usage of `rel="noreferrer"` and replaces/updates all instances to use `rel="noopener"`

## Screenshots
No visual updates with these adjustments

## Testing in `sage-lib`
Navigate to docs site http://localhost:4000/pages/index
Check attributes are now correct from list in #1059 
Verify any links with `target="_blank"` applied that the `rel` attribute is set to `noopener` and not `noreferrer`.

## Testing in `kajabi-products`
1. (**LOW**) Corrects and updates rel attributes to use `noopener` on links w/ `target=_blank`. No expected effects within kajabi-products.
  
## Related
Closes #1059

PR #1078, opened 2021-12-07
Next version bump ~v4.31.0

pixelflips:
1. (LOW) - kajabi/sage-lib#1077 Corrects and updates rel attributes to use noopener on links w/ target=_blank. No expected effects within kajabi-products.


monicawheeler (approved):
🎉

PR #1082, opened 2021-12-09
Panel divider patch

philschanely:
## Description

Quick patch for width of full-bleed panel dividers to ensure true full bleed occurs :)

## Testing in `sage-lib`

See Components > Panel and validate that dividers appear as expected.

1. (**LOW/MEDIUM/HIGH/BREAKING**) Tweak to ensure panel dividers go full width.


anechol (approved):
Looks good on desktop and mobile

PR #1088, opened 2021-12-13
feat(progress): Add optional color prop

joedakroub:
## Description
Adds a custom color prop to change the color of the progress bar value element.

## Screenshots
After

## Testing in `sage-lib`
1. Visit http://0.0.0.0:4000/pages/component/progress_bar
2. Ensure that the default color (sage_300) and custom color option (orange_300) appear

1. (**LOW/MEDIUM/HIGH/BREAKING**) This is an optional prop and should not affect existing implementations, but always good to check.

### DIFF
--- a/docs/app/views/examples/components/progress_bar/_props.html.erb
+++ b/docs/app/views/examples/components/progress_bar/_props.html.erb
@@ -1,6 +1,6 @@
 <tr>
   <td><%= md('`color`') %></td>
-  <td><%= md('Optional hex value for the progess bar value') %></td>
+  <td><%= md('Optional hex or `SageTokens::COLOR_PALETTE` value for the progress bar color.') %></td>
   <td><%= md('String') %></td>
   <td><%= md('`nil`') %></td>
 </tr>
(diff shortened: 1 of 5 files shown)


philschanely (approved):
Love the simplicity of this approach. Allows what we need but also open to future needs. 👍

pixelflips (approved):
LGTM! 👍🏼  Nice work!

docs/app/views/examples/components/progress_bar/_props.html.erb
```diff
@@ -1,7 +1,7 @@
 <tr>
-  <td><%= md('`percent`') %></td>
-  <td><%= md('Sets `value`, `aria-valuenow` and inline `width` style of progress bar.') %></td>
-  <td><%= md('Integer') %></td>
+  <td><%= md('`color`') %></td>
+  <td><%= md('Optional hex value for the progess bar value') %></td>
```
philschanely, line 3: Perhaps add a note regarding using the `SageTokens::COLOR_PALETTE` token?

joedakroub: That's a good idea, will do!

PR #1110, opened 2021-12-16
[SAGE-1109] SageAvatar ID on image_tag

pixelflips:
## Description
Adjusts the `SageAvatar` `img_tag` to allow for an optional `ID` to be set on the image element.

## Screenshots
Screen Shot 2021-12-15 at 4 08 17 PM

## Testing in `sage-lib`

RAILS:
- Navigate to [http://localhost:4000/pages/component/avatar](http://localhost:4000/pages/component/avatar)
- Inspect the avatar in the Sample Avatars section that has the photo and verify the image has an id applied.
- Additionally, open `docs/app/views/examples/components/avatar/_preview.html.erb` update the ID and make sure changes are reflected.

REACT/STORYBOOK:
- Navigate to http://localhost:4100/?path=/story/sage-avatar--default
- Add image src, alt, and id. Make sure changes are applied and appear as expected

## Testing in `kajabi-products`
1. (**LOW**) Adds an optional ID to the SageAvatar Image element. Not expected to affect KP.

## Related
Closes #1109

PR #1108, opened 2021-12-15
[Sage-1107] Stat Box - Fixes Default Props

anechol:
## Description

Fixes default props for icon, image, and link properties. This should address the issue of being forced to manually set those props to `null` to fix alignment issues based on the prop's existence.

## Screenshots

|  Before  |  After  |
|--------|--------|
Screen Shot 2021-12-15 at 5 08 09 PM|Screen Shot 2021-12-15 at 5 08 45 PM

## Testing in `sage-lib`

- Check that props `icon`, `image`, and `link` no longer have to be manually set to `null` for alignment issues to be fixed.

## Testing in `kajabi-products`

(LOW) UI bug. Fixes default props for icon, image, and link properties. This should address the issue of being forced to manually set those props to `null` to fix alignment issues based on the prop's existence.

## Related

Closes #1107

PR #1106, opened 2021-12-15
[Sage-618] Panel Controls - Spacing Overlap Bug

anechol:
## Description

Fixes issue with Panel Controls spacing overlapping on mobile. A small patch will be applied to the next bump to update override styling for the one instance of Panel Controls in Kajabi Products.

## Screenshots

|  Before  |  After  |
|--------|--------|
Screen Shot 2021-12-15 at 3 36 47 PM|Screen Shot 2021-12-15 at 3 27 45 PM

## Testing in `sage-lib`

- Check that Panel Controls elements do not overlap on mobile.

## Testing in `kajabi-products`

(LOW) Fixes issue with Panel Controls spacing overlapping on mobile. A small patch will be applied to the next bump to update override styling for the one instance of Panel Controls in Kajabi Products.

## Related

Closes #618

PR #1112, opened 2021-12-16
Next version bump ~v4.33.0

anechol:
1. (**LOW**) - kajabi/sage-lib#1106 - Fixes issue with Panel Controls spacing overlapping on mobile. A small patch will be applied to the next bump to update override styling for the one instance of Panel Controls in Kajabi Products.
2. (**LOW**) - kajabi/sage-lib#1108 - UI bug. Fixes default props for icon, image, and link properties. This should address the issue of being forced to manually set those props to `null` to fix alignment issues based on the prop's existence.
3. (**LOW**) - kajabi/sage-lib#1110 - Adds an optional ID to the SageAvatar Image element. Not expected to affect KP.

### DIFF
--- a/docs/app/views/examples/components/avatar/_preview.html.erb
+++ b/docs/app/views/examples/components/avatar/_preview.html.erb
@@ -17,7 +17,8 @@
   <%= sage_component SageAvatar, {
     image: {
       alt: "Court's profile image",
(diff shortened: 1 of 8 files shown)


monicawheeler (approved):
🎉

PR #1116, opened 2021-12-16
[Sage-674] Form Section - Subtitle Refactor

### LINKED ISSUE #674: [Improvement] Form Section `subtitle` needs consideration
## Description

Form Sections allow a `subtitle` that is built as a single paragraph, the incoming copy populating that paragraph. However, cases exist where multiple paragraphs are needed in this space. Consider adding a `content_for` block affordance that would allow cleaner content provided here.

NOTE: This component could also be updated to make use of Grid, GridCol, and Panel components.

## Design

See [Form Section](url) for type specs and spacing.

## Additional Information
Use case sample:

Screen Shot 2021-07-30 at 4 40 18 PM

anechol:
## Description

Adds a `content_for` affordance for the Form Section `subtitle` prop. Update allows existing subtitles to remain but allows flexibility for longer content within `content_for` if needed.

## Screenshots

N/A

### Rails
- View [Rails component](http://localhost:4000/pages/component/form_section)
- Check that new `content_for` affordance works as expected.

### React
- View [Storybook](http://localhost:4100/?path=/docs/sage-formsection--default)
- Check that `subtitle` refactor works as expected (able to use markup within the prop)

(LOW) Adds a `content_for` affordance for the Form Section `subtitle` prop. Update allows existing subtitles to remain but allows flexibility for longer content within `content_for` if needed. Existing instances of the `subtitle` prop should have no changes.

## Related

Closes #674

### DIFF
--- a/docs/app/views/examples/components/form_section/_preview.html.erb
+++ b/docs/app/views/examples/components/form_section/_preview.html.erb
@@ -12,6 +12,9 @@
         desc: "This component is designed to be very flexible and take any sort of content you want to pass to it.",
         icon_name: "sage-icon-info-circle",
       } %>
+    <%= content_for :sage_form_section_subtitle do %>
+      <div>Testing subtitle content</div>
+    <% end %>
   <% end %>
 
   <!-- Custom title_tag -->
--- a/docs/app/views/examples/components/form_section/_preview.html.erb
+++ b/docs/app/views/examples/components/form_section/_preview.html.erb
@@ -40,8 +40,8 @@
       icon_name: "sage-icon-info-circle",
     } %>
     <%= content_for :sage_form_section_subtitle do %>
-      <p>Helpful text that lets the customer have an idea how to work with the section.</p>
-      <p>Additional helpful text that lets the customer have an idea how to work with the section.</p>
+      <p class="<%= "#{SageClassnames::SPACERS::XS_BOTTOM} #{SageClassnames::TYPE::BODY}"%>">Helpful text that lets the customer have an idea how to work with the section.</p>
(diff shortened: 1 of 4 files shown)


docs/app/views/examples/components/form_section/_preview.html.erb
```diff
@@ -28,4 +28,20 @@
...
+    title: "Form Section Title",
+  } do %>
+    <%= sage_component SageAlert, {
+      title: "Form Section Content Will Go Here",
+      desc: "This component is designed to be very flexible and take any sort of content you want to pass to it.",
+      icon_name: "sage-icon-info-circle",
+    } %>
+    <%= content_for :sage_form_section_subtitle do %>
+      <p>Helpful text that lets the customer have an idea how to work with the section.</p>
+      <p>Additional helpful text that lets the customer have an idea how to work with the section.</p>
```
pixelflips, line 44: Not a blocker at all, but wondering if maybe we should surround these with a `sage-type` class or maybe add a spacer to the first `p` element? Initially, when looking at the example on the docs page I didn't realize those were two separate paragraphs.

pixelflips (approved):
One minor nit-pick, but overall this is great and is a much-needed update. Nice work! LGTM! 👍🏼

docs/app/views/examples/components/form_section/_preview.html.erb
monicawheeler:
```suggestion
      <p class="<%= "#{SageClassnames::SPACERS::XS_BOTTOM} #{SageClassnames::TYPE::BODY}" %>">Helpful text that lets the customer have an idea how to work with the section.</p>
      <p class="<%= SageClassnames::TYPE::BODY %>">Additional helpful text that lets the customer have an idea how to work with the section.</p>
```

monicawheeler (approved):

PR #1117, opened 2021-12-17
Removes duplicate `subtle` key

prsimp:
## Description
I was seeing this message while working on some specs:
```
<PATH>/sage_rails/app/views/sage_components/_sage_stat_box.html.erb:75: 
warning: key :subtle is duplicated and overwritten on line 78
```

This removes the additional `subtle` key. 
## Screenshots
N/A

## Testing in `sage-lib`
A little unclear on this but I'd check to verify that the StatBox component still renders with the default `subtle` value.

Verify that the StatBox component still renders with the default `subtle` value.

1. (**LOW**) Removes a duplicated key from an HTML element. No QA changes expected since both instances of the attribute were set to `true`

## Related
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Do you have an issue, bug, or suggestion you would like to add to the Design Sys
 ### Conventional Commits
 
 This repository utilizes the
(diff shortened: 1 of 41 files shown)


pixelflips (approved):
LGTM! Nice catch, thanks for updating!

monkeypox8 (approved):
👍🏼 Aw, no love for double secret subtle? Thanks Paul!

philschanely (approved):

monicawheeler (approved):

PR #1133, opened 2022-01-05
[Sage-1008] Table - New Prop, Caption Side

### LINKED ISSUE #1008: [Improvement] Table caption alignment properties
## Description
The Sage Table currently allows use of `caption` for an overall description, but our CSS reboot defaults `caption` placement to the bottom of each table.

Providing the option to align the `caption` content (using `caption-side`) will unlock flexibility for table layouts while nudging developers to use proper semantics.

## Additional Information
- https://developer.mozilla.org/en-US/docs/Web/CSS/caption-side
- https://www.w3.org/WAI/tutorials/tables/caption-summary/

anechol:
## Description

Adds new prop `caption_side` to coincide with caption placement via CSS property `caption-side`.

## Screenshots

Screen Shot 2022-01-06 at 11 03 44 AM

Screen Shot 2022-01-06 at 11 04 51 AM

### Rails
- View [Rails component](http://localhost:4000/pages/component/table)
- Check that caption can be positioned at the top with new prop

### React
- View [Storybook](http://localhost:4100/?path=/docs/sage-table--default)
- Check that new caption prop exists and can be placed either on top or bottom of table

(LOW) Adds new prop `caption_side` to coincide with caption placement via CSS property `caption-side`. Also adds missing caption prop to React component. No effect on existing Kajabi Products work.

## Related

Closes #1008

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_table.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_table.html.erb
@@ -21,7 +21,11 @@ is_responsive = component.responsive.present? ? component.responsive : true
     <%= component.generated_html_attributes.html_safe %>
   >
     <% if component.caption.present? %>
-      <caption><%= component.caption.html_safe %></caption>
+      <caption class="
+        <%= "sage-table__caption--#{component.caption_side}" %>
+      ">
+        <%= component.caption.html_safe %>
+      </caption>
     <% end %>
     <% if component.headers.present? %>
       <thead>
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_table.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_table.html.erb
@@ -22,7 +22,8 @@ is_responsive = component.responsive.present? ? component.responsive : true
   >
     <% if component.caption.present? %>
       <caption class="
-        <%= "sage-table__caption--#{component.caption_side}" %>
+        sage-table__caption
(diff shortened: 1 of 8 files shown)


monicawheeler (approved):
Looks great 👍🏻

docs/lib/sage_rails/app/views/sage_components/_sage_table.html.erb
```diff
@@ -21,7 +21,11 @@ is_responsive = component.responsive.present? ? component.responsive : true
     <%= component.generated_html_attributes.html_safe %>
   >
     <% if component.caption.present? %>
-      <caption><%= component.caption.html_safe %></caption>
+      <caption class="
+        <%= "sage-table__caption--#{component.caption_side}" %>
```
monkeypox8, line 25: One suggestion here as we're rendering an incomplete class `sage-table__caption--` if no value is set, such as in the striped table example. Would it be helpful to default this to `.sage-table__caption--bottom` (or alternatively, only set `sage-table__caption--top` when needed)?

monkeypox8 (approved):
👍🏼 LGTM, nice work! I've one suggestion around rendering the caption class to consider.

QuintonJason (approved):

docs/lib/sage_rails/app/views/sage_components/_sage_table.html.erb
anechol: Ahh good catch! I'll make the adjustment

PR #1141, opened 2022-01-10
Next version bump ~v4.35.0

pixelflips:
1. (**LOW**) - kajabi/sage-lib#1140 Homepage image optimization. This is a docs only update and should not affect the kajabi-product repo in any way.
2. (**LOW**) - kajabi/sage-lib#1144 Adds new prop for custom content slot in Checkbox component. No effect on existing Kajabi Products work.
3. (**LOW**) - kajabi/sage-lib#1133 Adds new prop `caption_side` to coincide with caption placement via CSS property `caption-side`. Also adds missing caption prop to React component. No effect on existing Kajabi Products work.


monicawheeler (approved):
🎉

Issue #1147, opened 2022-01-11
[Mock]: Product Selection


pixelflips:
[Figma](https://www.figma.com/file/aYygNca9JqBoQs1VwwFyRd/Product-Creation-Flow?node-id=2740%3A15652)

![Screen Shot 2021-12-13 at 4 38 52 PM](https://user-images.githubusercontent.com/17955295/145893001-ca7103d5-6d18-4591-900c-53b168dc88e4.png)

philschanely:
Now in Jira https://github.com/Kajabi/sage-lib/issues/1147

PR #1148, opened 2022-01-11
[SAGE_1147] Creation Wizard Product Selection

pixelflips:
## Description
Adds the creation wizard to the mocks page and the product selection portion.

## Screenshots
Screen Shot 2022-01-12 at 9 03 02 AM

## Testing in `sage-lib`

- Navigate to http://localhost:4000/pages/mocks
- Click on the creation wizard option
- Verify product selection looks and works as expected.

N/A

## Related
https://github.com/Kajabi/sage-lib/issues/1147


monicawheeler (approved):

docs/app/views/mocks/creation_wizard/_product_selection.html.erb
```diff
@@ -0,0 +1,34 @@
+<%= sage_component SageGridRow, {} do %>
+  <%= sage_component SageGridCol, { breakpoint: "md", size: "5" } do %>
+    <h2 class="<%= SageClassnames::TYPE::HEADING_5 %> <%= SageClassnames::SPACERS::MD_BOTTOM %>">What do you want to create?</h2>
+    <div class="sage-tabs-container">
+      <%= sage_component SageTabs, {
+        css_classes: "creation-product-tabs",
+        id: "creation-prod-tabs",
+        items: creation_product_types[:product_types],
+        style: "choice",
```
monkeypox8, line 9: This works great, though one drawback I see is that the tabs will not be focusable due to the `SageChoice` fallback to `div`s. As a follow-up, I'm wondering if we'd want to consider adding content ability to `SageTab`. Or perhaps modifying `SageChoice` with the equivalent `role="button"`, etc when content is present?

monkeypox8 (approved):
👍🏼 LGTM!  One question around tab a11y for a potential follow-up. Great work!

QuintonJason (approved):

pixelflips:
> Some notes for potential tasking/discussion:

@philschanely I don't believe any of these items should hold up this PR being merged? If possible, I'd like to get this one in so I can continue with the additional dropdowns. Let me know and I am happy to create any follow-up PRs as needed. Just let me know.

I'll also coordinate with Quinton as I know he would like to add to this mock as well.

philschanely:
@pixelflips Yes; but please add a comment in the custom styles noting that we expect some if not all of these to be incorporated into components for the final release as a heads up to any engineers who begin working on this. We should also probably duplicate this setup in React but that could be done in a follow up if you prefer so this is not holding back other mocks.

pixelflips:
> some if not all of these to be incorporated into components for the final release

@philschanely Sounds good. I'll add the notes within the stylesheet and I'll reach out regarding a follow-up for the react version. Appreciate all the direction and help!

PR #1151, opened 2022-01-12
[Sage-792] Form Input - Multiple Words for Affixes

### LINKED ISSUE #792: [BUG] SageFormInput prefix & suffix don't support strings with spaces
## Detailed description

When adding a prefix or suffix where the string contains a space (e.g. two words), the form input component misplaces the string for the data attribute and the label does not display any additional words after a space.

SageFormInput

Expect prefix or suffix to display full string.

Chrome / Firefox

## Steps to reproduce

1. Spin up docs site
2. Visit http://localhost:4000/pages/component/form_input
3. Update prefix or suffix for an input with two words in `docs/app/views/examples/components/form_input/_preview.html.erb`
4. Notice that the prefix or suffix ignores words after space.

## Screenshots
For the provided screenshot the prefix and suffix strings were set to "two words" 

Screen Shot 2021-09-13 at 4 03 18 PM

anechol:
## Description

Fixes an issue with the prefixes and suffixes only outputting the first word in a string.

The issue seems to be caused by the expectation that the data attribute value is wrapped in quotes. Writing the data attribute conditional inline doesn't allow for that:

Screen Shot 2022-01-12 at 4 12 56 PM

It's likely that any other instance of this approach to data attributes within Sage is suffering from the same bug, just unnoticeable because they haven't been used for strings longer than one word.

## Screenshots

|  Before  |  After  |
|--------|--------|
|||
Screen Shot 2022-01-12 at 4 17 36 PM|Screen Shot 2022-01-12 at 4 17 20 PM
Screen Shot 2022-01-12 at 4 22 43 PM|Screen Shot 2022-01-12 at 4 23 23 PM

### Rails
- View [Rails component](http://localhost:4000/pages/component/form_input)
- Check that inputting a prefix or suffix with multiple words in `_preview.html.erb` renders in data attribute as expected

(LOW) Fixes a bug where multiple words were unable to be used as prefixes and suffixes for form inputs.

## Related

Closes #792

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_form_input.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_form_input.html.erb
@@ -1,6 +1,3 @@
-<script>
-console.log("<%= "#{component.prefix}" %>")
-</script>
 <div
   class="sage-input
     <%= "sage-form-field--error" if component.has_error %>
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_form_input.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_form_input.html.erb
@@ -1,3 +1,6 @@
+<script>
+console.log("<%= "#{component.prefix}" %>")
+</script>
 <div
   class="sage-input
     <%= "sage-form-field--error" if component.has_error %>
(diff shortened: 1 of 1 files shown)


docs/lib/sage_rails/app/views/sage_components/_sage_form_input.html.erb
```diff
@@ -1,3 +1,6 @@
+<script>
```
QuintonJason, line 1: Is this console.log still needed?

monicawheeler (approved):
🙌

QuintonJason (approved):

PR #1153, opened 2022-01-13
[Sage-932] Docs - Adds Flag to Required Props

### LINKED ISSUE #932: [Suggestion] Update Docs to Reflect Required Props
## Description

Currently, the vast majority of components lack any indication of which props are required for the component to be rendered properly. It would be helpful for developers to know exactly which props they need at a bare minimum for the components they use.

- [ ] Add a visual indicator (TBD) next to required props in Rails documentation
- [ ] Audit all components to ensure this flag is present for required props

anechol:
## Description

Adds a required flag to all required props.

## Screenshots

|  Before  |  After  |
|--------|--------|
|||
WIP|WIP

## Testing in `sage-lib`

WIP

(LOW) Adds a required flag to all required props. Sage docs only, no effect on Kajabi Products.

## Related

Closes #932


pixelflips (approved):
LGTM! Nice addition! 👍🏼

docs/app/views/examples/components/label/_props.html.erb
```diff
@@ -79,7 +79,7 @@
   <td><%= md('`"remove"`') %></td>
 </tr>
 <tr>
-  <td><%= md('`value`') %></td>
+  <td><%= md('`value`') %><%= sage_component SageLabel, { color: "published", value: "required" } %></td>
```
QuintonJason, line 82: Can this required be moved to the `value` on line 57?

monkeypox8 (approved):
👍🏼 Nice work! Spotted a couple of components with props that could also use an update (`label` and `table` were the two that I recall off the top of my head), but barring a full, exhaustive audit I think they (and any others) can be handled as follow-ups.

monicawheeler (approved):

docs/app/views/examples/components/label/_props.html.erb
anechol: Thanks for catching that! Just updated with the correct prop.

QuintonJason (approved):

PR #1159, opened 2022-01-19
Next version bump ~v4.37.0

anechol:
1. (LOW) - kajabi/sage-lib#1153 - Adds a required flag to all required props. Sage docs only, no effect on Kajabi Products.
2. (LOW) - kajabi/sage-lib#1164 -  Allow nil as an option in dropdown item style props. No effect is expected on Kajabi Products.
3. (LOW) - kajabi/sage-lib#1163 - New Toolbar component as an alternative to PanelControls


pixelflips (approved):
LGTM! 👍🏼

philschanely (approved):

PR #1161, opened 2022-01-19
[Page Heading] add code snippets per example

QuintonJason:
## Description

- [x] rails - add code snippets per component example
- [ ] - add description text regarding each variation

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-01-19 at 3 03 01 PM|page-heading-after|

## Testing in `sage-lib`

Rails: Visit the the page heading page and verify that the code snippets appear on the `preview` tab

1. (**N/A**) Doc site update only. No affect on the app.


QuintonJason:
There's still a need to add descriptions per example. This may come in a followup, but the primary focus of this PR is to establish the pattern that will display code snippets per example

docs/app/views/examples/components/page_heading/_preview.html.erb
```diff
@@ -5,6 +5,14 @@
   spacer: { bottom: :xl },
 } %>
 
+<pre class="prettyprint">
```
monicawheeler, line 8: Could we remove the `spacer: { bottom :xl }` from the above line 5 and add the space below this line, the `pre` element?

monicawheeler:
@philschanely - added you on as reviewer as well to get your thoughts here. Thanks!

docs/app/views/examples/components/page_heading/_preview.html.erb
```diff
@@ -1,15 +1,22 @@
 <h3 class="t-sage-heading-6">Standalone Page Heading</h3>
+<p>Use a default page header for a title underneath breadcrumbs</p>
 
 <%= sage_component SagePageHeading, {
   title: "Page Title",
-  spacer: { bottom: :xl },
 } %>
 
+<pre class="prettyprint <%= "#{SageClassnames::SPACERS::XL_BOTTOM} "%> ">
```
philschanely, line 8: Cool idea we've been talking about for a while! 🙌 Can this be a partial instead though? Something like:

```erb
<%= render "page_heading/snippets/_standalone.html.erb" %>
```

I'd also suggest we look at a more systematic/streamlined data-driven approach before rolling this out everywhere... save a lot of repeat code if we can be strategic.

docs/app/views/examples/components/page_heading/snippets/_pageheading_action_buttons.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= sage_component SagePageHeading, {
```
philschanely, line 1: ❤️ I'd love to see us also pull the documentation notes into a central `.md` or other similar file. I believe the whole folder example folder could be refactored to run off of a central data source and significantly slim down the repeat code we have currently.

philschanely (commented):
Sweet idea! Totally on board with the idea but would like to debrief on a strategy for the whole picture before we roll forward with this as-is.

philschanely:
FWIW some Ideas to discuss for further refinement here of in future for a strategic systematic approach:

- consolidate the code snippet into a partial
- use a central data file for running this to further trim down on repeat code and streamline documentation process
- use tabs for each example so the page isn't overwhelmed with code snippets?
- add a copy code button/affordance?

docs/app/views/examples/components/page_heading/_preview.html.erb
QuintonJason: agreed, this is the pr to establish the pattern so i'm not rushing it

PR #1162, opened 2022-01-20
[Sage-94] Choice - Disabled Prop Fix

anechol:
## Description

Updates logic for handling the disabled prop. Adds `aria-disabled` to the div variation of this component as well as a missing `tabindex` for keyboard navigation.

This direction was taken over a docs update to improve accessibility and leave the heavy-lifting for a11y within Sage.

## Screenshots

Screen Shot 2022-01-20 at 4 29 28 PM|Screen Shot 2022-01-20 at 4 29 34 PM

### Rails
- View [Rails component](http://localhost:4000/pages/component/choice)
- Add content to any preview component.
- Check that `aria-disabled` is properly added to the div variation of the component.
- Check that `tabindex` also exists for the div variation.

### React
- View [Storybook](http://localhost:4100/?path=/docs/sage-choice--default)
- Add custom content in the `children` prop.
- Check that `aria-disabled` is properly added to the div variation of the component.
- Check that `disabled` prop does not exist on the div variation.
- Check that `tabindex` also exists for the div variation.

(LOW) Updates logic for handling the disabled prop. Adds `aria-disabled` to the div variation of this component as well as a missing `tabindex` for keyboard navigation.

## Related

[Sage-94](https://kajabi.atlassian.net/browse/SAGE-94)

### DIFF
--- a/packages/sage-react/lib/Choice/Choice.jsx
+++ b/packages/sage-react/lib/Choice/Choice.jsx
@@ -62,7 +62,7 @@ export const Choice = ({
     'data-js-tabs-target': target,
     disabled: TagName !== 'div' && disabled,
     tabIndex: TagName === 'div' ? '0' : null,
-    role: "tab",
+    role: 'tab',
     ...(hasRadioConfigs && { htmlFor: radioConfigs.id }),
     ...(isButton && { type: 'button' }),
     ...rest,
--- a/packages/sage-react/lib/Choice/Choice.jsx
+++ b/packages/sage-react/lib/Choice/Choice.jsx
@@ -62,6 +62,7 @@ export const Choice = ({
     'data-js-tabs-target': target,
     disabled: TagName !== 'div' && disabled,
     tabIndex: TagName === 'div' ? '0' : null,
+    role: "tab",
     ...(hasRadioConfigs && { htmlFor: radioConfigs.id }),
     ...(isButton && { type: 'button' }),
     ...rest,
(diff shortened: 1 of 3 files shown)


packages/sage-react/lib/Choice/Choice.jsx
```diff
@@ -56,10 +56,13 @@ export const Choice = ({
 
   const attrs = {
     'aria-controls': target,
+    'aria-disabled': TagName === 'div' ? disabled : null,
```
anechol, line 59: Would love some feedback here if there's a better/cleaner way to write this logic.

packages/sage-react/lib/Choice/Choice.jsx
```diff
@@ -56,10 +56,13 @@ export const Choice = ({
...
     'aria-selected': isActive,
     className: classNames,
     'data-js-tabs-target': target,
-    disabled,
+    disabled: TagName !== 'div' && disabled,
```
anechol, line 63: Same here.

joedakroub: Looks good to me. I know sometimes these conditions don't always feel good to write, but they do the job. 😄

monicawheeler (approved):

joedakroub (approved):
Looking great Ashley! Really nice additions too. 👍🏼

packages/sage-react/lib/Choice/Choice.jsx
anechol: Thanks a lot!

PR #1163, opened 2022-01-21
Add Toolbar as alternate to Panel Controls

philschanely:
## Description

This is an MVP effort to create a minimal alternative to `PanelControls`. Here we allow for `Toolbar` to wrap a row of controls to provide desired border styling, and particularly to group/merge borders with `ToolbarGroup`. Yet, this leaves all functionality for the controls up to the engineer to configure as desired.

Note this is a first step towards deprecating Panel Controls. Once this feature is mature we can further investigate how to replace uses of Panel Controls with this component. WIP for React uses within the Manage team (only current React uses ATM) are here: https://github.com/Kajabi/kajabi-products/pull/22070

## Screenshots

Screen Shot 2022-01-21 at 11 41 09 AM

## Testing in `sage-lib`

See Components > [Toolbar](http://localhost:4000/pages/component/toolbar) and the same in [Storybook](http://localhost:4100/?path=/story/sage-toolbar--default)

1. (**LOW**) New Toolbar component as an alternative to PanelControls


pixelflips (commented):
This is great and a huge step toward deprecating the panel controls. A couple of minor things I noticed and are mostly related to the SageToolbarGroup.

1. It may not be a show stopper at this point, but I notice the last two items get the borders doubled up.

Screen Shot 2022-01-21 at 9 59 19 AM

2. The SageToolbarGroup options don't wrap so well on small screens. This could also be addressed in a follow-up, but just thought I'd mention it.

Screen Shot 2022-01-21 at 10 11 22 AM

philschanely:
@pixelflips thanks for the comments! These both appear to be problems on PanelControls now so I might suggest we defer those as fixes in a follow up. My chief concern is that we not introduce a new problem with this effort if that makes sense to you too!

philschanely:
@pixelflips FYI https://kajabi.atlassian.net/browse/SAGE-150

pixelflips (approved):
LGTM!

PR #653, opened 2021-07-23
[Catalog Item]  update action button alignment when title wraps

QuintonJason:
## Description

- [x] update action button alignment when title wraps

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2021-07-23 at 2 14 13 PM|Screen Shot 2021-07-23 at 2 14 06 PM|

## Testing in `sage-lib`

Visit the catalog item page. (You may need to update the text to wrap)
Verify that the action buttons don't stretch to match the height of the wrapped title

## Testing in `kajabi-products`

1. (LOW)  Description of the change and its impact with QA as the audience.
   - [ ] Product Outlines view

PR #1171, opened 2022-01-24
[SAGE-157] Page Heading Actions Alignment

pixelflips:
## Description
Within the page heading component, buttons do not align vertically when using a subtle button alongside a standard button with the `:sage_page_heading_actions` content area.

Based on the new page heading design and incoming tabs, these will need to align vertically.

Also, see https://github.com/Kajabi/sage-lib/pull/653 for why a child div element was added.

Screen Shot 2022-01-24 at 12 38 53 PM

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-01-24 at 1 11 46 PM|Screen Shot 2022-01-24 at 1 11 17 PM|

## Testing in `sage-lib`

Navigate to [page heading component](http://localhost:4000/pages/component/page_heading)
In the last example at the bottom of the page, verify the preview icon button and full-size button align. (see screenshot above)

1. (**LOW**) Adjusts alignment of the subtle button and full-size buttons when used within `:sage_page_heading_actions` area. Quick sanity of page headings within kajabi-products should be performed.

## Related
https://kajabi.atlassian.net/browse/SAGE-157

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_page_heading.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_page_heading.scss
@@ -62,6 +62,7 @@ $-page-heading-image-height-mobile: rem(120px);
 .sage-page-heading__title {
   @extend %t-sage-heading-1;
   grid-area: title;
+  margin-right: sage-spacing(sm);
 }
 
 .sage-page-heading__intro {
@@ -111,7 +112,7 @@ $-page-heading-image-height-mobile: rem(120px);
 .sage-page-heading__actions {
   display: flex;
   grid-area: actions;
-  align-items: center;
+  align-items: flex-start;
 
   @media (max-width: sage-breakpoint(sm-max)) {
     flex-wrap: wrap;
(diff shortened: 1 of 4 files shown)


packages/sage-assets/lib/stylesheets/components/_page_heading.scss
```diff
@@ -111,7 +112,7 @@ $-page-heading-image-height-mobile: rem(120px);
 .sage-page-heading__actions {
   display: flex;
   grid-area: actions;
-  align-items: center;
+  align-items: flex-start;
```
monkeypox8, line 115: Something to consider here with the new wrapper in place, the flex display and the line here are no longer needed from what I can tell. Testing with the bridge appeared to confirm that we may want to consider removing those and instead apply the flex container only to `.sage-page-heading__actions-inner` for cases of multi-button alignment.

monkeypox8 (approved):
👍🏼 LGTM! One suggestion to consider, not a blocker. Great work!

packages/sage-assets/lib/stylesheets/components/_page_heading.scss
pixelflips: @monkeypox8 good call. I've adjusted so when you have a sec, please let me know that the new change is what you had in mind?

QuintonJason (approved):

PR #1172, opened 2022-01-24
[SAGE-161] Optional Tab Icon

pixelflips:
## Description
With the new design updates that will incorporate tabs into the page heading component, the tab component will need to be able to support an optional icon to be displayed next to the tab text.

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-01-24 at 3 22 28 PM|Screen Shot 2022-01-24 at 3 22 21 PM|

## Testing in `sage-lib`
Navigate to tab: http://localhost:4000/pages/component/tab
Verify updated preview displays example of tab with icon.

Navigate to tabs: http://localhost:4000/pages/component/tabs
Verify updated preview displays an example of multiple tabs with icons.

1. (**LOW**) Adds an optional icon to the tab component. No effect on kajabi-products expected.

## Related
https://kajabi.atlassian.net/browse/SAGE-161

### DIFF
--- a/docs/app/views/examples/components/tab/_preview.html.erb
+++ b/docs/app/views/examples/components/tab/_preview.html.erb
@@ -2,10 +2,22 @@
 <%= sage_component SagePanelBlock, {} do %>
   <%= sage_component SageTab, {
       target: "example-tab1",
-      text: "Page 1"
+      text: "Page 1",
     }
   %>
 <% end %>
+
+<h3 class="t-sage-heading-6">As a button and icon with a target pane in mind</h3>
+<%= sage_component SagePanelBlock, {} do %>
+  <%= sage_component SageTab, {
+      target: "example-tab2",
+      text: "Page 1",
+      icon: "pen"
+    }
+  %>
+<% end %>
+
+
 
 <h3 class="t-sage-heading-6">As a link with a route or site in mind</h3>
 <%= sage_component SagePanelBlock, {} do %>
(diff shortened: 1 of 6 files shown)


monkeypox8 (approved):
👍🏼 🔥 LGTM!

QuintonJason:
@pixelflips are you able to update react side as well or is this out of scope for this task?

pixelflips:
> @pixelflips are you able to update react side as well or is this out of scope for this task?

@QuintonJason - I did some digging and paired up with @joedakroub and while it's not exactly obvious with the current setup, the react side supports icons already. We could/should probably refactor at some point to get things in parody but if you pass in `tabChoiceIcon: "pen"` (naming could be cleaned up) to a tab, the icon appears.

For now, I think we're good to go on that side besides noting somewhere to take another look at tabs in react as a future follow-up, cool?

QuintonJason:
> > @pixelflips are you able to update react side as well or is this out of scope for this task?
> 
> @QuintonJason - I did some digging and paired up with @joedakroub and while it's not exactly obvious with the current setup, the react side supports icons already. We could/should probably refactor at some point to get things in parody but if you pass in `tabChoiceIcon: "pen"` (naming could be cleaned up) to a tab, the icon appears.
> 
> For now, I think we're good to go on that side besides noting somewhere to take another look at tabs in react as a future follow-up, cool?

Created [SAGE-170](https://kajabi.atlassian.net/browse/SAGE-170) to track this issue

QuintonJason (approved):

PR #1178, opened 2022-01-27
feat(typography): Body font selector scoping

joedakroub:
## Description
This PR updates the body selector in `packages/sage-assets/lib/stylesheets/core/_typography.scss` to only apply if the `sage-excluded` class is not set on `body`. Then when styles are desired in a scoped context the `.sage-scoped` class can be used on the closer container.

This will prevent the `Inter` font from applying to the body which is helpful when using Sage on the member side or other environments. It also ensures any global tag selectors are moved and reconciled in the global reboot file so that core files can be included in a project independently as needed.

We are not documenting this on the docs site as this is not a fully robust solution that we want to publicize yet. Build team is using it in a limited context for the Quizzes redesign effort underway.

## Screenshots
|  Class Applied  |  Class Not Applied  |
|--------|--------|
|Class Applied|Class Not Applied|

## Testing in `sage-lib`
1. Open http://0.0.0.0:4000/
2. Ensure that the body font is still `Inter` or `Inter var`.

1. In Admin, open the developer tools
2. Inspect the body element
3. Add the `sage-excluded` class to the body element and ensure that `Inter var` and/or `Inter` does not get applied. Add `sage-scoped` at a lower level container to see Sage type apply therein.

## Notes
(LOW) Adds some MVP scoping options for Sage to be used in scoped member-facing contexts.
_NOTE: Sage Engineering has reviewed this work in a number of `kp` contexts with the Bridge on and found no regressions._

## Related
https://kajabi.atlassian.net/browse/SAGE-156


packages/sage-assets/lib/stylesheets/core/_typography.scss
```diff
@@ -19,20 +18,6 @@ $-headings-margin-bottom: sage-spacing(sm);
   }
 }
```
philschanely, line 20: The global styles for `a`, etc. lower down in this file present another reset/reboot problem we should scope, right? What if they were moved up into revised `body` selector and we addd `:not([class])` after each tag? So I'm thinking the output selector would look like:

```css
body:not([class*="sage-body--disable-font"]) a:not([class]),
.sage-body-font a:not([class]) {
  ...
}
```

Does that make sense to you too?

packages/sage-assets/lib/stylesheets/core/_system-setup.scss
```diff
@@ -0,0 +1,17 @@
+// Font definitions
+$-body-font-stack: "Inter", -apple-system, "BlinkMacSystemFont", "Segoe UI", "Roboto", "Helvetica Neue", "Arial", "Noto Sans", sans-serif;
+
+
+body:not([class*="sage-body--disable-font"]) {
```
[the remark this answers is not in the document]
philschanely: `.sage-scoped` and `.not-sage-scoped`

JulaB (commented):
👍 This looks great! I was able to remove all hardcoded typography from the quizzes styles (https://github.com/Kajabi/kajabi-products/pull/22306/files#diff-e5146c6cc17d4036759eda203cc3612c8d1777d27b049fe0aae41cd4a0ce2274L27-L196)

I checked it in different themes. Here are small issues that I found:
- theme's header font takes over the Sage font
Screen Shot 2022-02-08 at 9 19 10 PM
Screen Shot 2022-02-08 at 4 51 33 PM

- line-height issue for themes with percentage root font-size
Screen Shot 2022-02-08 at 9 43 50 PM

Note: Here are some css properties that I needed to hardcode: https://github.com/Kajabi/kajabi-products/pull/22306/files#diff-e5146c6cc17d4036759eda203cc3612c8d1777d27b049fe0aae41cd4a0ce2274R90-R148

https://github.com/Kajabi/kajabi-products/pull/22306/files#diff-e5146c6cc17d4036759eda203cc3612c8d1777d27b049fe0aae41cd4a0ce2274R40-R48

joedakroub:
@JulaB Have you tried adding the `sage-scoped` class to or around the modal to see if that helps prevent the heading font from overriding Sage?

JulaB:
> Have you tried adding the sage-scoped class to or around the modal to see if that helps prevent the heading font from overriding Sage?

Yes, I added it here https://github.com/Kajabi/kajabi-products/pull/22306/files#diff-0e158ff96225a99f22477edde3c3ff8329e3126236182794195c72c22b71fcd0R20 and in other places.

packages/sage-assets/lib/stylesheets/core/_typography.scss
```diff
@@ -11,25 +11,77 @@ $-body-margin-bottom: sage-spacing(xs);
 $-headings-margin-bottom: sage-spacing(sm);
 
 
-// Automate setting up classes that extend the specs
-@each $spec-name, $spec in $sage-type-specs {
-  .t-sage-#{$spec-name} {
-    @extend %t-sage-#{$spec-name};
-    color: inherit;
-  }
+body:not([class~="sage-excluded"]) {
```
monkeypox8, line 14: I'm wondering if we'd prefer to lock these selectors down to prevent a partial match, however unlikely that may be:

```
body:not(.sage-excluded)
```

monkeypox8 (approved):
👍🏼 LGTM! Working great with the bridge on.

philschanely (approved):

PR #1184, opened 2022-01-28
[SAGE-189] Fix(creation wizard mock): scopes payment dropdown JS to page

monkeypox8:
## Description
Reduces JS scope to Creation Wizard mock page only. The script is currently running on all pages, locking up the remainder of the JS and blocking rendering in some cases.

console

Follow-ups to consider:
- separate namespace for Sage mock JS modules
- router to conditionally init mock JS
- apply a custom class to mock pages for scoping (rather than individual components)

## Testing in `sage-lib`
1. View the Sage docs site
2. Navigate to any page and inspect using your browser dev tools
3. Confirm that no errors are displayed in the console.

1. (N/A) Creation Wizard Payment Dropdown: Affects Sage docs site only.

### DIFF
--- a/docs/lib/sage-frontend/javascript/docs/mocks/creation_wizard/index.js
+++ b/docs/lib/sage-frontend/javascript/docs/mocks/creation_wizard/index.js
@@ -1,41 +1,42 @@
 // ==================================================
 // creation wizard payment dropdown
 // ==================================================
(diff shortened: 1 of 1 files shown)


pixelflips (approved):
LGTM! 👍🏼  - Thanks for catching that!

philschanely (approved):

PR #1186, opened 2022-01-31
[Modal] - add z-index to fullscreen modal header

QuintonJason:
## Description

- [x] add `z-index`

## Screenshots

|  Before  |  After  |
|--------|--------|
|fullscreenModalBefore|fullscreenModalAfter|

## Testing in `sage-lib`

On the fullscreen modal variation, scroll overflowed body text to verify that the content falls behind the header

1. (**MEDIUM**) Description of the change and its impact with QA as the audience.
   - [ ] Product Product Creation Wizard. (behind FF: `:product_creation_wizard`)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_modal.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_modal.scss
@@ -146,7 +146,7 @@ $-modal-fullscreen-top-spacing: rem(104px);
 
   .sage-modal--fullscreen & {
     position: fixed;
-    z-index: sage-z-indexes(default, 1);
+    z-index: sage-z-index(default, 1);
     left: 0;
     right: 0;
     margin: 0;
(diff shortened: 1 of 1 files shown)


pixelflips (approved):
LGTM!

packages/sage-assets/lib/stylesheets/components/_modal.scss
```diff
@@ -146,6 +146,7 @@ $-modal-fullscreen-top-spacing: rem(104px);
 
   .sage-modal--fullscreen & {
     position: fixed;
+    z-index: sage-z-indexes(default, 1);
```
monkeypox8, line 149: Should this be `sage-z-index`?

monkeypox8 (changes requested):
Small typo in the token use, should be good to go once that's cleared up though!

Something that we might want to look into later is applying `overscroll-behavior` on modals as default behavior. No Safari support yet, but it could be useful to lock the background page from scrolling.

packages/sage-assets/lib/stylesheets/components/_modal.scss
QuintonJason: @monkeypox8 nice catch. Pushed a fix

monkeypox8 (approved):
👍🏼 LGTM!

PR #1187, opened 2022-01-31
[SAGE-184] Hero layout and slots update

QuintonJason:
## Description

- [x] add button slot
- [x] provide guide to update breaking change on `kp`
- [x] rework hero header
- [x] provide guide to update breaking change on `kp`
- [x] adjust layout
- [ ] add alignment options
- [x] create React version

### Breaking changes
- `button` property is now the `section` `hero_footer_actions` and can now take one or more than one button
    - follow the pattern set in [this commit](https://github.com/Kajabi/sage-lib/pull/1187/commits/2fde99b55a77102227d3e83ef54e3558da6bb2dd) as one pattern to follow
- Rails: renamed `alt_text` to `media_alt_text`
- Rails: renamed `size` to `hero_size`

#### Affected files for implementation
- [app/views/admin/beta_features/_hero_modal_dismiss_button.html.erb](https://github.com/Kajabi/kajabi-products/blob/871986d698eed7805fc29c68b7f2497e2831130b/app/views/admin/beta_features/_hero_modal_dismiss_button.html.erb)
- https://github.com/Kajabi/kajabi-products/blob/02dc2067b542ba3123d6d97b92a0e0d057d0f5bb/app/views/admin/pipeline_blueprints/v2/_index.html.erb

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-01-31 at 5 04 39 PM|Screen Shot 2022-01-31 at 5 04 31 PM|

## Testing in `sage-lib`

- [Rails version](http://localhost:4000/pages/component/hero)
- [React version](http://localhost:4100/?path=/story/sage-hero--default)

1. (**BREAKING**) Update property names and slots for the Hero component.
   - [ ] Beta Features / Labs
   - [ ] Product Blueprints

## Related

Closes [SAGE-184](https://kajabi.atlassian.net/browse/SAGE-184)


docs/lib/sage_rails/app/views/sage_components/_sage_hero.html.erb
```diff
@@ -13,6 +20,14 @@
     <% if component.button %>
       <%= component.button.html_safe %>
     <% end %>
+
+    <% if content_for? :sage_hero_footer_actions %>
+      <div class="sage-hero__footer-actions ">
+        <%= sage_component SageButtonGroup, { gap: :md } do %>
```
QuintonJason, line 26: Does it make sense to have the button group gap size locked down so that devs on need to worry about the actual buttons?

Even with proper documentation would folks always include the button group with buttons?

packages/sage-react/lib/Hero/Hero.jsx
```diff
@@ -0,0 +1,105 @@
...
+          {footerActions && (
+            <div className="sage-hero__footer-actions">
+              <Button.Group gap="xs">
+                {footerActions}
+              </Button.Group>
+            </div>
+          )}
+        </div>
+      </div>
+      {/* TODO - hook up ctaAttributes */}
```
philschanely, line 62: Note as well that this may need an option to be treated like a button with just an `onClick` rather than an href.

docs/lib/sage_rails/app/views/sage_components/_sage_hero.html.erb
```diff
@@ -4,15 +4,30 @@
...
+      <% if content_for? :sage_hero_footer_actions %>
+        <div class="sage-hero__footer-actions">
+          <%= sage_component SageButtonGroup, { gap: :md } do %>
+            <%= content_for :sage_hero_footer_actions %>
+          <% end %>
+        </div>
+      <% end %>
+    </div>
   </div>
   <a
```
philschanely, line 32: In React a linting rule will make using `a` and `href` hard if we actually just want an `onClick` for this. Here in Rails, some treatment like `button` is worth consider as an option for alignment between the two in the case that the attributes end up being used rather than href.

docs/lib/sage_rails/app/views/sage_components/_sage_hero.html.erb
```diff
@@ -4,15 +4,30 @@
...
+        <div class="sage-hero__branding">
+          <%= content_for :sage_hero_branding %>
+        </div>
+      <% end %>
+      <<%= html_tag %> class="sage-hero__title"><%= component.title %></<%= html_tag %>>
+    </div>
+    <div class="sage-hero__body">
+      <% if component.description %>
+        <div class="sage-hero__description">
+          <%= component.description %>
```
philschanely, line 19: Do you think this actually warrants a slot option as well? Such as in the case that multiple paragraphs are needed/

packages/sage-react/lib/Hero/Hero.jsx
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import { Button } from '../Button';
+import { HERO_SIZES } from './configs';
+
+export const Hero = ({
+  altText,
```
philschanely, line 8: Double-check the prop names align between here and Rails 🙏

docs/app/views/examples/components/hero/_preview.html.erb
```diff
@@ -1,43 +1,97 @@
...
-      },
-      attributes: {
-        "data-js-hero--dismiss": ""
-      }
-    })
-  } %>
+    title_tag: "h3"
+  } do %>
+    <% content_for :sage_hero_branding do %>
+      <%= sage_component SageCardRow, {
```
philschanely, line 30: ❤️

philschanely (commented):
Looking good! Some suggestions for final touches.

PR #1189, opened 2022-01-31
Update Dropdown for custom items and improved trigger setup

philschanely:
## Description

This PR makes the following additions to Dropdown in Rails:

- Adds a SageDropdownTrigger component
  - This sustains existing functionality with no regressions but buttresses what is possible for `select` types
  - Deprecation effort can begin to tidy up uses of the `trigger_type: "select"` and instead to use the new `trigger: { type: "select", label: "...", value: "" }` syntax and thus further reduce manual sage classes and tighten reliability.
  - Deprecation effort can begin to update existing dropdowns and place the existing triggers that are directly inside `content` slots to instead be mapped specifically to the `sage_dropdown_trigger` content slot.
  - Follow up will switch the `component.content` to instead contain panel content rather than needing the `sage_dropdown_custom_panel` slot.
- Adds a SageDropdownItem component
  - This preserves all existing functionality while strengthening the internal composition of Dropdown to use this component.
  - This also now allows for safer construction of custom dropdown items by using a manual loop through an array of items to format content inside the `content` of `SageDropdownItem`. 
  - A new `sage_dropdown_items` slot is also added to make this affordance easy to use.

### DIFF
--- a/docs/app/views/examples/components/dropdown/_preview.html.erb
+++ b/docs/app/views/examples/components/dropdown/_preview.html.erb
@@ -1,3 +1,8 @@
+<%# 
+  JAY: FFR this whole file is the full "demo" page for dropdown;
+  I'll put my desired demos below but you can see how other aspects
+  of dropdown are demonstrated here.
+%>
 <h3 class="t-sage-heading-6">Menu Button</h3>
 <%= sage_component SageDropdown, {
   items: [{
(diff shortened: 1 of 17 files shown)


docs/app/views/examples/components/dropdown/_preview.html.erb
```diff
@@ -285,3 +290,96 @@
...
+    }, {
+      value: {
+        image: "//multiple-image",
+        name: "Multi-pay",
+        description: "Require several...",
+      },
+      attributes: { "href": "#" },
+    },
+  ],
+} do %>
```
jaymcgavren, line 324: @philschanely I think I underestimated what I was getting myself into here. 😆  I see from the Git logs that you wrote both `sage_table_for` and the `SageTableFor` class (impressive!), so I don't know how much I can tell you that you don't already know...

If your question is "how can I use `content_for` in a `sage_component` block (as you're doing here, with the `sage_component` call starting on line 298), I'm afraid this is outside my expertise.

If, on the other hand, you just want to render the content from line 325-339 and make the rendered HTML available to the `SageDropdown` object, that's already been done for you on this line: https://github.com/Kajabi/sage-lib/blob/SAGE-39_pjs-demo-for-input/docs/lib/sage_rails/app/helpers/sage_component_helper.rb#L8 I'm not too confident with how the Rails `capture` method works, but it calls the block, renders its content, and makes it available on your `SageDropdown`'s `content` attribute. So if you just call `content` within any of the methods in `SageDropdown`, you should get the rendered HTML for line 325-339 (in this example) back.

Hope that's what you actually needed. 😛  Let me know if I've missed the mark!

docs/app/views/examples/components/dropdown/_props.html.erb
```diff
@@ -1,113 +1,224 @@
...
+</thead>
+<tbody>
+  <tr>
+    <td><%= md('`data-sage-dropdown-exit`') %></td>
+    <td><%= md('This custom attribute can be used within custom dropdown panels to allow elements to close the dropdown panel with clicked.') %></td>
+    <td><%= md('See example "Dropdown Menu Button with Custom Content"') %></td>
+  </tr>
+</tbody>
+
+<thead>
```
philschanely, line 132: FTR I admit this is not ideal, and yet I don't see the value in adding a new page specifically for this subcomponent. Will mark to think more about this for cases like this and other similar places.

docs/app/views/examples/components/dropdown/_props.html.erb
```diff
@@ -1,113 +1,224 @@
...
+  </tr>
+  <tr>
+    <th>Property</th>
+    <th>Description</th>
+    <th>Type</th>
+    <th>Default</th>
+  </tr>
+</thead>
+<tbody>
+  <tr>
```
philschanely, line 147: Moved schema details for dropdown items here.

monkeypox8 (approved):
👍🏼 LGTM, nice cleanup! Tested using the bridge for good measure, no concerns here.

docs/app/views/examples/components/dropdown/_props.html.erb
```diff
@@ -1,113 +1,224 @@
...
+    <td><%= md('`"small"`') %></td>
+    <td><%= md('`nil`') %></td>
+  </tr>
+  <tr>
+    <td><%= md('`search`') %></td>
+    <td><%= md('When true, adds a search component to the dropdown.') %></td>
+    <td><%= md('Boolean') %></td>
+    <td><%= md('`nil`') %></td>
+  </tr>
+  <tr>
```
philschanely, line 64: New property used to pass configs straight to a DropdownTrigger to make setting up `select` variations simpler.

philschanely:
@monkeypox8 Thanks! Hopefully this is also set up for a refactor/deprecation of the following:

- Trigger for select variant: now built in so places this is manually added can be updated.
- Trigger otherwise now can be placed in a new slot `sage_drodpdown_trigger` as a first step. After all dropdowns are adjusted to use this slot we can then look into using the default `component.content` to provide panel contents if desired.
- Cases where custom element content is sent through the items' `value` property can be refactored to use the new templating approach with `sage_dropdown_items` slot and a loop of `SageDropdownItem`s.

docs/lib/sage_rails/app/sage_components/sage_dropdown.rb
```diff
@@ -1,22 +1,7 @@
...
-    id: [:optional, NilClass, String],
-    items: [:optional, [[SageSchemas::DROPDOWN_ITEM]]],
-    panel_size: [:optional, NilClass, Set.new(["small"])],
-    panel_type: [:optional, NilClass, Set.new(["custom", "dropdown", "choice", "checkbox", "status", "searchable"])],
-    panel_width: [:optional, NilClass, String],
-    search: [:optional, NilClass, TrueClass],
-    trigger_type: [:optional, NilClass, Set.new(["select", "select-labeled"])],
-    wrap_footer: [:optional, NilClass, TrueClass],
-  })
+  set_attribute_schema(SageSchemas::DROPDOWN)
```
QuintonJason, line 2: I see that we're moving the schema to its own file, but the sections remain in this file. Is there a way to they both can be moved?

QuintonJason:
@philschanely are you able to add the react version of the new components props or is that out of scope for this task?

docs/lib/sage_rails/app/sage_components/sage_dropdown.rb
philschanely: That's a great point. We don't have a setup for that yet but I'll def add it to the list!

philschanely:
@QuintonJason React affordance added and ready for review :)

QuintonJason (approved):
Good stuff!

PR #1190, opened 2022-02-01
[Sage-194] Dropdown - Custom Panel Width

anechol:
## Description

Adds a new prop to allow a custom panel with for the Dropdown component.

## Screenshots

|  Before  |  After  |
|--------|--------|
Screen Shot 2022-02-01 at 11 20 27 AM|Screen Shot 2022-02-01 at 11 19 22 AM

## Testing in `sage-lib`

- View [Rails component](http://localhost:4000/pages/component/dropdown)
- Adds prop `panel_width` to any preview component
- Verify that prop works as intended.

## Testing in `kajabi-products`

(LOW) Adds a new prop to allow a custom panel with for the Dropdown component. No effect on existing Kajabi Products work.

## Related

[Sage-194](https://kajabi.atlassian.net/browse/SAGE-194)

PR #1191, opened 2022-02-01
Next version bump v4.43.0

anechol:
1. (LOW) - kajabi/sage-lib#1190 - Adds a new prop to allow a custom panel with for the Dropdown component. No effect on existing Kajabi Products work.


monicawheeler (approved):
🎉

PR #1202, opened 2022-02-08
[Sage-208] Link - Foundations Update

anechol:
## Description

Updates Link component with the following:
- Adds `text-decoration: underline`
- Updates focus outline
- Updates colors and type spec

## Screenshots

|  Before  |  After  |
|--------|--------|
Screen Shot 2022-02-22 at 1 04 42 PM|Screen Shot 2022-02-22 at 1 05 47 PM

### Rails
- View [Rails component](http://localhost:4000/pages/component/link)
- Check that component is properly updated to design specs.

### React
- View [Storybook](http://localhost:4100/?path=/docs/sage-link--primary)
- Check that component is properly updated to design specs.

(LOW) Visual updates for Link component.

## Related

[Sage-208](https://kajabi.atlassian.net/browse/SAGE-208)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_link.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_link.scss
@@ -16,6 +16,7 @@
 
   &:hover {
     color: sage-color(primary, 700);
+    text-decoration: underline;
   }
 
   &:focus {
@@ -25,7 +26,7 @@
   @include sage-focus-outline(
     $outline-width: 4px,
     $outline-offset-inline: 4px,
-    $outline-border-radius: 10px,
+    $outline-border-radius: sage-border(radius-input),
   );
   @include sage-focus-outline--update-color(sage-color(primary, 300));
 }
--- a/packages/sage-assets/lib/stylesheets/components/_link.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_link.scss
@@ -6,13 +6,28 @@
 
 
 .sage-link {
+  @extend %t-sage-body-med;
+
   display: inline-flex;
   position: relative;
   max-width: 100%;
   min-width: 0; /* the is needed so that truncation work when the link doesn't have an explicit width set  */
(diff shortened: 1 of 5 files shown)


packages/sage-assets/lib/stylesheets/components/_link.scss
```diff
@@ -27,6 +42,7 @@
 .sage-link--help,
 .sage-link--help-icon-only {
   position: relative;
+  text-decoration: none;
```
monicawheeler, line 45: This selector is losing on specificity and not getting applied.
Screen Shot 2022-02-22 at 2 35 38 PM

monicawheeler (changes requested):
Help icon is still receiving underline, added a note in the CSS.

anechol:
> Help icon is still receiving underline, added a note in the CSS.

@monicawheeler I'm making a minor update to prevent the specificity issue, but it looks like removing the underline from the icon only will require some interesting updates to the `display` property, namely changing both the link and the `::before` to `inline-block`. We have a ticket open to either deprecate or repurpose the help and launch variants. Given that, would it be worth making this potentially temporary change now and changing it back after a decision has been made on whether to deprecate the variant? I can definitely do this and don't see any adverse effects save for having to manually adjust icon alignment with margins. Let me know your thoughts.

monicawheeler:
@anechol - given that we have a ticket open to deal with the help and launch variants, this request can be ignored. In addition, what do you think about removing the help/info link examples in the documentation site?

monicawheeler (approved):
Updated to Approved. Non-blocking note in the comments. Thanks for your consideration and great work!

anechol:
> Updated to Approved. Non-blocking note in the comments. Thanks for your consideration and great work!

@monicawheeler Yep, I can definitely remove the examples for now.

pixelflips (approved):
LGTM! :shipit:

PR #1205, opened 2022-02-09
[SAGE-235] Dropdown width

joedakroub:
## Description
This PR replaces the `panelMaxWidth` prop with `panelWidth` in the React DropdownPanel to better match how the Rails version works.

`panelMaxWidth` is not currently used in KP nor is it in the Rails component.

This enables one to use `panelWidth: "500px"` to be able to set a custom width. The custom width does not override the `min-width` setting as doing so could allow the menu to get cut off and override the `small` panel size.

## Testing in `sage-lib`
1. Visit http://0.0.0.0:4100/?path=/story/sage-dropdown--default
2. Ensure that the `panelSize` and `panelWidth` settings work as expected

Please test any area in KP that uses React dropdowns:

1. (**MED**) Contacts
   - [ ] http://www.kajabi.test:3000/admin/sites/1/contacts
2. (**MED**) Website -> Design
   - [ ] http://www.kajabi.test:3000/admin/sites/1/website_designs

## Related
https://kajabi.atlassian.net/browse/SAGE-235

### DIFF
--- a/packages/sage-react/lib/Dropdown/DropdownPanel.jsx
+++ b/packages/sage-react/lib/Dropdown/DropdownPanel.jsx
@@ -7,10 +7,10 @@ import classnames from 'classnames';
 export const DropdownPanel = ({
   children,
   coords,
-  maxWidth,
   modifier,
   onClickScreen,
   onExit,
+  width,
 }) => {
   const menuEl = useRef(null);
   const classNames = classnames(
@@ -35,7 +35,7 @@ export const DropdownPanel = ({
       onClick={handlePanelClick}
       role="dialog"
       style={{
-        maxWidth,
+        width,
         ...positioningCoords
       }}
     >
@@ -47,10 +47,10 @@ export const DropdownPanel = ({
 DropdownPanel.defaultProps = {
   coords: null,
   children: null,
-  maxWidth: null,
   modifier: null,
   onClickScreen: (evt) => evt,
   onExit: (evt) => evt,
+  width: null,
 };
 
 DropdownPanel.propTypes = {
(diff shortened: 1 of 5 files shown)


packages/sage-react/lib/Dropdown/DropdownPanel.jsx
```diff
@@ -7,16 +7,17 @@ import classnames from 'classnames';
...
   onClickScreen,
   onExit,
+  width,
 }) => {
   const menuEl = useRef(null);
   const classNames = classnames(
     'sage-dropdown__panel',
     {
       [`sage-dropdown__panel--${modifier}`]: modifier,
+      [`sage-dropdown__panel--custom-width`]: width,
```
joedakroub, line 20: This matches the Rails implementation.

packages/sage-react/lib/Dropdown/DropdownPanel.jsx
```diff
@@ -35,7 +36,7 @@ export const DropdownPanel = ({
       onClick={handlePanelClick}
       role="dialog"
       style={{
-        maxWidth,
+        "--custom-panel-width": width,
```
joedakroub, line 39: This matches the Rails implementation.

monicawheeler (approved):
🙌🏻

philschanely (changes requested):
@joedakroub while these changes make sense I'm hesitant to remove the changed props altogether without an audit of uses to first ensure we're not creating a regression. Also, the ticket perhaps was not clear enough that the hope was to add ability to control the dropdown trigger size, not adjust how the panel size is controlled. I'll update the ticket wording for clarity. Sorry about that!

PR #1210, opened 2022-02-14
Version bump ~v4.45.0

monicawheeler:
1. (**LOW**) - kajabi/sage-lib#1178 - Adds some MVP scoping options for Sage to be used in scoped member-facing contexts. _NOTE: Sage Engineering has reviewed this work in a number of `kp` contexts with the Bridge on and found no regressions._


anechol (approved):
🚀

PR #1215, opened 2022-02-17
Add Dot component

philschanely:
## Description

This PR adds a new component, `SageDot`, for adding a small colored dot distinctly into a layout or pattern, such as what is currently in the Legend and Statbox.

## Screenshots

Screen Shot 2022-02-16 at 9 34 41 PM

## Testing in `sage-lib`

See Components > Dot and the same in React

1. (**N/A**) Adds new `SageDot` component.

### DIFF
--- a/docs/app/helpers/components_helper.rb
+++ b/docs/app/helpers/components_helper.rb
@@ -172,10 +172,10 @@ module ComponentsHelper
       {
         title: "dot",
         description: "Dots provide a subtle color cue to place beside text or other elements.", 
-        scss: "todo",
-        rails: "todo",
-        react: "todo",
-        a11y: "todo",
+        scss: "done",
+        rails: "done",
+        react: "done",
+        a11y: "done",
         react_component_slug: "sage-dot--default",
         figma_embed: nil, # TODO: We'll have a link for this in the new design system
       },
--- a/docs/app/helpers/components_helper.rb
+++ b/docs/app/helpers/components_helper.rb
@@ -170,6 +170,16 @@ module ComponentsHelper
         figma_embed: "https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2F9Km09NjlZHYWsMP7EGT8tI%2FSage-3-for-Admin%3Fnode-id%3D4195%253A20283",
       },
       {
+        title: "dot",
+        description: "Dots provide a subtle color cue to place beside text or other elements.", 
+        scss: "todo",
+        rails: "todo",
+        react: "todo",
+        a11y: "todo",
+        react_component_slug: "sage-dot--default",
+        figma_embed: nil, # TODO: We'll have a link for this in the new design system
+      },
+      {
         title: "description",
         description: "A description is composed of title and data content and is structured as a definition list.",
         scss: "done",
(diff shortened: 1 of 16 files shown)


packages/sage-react/lib/Dot/configs.js
```diff
@@ -0,0 +1,9 @@
+export const DOT_COLORS = {
```
monicawheeler, line 1: It appears that `grey` is missing here.

monicawheeler (changes requested):
Looking great! A small request/possible change in the React side noted inline and also, a couple of questions:
* Is there a Figma file that can be added to the PRs description to verify design parity? 
* Is there a need for the other instances of "dot-like" components like you mentioned with Legend and StatBox? If not, is there deprecation considerations for `kp` to be noted for this?

docs/app/helpers/components_helper.rb
```diff
@@ -169,6 +169,16 @@ def sage_components
...
       },
+      {
+        title: "dot",
+        description: "Dots provide a subtle color cue to place beside text or other elements.", 
+        scss: "todo",
+        rails: "todo",
+        react: "todo",
+        a11y: "todo",
+        react_component_slug: "sage-dot--default",
+        figma_embed: nil, # TODO: We'll have a link for this in the new design system
```
philschanely, line 180: @monicawheeler see comment here. This is a new "atom" that is needed for some product teams but it doesn't seem worth asking design to add it into Figma's 3.0 file at this time. Def on the list for the new system though.

If a comparable dot is still desired, see the dots in the StatBox component here https://www.figma.com/file/9Km09NjlZHYWsMP7EGT8tI/Sage-3-for-Admin?node-id=6999%3A22013

packages/sage-react/lib/Dot/configs.js
philschanely: Got it now; good catch!

philschanely:
@monicawheeler mind another look now? I've added a comment regarding the Figma spec and fixed the missing color.

monicawheeler (approved):
LGTM!

docs/app/helpers/components_helper.rb
```diff
@@ -169,6 +169,16 @@ def sage_components
         react_component_slug: "sage-copytext--default",
         figma_embed: "https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2F9Km09NjlZHYWsMP7EGT8tI%2FSage-3-for-Admin%3Fnode-id%3D4195%253A20283",
       },
+      {
+        title: "dot",
+        description: "Dots provide a subtle color cue to place beside text or other elements.", 
+        scss: "todo",
```
QuintonJason, line 175: Can these values be updated to reflect on the status page?

philschanely: Any `a11y` concerns for that item's status?

QuintonJason: you added aria stuff so it looks completed

philschanely: These are updated and ready for another look @QuintonJason .

docs/app/helpers/components_helper.rb
philschanely: Sorry to say there is no great spec for this to link in yet though the styling has been in the system nested in other components. I separated that out for distinct usage and we can def circle back in the redesign to get a spec and tighten this up.

QuintonJason (approved):

PR #1216, opened 2022-02-18
Next version bump ~4.45.0

philschanely:
- (N/A) - #1215  Adds new SageDot component.
- (LOW) - #1220 adds truncation mixin. Not used in app but add rules to limit breakouts in CSS grids


monicawheeler (approved):
🎉

PR #1232, opened 2022-02-25
Sage bump patch: Fix button type formatting

philschanely:
## Description

Patches a typo on DropdownItem where type="button" was coming through incorrectly.

1. (**LOW**) Patches new dropdown item buttons `type` output


anechol (approved):
:shipit:

PR #1235, opened 2022-02-25
[SAGE-306]: Rename `radius-input` to `radius-medium` (border-radius)

pixelflips:
## Description
Renames `radius-input` to `radius-medium` to reduce confusion around this new value for use with `border-radius`. After further investigation, it appears the value is used in more locations than just inputs/form elements.

## Screenshots
N/A

## Testing in `sage-lib`
Visit sidebar, link, and input pages and verify border-radius of elements still appears as expected.

N/A

## Related
https://kajabi.atlassian.net/browse/SAGE-306


anechol (approved):
:shipit: 🚀

monicawheeler (approved):
🙌🏻

PR #1234, opened 2022-02-25
Add product creation wizard course mocks

philschanely:
## Description

Adds first round of mocks for Course portion of Product Creation Wizard.

## Screenshots

Screen Shot 2022-02-25 at 2 00 21 PM

## Testing in `sage-lib`

See http://localhost:4000/pages/mock/product_creation_wizard and http://localhost:4100/?path=/story/mocks-product-creation-wizard--default

## Testing in `kajabi-products`

N/A Mocks only

PR #1245, opened 2022-03-02
SAGE-302 - Restore Creation Wizard mocks

monicawheeler:
# Restore PR

This PR is to restore work from #1234 that was reverted due to causing the Rails docs site to error and not display.

# From original PR:

## Description

Adds first round of mocks for Course portion of Product Creation Wizard.

## Screenshots

Screen Shot 2022-02-25 at 2 00 21 PM

## Testing in `sage-lib`

- http://localhost:4000/pages/mock/product_creation_wizard
- http://localhost:4100/?path=/story/mocks-product-creation-wizard--default

N/A Mocks only

### DIFF
--- a/docs/app/helpers/mocks_helper.rb
+++ b/docs/app/helpers/mocks_helper.rb
@@ -37,61 +37,18 @@ module MocksHelper
 
   def sage_mocks
     [
+      # DOING
       {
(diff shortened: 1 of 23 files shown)


anechol (approved):
🚀

PR #1258, opened 2022-03-09
[SAGE-343] utility - alignment

QuintonJason:
## Description

- [x] add flex properties to existing 

## Screenshots

Screen Shot 2022-03-10 at 5 32 25 PM

## Testing in `sage-lib`

Visit the alignment page to see the utility classes being used:
- [Alignment Helper Page](http://localhost:4000/pages/helpers/alignment)

1. (**LOW**) Add flex properties to type alignment utility classes.

## Related

Closes [SAGE-343](https://kajabi.atlassian.net/browse/SAGE-343)


monicawheeler (changes requested):
@QuintonJason - what do you think about expanding this a bit to use `start` and `end` keywords rather than `left` and `right`?  This would require a bit more explanation to clarify `start` is `left` in `ltr` languages but it prepares the app for a future state of internationalization.  

_Note: I am marking as "Request Changes" for the question, not as a requirement to make inquired changes._

PR #1264, opened 2022-03-11
[SAGE-351] Grid - add generic col class when no size is specified

QuintonJason:
## Description

- [x] add generic col class when `size` not specified show that row item has a `width`

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-03-11 at 2 41 14 PM|Screen Shot 2022-03-11 at 2 08 11 PM|

## Testing in `sage-lib`

In the react Grid story, remove a `size` from one of the grid columns and verify that the grid doesn't break due to weird wrapping

1. (**LOW**) Adds generic grid column for responsive grid

## Related

Closes [SAGE-332](https://kajabi.atlassian.net/browse/SAGE-332)


pixelflips (approved):
LGTM! 🔥  Nice work!

PR #1281, opened 2022-03-24
Select dropdown rework

philschanely:
## Description

Draft only here. The necessary fix was implemented in `kp` here: https://github.com/Kajabi/kajabi-products/pull/22877

## Screenshots

|  Before  |  After  |
|--------|--------|
|||


philschanely:
Closing this as a solution was provided to team that needed it and this work is likely to sit stale for a while. We can return to this PR in future if needed.

PR #1286, opened 2022-03-24
[Sage-300] Link - Default Class Patch

anechol:
## Description

Adds new prop to remove default link class in React as needed after Foundation updates to the Link component.

The goal of this is to prevent style clashing from happening with components that use the Sage Link component in their architecture. The `sage-link` class should, ideally, be used by default, but some components have their own styling for those instances of the Link.

### Implementation Updates
- [React uses in `kp` may need default classname suppressed (Sage-301)](https://kajabi.atlassian.net/browse/SAGE-301)

### Follow-up Tickets
N/A

## Screenshots

Screen Shot 2022-03-24 at 3 24 51 PM

### React
- Navigate to [Link](http://localhost:4100/?path=/docs/sage-link--primary) component
- Check the following components to ensure Link styles are showing correctly:
  - Alert
  - Breadcrumbs
  - DropdownItem
  - HelpLink
  - MediaTile
  - TransactionCard
  - TypeaheadItem

(MEDIUM) Adds new prop to remove default link class in React as needed after Foundation updates to the Link component. Effect on Kajabi Products will be addressed in the above implementation ticket.

## Related

[Sage-300](https://kajabi.atlassian.net/browse/SAGE-300)

### DIFF
--- a/packages/sage-react/lib/Alert/Alert.story.jsx
+++ b/packages/sage-react/lib/Alert/Alert.story.jsx
@@ -23,7 +23,7 @@ export default {
         >
           Get unlimited pages
         </Button>
-        <Link href="//example.com">Check Usage</Link>
+        <Link suppressDefaultClass={true} href="//example.com">Check Usage</Link>
       </>
     ),
     color: Alert.COLORS.APPROACHING,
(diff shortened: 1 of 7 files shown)


monkeypox8 (approved):
👍 LGTM

monkeypox8:
One oddity I noticed during testing was that the typeahead dropdown items are misaligned/truncated at all screen sizes. @monicawheeler I wasn't able to locate an update issue for that particular component in Jira—happy to file or add to the spot check spreadsheet (whichever is preferred).

<details>
<summary>typeahead</summary>

</details>

philschanely (approved):

PR #1287, opened 2022-03-25
SAGE-413 Dropdown preview fix

monkeypox8:
## Description
Addresses syntax errors with Rails dropdown in docs site preview page. Updates unclosed blocks and cleans up formatting.

Screen Shot 2022-03-25 at 10 08 12 AM 10 12 24 AM

## Testing in `sage-lib`
1. Navigate to the [dropdown component page](http://localhost:4000/pages/component/dropdown)
2. Verify that the preview renders as expected

1. (**N/A**) Dropdown: addresses preview on docs site only. No effect on component use.

### DIFF
--- a/docs/app/views/examples/components/dropdown/_preview.html.erb
+++ b/docs/app/views/examples/components/dropdown/_preview.html.erb
@@ -1,4 +1,4 @@
-<%
+<%-
 menu_items = [
   {
     attributes: { "href": "#" },
@@ -60,7 +60,8 @@ custom_items = [
       graphic: { icon: "pen", card_color: "published" },
       name: "Subscription",
     },
-  }, {
+  },
+  {
     attributes: { "href": "#" },
     value: {
       description: "Require several...",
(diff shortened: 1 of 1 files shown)


anechol (approved):
🚀

pixelflips (approved):
The preview does render and I don't believe it's related to these updates, but I am getting some js errors and all but the top 3 dropdowns don't seem to function and the UI doesn't render properly.

Going to approve as the other issues can be in a follow-up as needed and I'm pretty confident this PR isn't the cause.

monkeypox8:
> The preview does render and I don't believe it's related to these updates, but I am getting some js errors and all but the top 3 dropdowns don't seem to function and the UI doesn't render properly.

Yeah, [SAGE-292](https://kajabi.atlassian.net/browse/SAGE-292) should cover this. The component itself is the issue. Something in dropdown trigger has been changed recently for the foundations work, but the JS here has not been updated yet.

PR #1291, opened 2022-03-25
[Focus-Ring] use box-shadow instead of outline

QuintonJason:
## Description

- [x] use `box-shadow` property instead of `outline` to create the focus ring

## Screenshots

|  Before  |  After  |
|--------|--------|
|box-shadow-focus-before|box-shadow-focus|

## Testing in `sage-lib`

Go to the Pagination Rails views and verify that when tabbing through the page there is not a `1px` gap between the focus ring and the element's background

DO NOT test by inspecting the element, and right clicking. Doing that will not recreate the issue.

1. (**HIGH**) Affects the `:focus` state of all interactive elements.
   - [ ] all interactive elements: buttons, links, form input


monicawheeler (approved):
👍 LGTM

pixelflips (approved):
LGTM! Thanks for the whitespace clean up too! 👍🏼

PR #1296, opened 2022-03-30
[SAGE-390] Upload Card foundations update

QuintonJason:
## Description

- [x] update file upload to align with foundations styles

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-03-29 at 8 06 33 PM|Screen Shot 2022-03-29 at 8 06 12 PM|

## Testing in `sage-lib`

Visit the Upload Card views and verify alignment with the foundations styles:
- [Rails](http://localhost:4000/pages/component/upload_card)
- [React](http://localhost:4100/?path=/docs/sage-uploadcard--default)

1. (**LOW**) Minor visual style update to Upload Card component.

## Related

Closes [SAGE-390](https://kajabi.atlassian.net/browse/SAGE-390)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_upload_card.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_upload_card.scss
@@ -12,9 +12,12 @@ $-upload-card-preview-width: rem(32px);
 $-upload-card-preview-max-width: rem(190px);
 $-upload-card-preview-max-width-mobile: rem(304px);
 $-upload-card-border-color-default: sage-color(grey, 300);
-$-upload-card-border-color-hover: sage-color(grey, 400);
-$-upload-card-border-color-error: sage-color(red, 500);
-$-upload-card-border-color-focus: sage-color(primary, 500);
+$-upload-card-border-color-hover: sage-color(primary, 600);
+$-upload-card-border-color-error: sage-color(red, 600);
+$-upload-card-border-color-error-hover: sage-color(red, 800);
+$-upload-card-border-color-focus: sage-color(primary, 600);
+$-upload-card-border-color-focus-ring: 0 0 0 4px sage-color(primary, 300);
+$-upload-card-border-color-error-focus-ring: 0 0 0 4px sage-color(red, 300);
 $-upload-card-background: sage-color(white);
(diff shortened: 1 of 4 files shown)


pixelflips (changes requested):
I think the color for the `.sage-upload-card__subtext` is supposed to be Grey/600. Also seems the down arrow on the upload/action button was removed and the copy was adjusted. Not entirely sure "Click to upload" is the greatest CTA text though.

packages/sage-assets/lib/stylesheets/components/_upload_card.scss
```diff
@@ -104,7 +122,7 @@ $-upload-card-background-hover: sage-color(grey, 100);
   > p {
     @extend %t-sage-body-small;
```
anechol, line 123: The design file doesn't explicitly say this, but I think the subtexts are now supposed to be a font weight of Book (Medium).

anechol (changes requested):
Just a couple things to update if possible:
- Can we update the icon size to `24px`? That's about the closest the Figma design has currently
- Can we update the button font size to be 16px? Would that require an update to buttons overall?

And one question: is there a reasoning behind using SVGs for our dotted borders over the CSS property?

Other than that, good update work!!

QuintonJason:
> * Can we update the icon size to `24px`? That's about the closest the Figma design has currently
Can do

> * Can we update the button font size to be 16px? Would that require an update to buttons overall?
That would need to happen in a button update as their base font size is rendering at `16px` 
> And one question: is there a reasoning behind using SVGs for our dotted borders over the CSS property?
It's because the design is using a custom `dotted` border. There currently no way to customized the length of the dash

QuintonJason:
> I think the color for the `.sage-upload-card__subtext` is supposed to be Grey/600. Also seems the down arrow on the upload/action button was removed and the copy was adjusted. Not entirely sure "Click to upload" is the greatest CTA text though.

Updated the color, but I'll leave the copy since that prop can be set on the component

pixelflips (approved):
Besides the linter yelling, LGTM! 👍🏼

anechol (approved):
:shipit:

PR #1297, opened 2022-03-30
Table: Select All Checkbox

philschanely:
## Description

This PR adds functionality in React for a select all toggle in Table and a demo of how similar functionality can be added in Rails `sage_table_for`. 

## Screenshots

Screen Shot 2022-03-30 at 10 30 47 AM

Screen Shot 2022-03-30 at 10 30 56 AM

## Testing in `sage-lib`

See http://localhost:4100/?path=/story/sage-table--default
See http://localhost:4000/pages/component/table 

1. (LOW): Adds select all checkbox abilities in React Table (Rails requires manual construction using `sage_table_for`. See Docs site > Components > Table

## Related

https://kajabi.atlassian.net/browse/SAGE-236


monicawheeler (approved):
LGTM! Nice misspelling catches as well!

PR #1298, opened 2022-03-30
[SAGE-294] Search foundations update

QuintonJason:
## Description

- [x] align search component to foundations 

## Screenshots

|  Before  |  After  |
|--------|--------|
|search-before-before|search-after|

## Testing in `sage-lib`

Visit the search component views and verify alignment with foundations spec:
- [Rails](http://localhost:4000/pages/component/search)
- [React](http://localhost:4100/?path=/story/sage-search--default)

A part of the foundations Epic PR

## Related

Closes [SAGE-294](https://kajabi.atlassian.net/browse/SAGE-294)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_search.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_search.scss
@@ -128,6 +128,10 @@ $-search-icon: "::before";
     border-radius: sage-border(radius);
   }
 
+  .sage-search--contained & {
+    border-radius: sage-border(radius);
+  }
+
   .sage-search--contained &:focus,
   .sage-search--contained &:focus:not(:placeholder-shown),
   .sage-search--contained &:active,
--- a/packages/sage-assets/lib/stylesheets/components/_search.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_search.scss
@@ -18,26 +18,23 @@ $-search-icon: "::before";
     z-index: 2;
     margin-left: sage-spacing(xs);
     margin-right: -1 * sage-spacing();
-    color: sage-color(grey, 500);
+    color: sage-color(charcoal, 400);
     transition: map-get($sage-transitions, default);
     transition-property: color;
-
-    :focus-within > & {
-      color: sage-color(primary, 500);
-    }
   }
 
   &:not(.sage-search--contained) {
     border-bottom: 1px solid sage-color(grey, 300);
 
     &:hover {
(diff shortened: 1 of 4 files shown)


packages/sage-assets/lib/stylesheets/components/_search.scss
```diff
@@ -18,26 +18,23 @@ $-search-icon: "::before";
     z-index: 2;
     margin-left: sage-spacing(xs);
     margin-right: -1 * sage-spacing();
-    color: sage-color(grey, 500);
+    color: sage-color(charcoal, 400);
```
monicawheeler, line 21: Looks like this value should be:

```suggestion
    color: sage-color(grey, 600);
```

packages/sage-assets/lib/stylesheets/components/_search.scss
```diff
@@ -58,7 +55,10 @@ $-search-icon: "::before";
 
     &:focus-within {
       z-index: sage-z-index(default, 2);
-      box-shadow: map-get($sage-toolbar-button-borders, focus);
+      border-radius: sage-border(radius);
+      // box-shadow: map-get($sage-toolbar-button-borders, focus);
```
monicawheeler, line 59: Do we need to keep this comment?

packages/sage-assets/lib/stylesheets/components/_search.scss
```diff
@@ -128,7 +128,7 @@ $-search-icon: "::before";
   &:focus:not(:disabled),
   &:active:not(:disabled),
   &:valid:not(:placeholder-shown) {
-    box-shadow: none;
+    // box-shadow: none;
```
monicawheeler, line 131: Also, should this comment be removed?

monicawheeler (changes requested):
A couple of notes to review!

anechol (changes requested):
Nice work! Just a few changes:
- Can we update the disabled font color to `Grey 600`. The icon looks like it should be about the same color but it's listed as `Charcoal 400` in the Design files. Judging by the color system, it does seem to line up with what the current Grey color should be so the icon might be fine as is.
- You can barely see this, but can we fix the border spacing here to match with the background?
Screen Shot 2022-03-31 at 10 32 07 AM

General question: 
I was under the impression that we were deprecating floating labels? Is this no longer the case?

anechol (approved):
🚀

PR #1305, opened 2022-04-01
[Sage-395] Avatar - Foundations Update

anechol:
## Description

No significant changes. Updates Sage Docs.

[Figma](https://www.figma.com/file/sMbtLUHSt2vfKgKjyQ3052/%E2%9A%A0%EF%B8%8F-%5BDO-NOT-USE-WIP%5D-Sage-Update?node-id=484%3A54749)

### Implementation Changes
N/A

### Follow-up Items
- [Deprecate initials variation](https://kajabi.atlassian.net/browse/SAGE-427)

## Screenshots

|  Before  |  After  |
|--------|--------|
Screen Shot 2022-04-01 at 4 05 53 PM|Screen Shot 2022-04-01 at 4 25 23 PM

### Rails
- View [Rails component](http://localhost:4000/pages/component/avatar)
- Check that initials examples are removed

(**LOW**) No significant changes. Updates Sage Docs.

## Related

[Sage-395](https://kajabi.atlassian.net/browse/SAGE-395)


QuintonJason (approved):

monkeypox8 (commented):
Rails side LGTM, however I wanted to confirm that the React story will keep the initials in place.

Also, since `initials` (and `lazy_load_initials`) are to be deprecated, I'm wondering if we label those properties/remove from the documentation.

monkeypox8:
> > Rails side LGTM, however I wanted to confirm that the React story will keep the initials in place.
> > Also, since `initials` (and `lazy_load_initials`) are to be deprecated, I'm wondering if we label those properties/remove from the documentation.
> 
> @monkeypox8 Initials should be removed from both. I'm not sure how to remove these props from Storybook without removing the prop from the component itself. There are some Avatars in usage in Products that are using the `initials` prop.

Oh, I see. In this case, I think we're safe to remove both props from both the story and component, and add a follow-up issue to patch the use in `kp` since the feature branch won't merge for a little while yet. Otherwise, if left alone others may continue to use `initials` in React without knowing it's deprecated, creating more areas to patch later. Let me know what you think.

anechol:
> > > Rails side LGTM, however I wanted to confirm that the React story will keep the initials in place.
> > > Also, since `initials` (and `lazy_load_initials`) are to be deprecated, I'm wondering if we label those properties/remove from the documentation.
> > 
> > 
> > @monkeypox8 Initials should be removed from both. I'm not sure how to remove these props from Storybook without removing the prop from the component itself. There are some Avatars in usage in Products that are using the `initials` prop.
> 
> Oh, I see. In this case, I think we're safe to remove both props from both the story and component, and add a follow-up issue to patch the use in `kp` since the feature branch won't merge for a little while yet. Otherwise, if left alone others may continue to use `initials` in React without knowing it's deprecated, creating more areas to patch later. Let me know what you think.

Yep, that totally makes sense. I've removed the props from React. The deprecation follow-up ticket has already been created so we're good to go on that end.

monkeypox8 (approved):
👍 LGTM, nice work!

PR #1318, opened 2022-04-11
[Sage-369] Modal - Content Scrolling Prop

anechol:
## Description

Adds `allow_scroll`/`allowScroll` prop to Rails and React to allow modal content to scroll.

## Screenshots

https://user-images.githubusercontent.com/14791307/162787300-1eab1d3c-316d-4f04-90e2-5623957995b8.mov

## Testing in `sage-lib`

- View [Modal](http://localhost:4000/pages/component/modal)
- Add extra body content and `allow_scroll` prop
- Check that body content scrolls as necessary.

(LOW) Adds `allow_scroll`/`allowScroll` prop to Rails and React to allow modal content to scroll. Scrolling is guarded behind an opt-in prop, so modals across the app will not be affected.

## Related

[Sage-369](https://kajabi.atlassian.net/browse/SAGE-369)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_modal.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_modal.scss
@@ -67,19 +67,6 @@ $-modal-fullscreen-top-spacing: rem(104px);
 
 .sage-modal--scrollable {
   overflow-y: initial;
-
-  .sage-modal__container,
-  .sage-modal__container > .simple_form {
-    display: grid;
-    grid-template-rows: auto 1fr auto;
-    max-height: 85vh;
-  }
-
-  .sage-modal__container .sage-modal__content,
-  .sage-modal__container > .simple_form {
-    overflow: auto;
-    padding: sage-spacing(xs) sage-spacing(3xs);
-  }
 }
 
 .sage-modal__container {
@@ -153,6 +140,18 @@ $-modal-fullscreen-top-spacing: rem(104px);
     border-radius: 0;
     opacity: 1;
   }
+
+  .sage-modal--scrollable &,
+  .sage-modal--scrollable & > .simple_form {
+    display: grid;
+    grid-template-rows: auto 1fr auto;
+    max-height: 85vh;
+  }
+  
+  .sage-modal--scrollable & > .simple_form {
+    @include overflow-scroll(y);
+    padding: sage-spacing(xs) sage-spacing(3xs);
+  }
 }
 
 .sage-modal__header {
@@ -221,6 +220,11 @@ $-modal-fullscreen-top-spacing: rem(104px);
 
   .sage-modal--fullscreen & {
     margin-top: $-modal-fullscreen-top-spacing;
+  }
+
+  .sage-modal--scrollable & {
+    @include overflow-scroll(y);
+    padding: sage-spacing(xs) sage-spacing(3xs);
   }
 
   .sage-drawer & {
--- a/packages/sage-assets/lib/stylesheets/components/_modal.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_modal.scss
@@ -78,7 +78,7 @@ $-modal-fullscreen-top-spacing: rem(104px);
   .sage-modal__container .sage-modal__content,
   .sage-modal__container > .simple_form {
     overflow: auto;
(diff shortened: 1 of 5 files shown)


anechol:
Hey @philschanely, I would definitely love your input here. This should give Manage the flexibility they're expecting without affecting other modals across the app. That being said, I've had some conversations with @QuintonJason about Modals that left me wondering if we should create a follow-up ticket to refactor the component to fix clipping issues with other components like Dropdown.

packages/sage-assets/lib/stylesheets/components/_modal.scss
```diff
@@ -65,6 +65,23 @@ $-modal-fullscreen-top-spacing: rem(104px);
...
+  .sage-modal__container > .simple_form {
+    display: grid;
+    grid-template-rows: auto 1fr auto;
+    max-height: 85vh;
+  }
+
+  .sage-modal__container .sage-modal__content,
+  .sage-modal__container > .simple_form {
+    overflow: auto;
+    padding: 8px 2px;
```
QuintonJason, line 81: Can these be updated to use the `sage-spacing()` values?

QuintonJason:
Is it possible to add a Scrollableable Content modal demo / button with a long text var for the overflow content?

monkeypox8 (approved):
👍 LGTM!

packages/sage-assets/lib/stylesheets/components/_modal.scss
```diff
@@ -65,6 +65,23 @@ $-modal-fullscreen-top-spacing: rem(104px);
   }
 }
 
+.sage-modal--scrollable {
+  overflow-y: initial;
+
+  .sage-modal__container,
```
philschanely, line 71: For these nested selectors, precedent might suggest that we place these under the other styles for `sage-modal__container like this:

```scss
.sage-modal__container {
  // other existing styles
  
  .sage-modal--scrollable &,
  .sage-modal--scrollable & > .simple_form {
    // styles here
  }
  
  // etc...
}
```

Aim being, as much as practical, to have a given element's styles all consolidated in one place so we can more easily see what is affecting it together.

philschanely: Admittedly, it gets thicker when we add the `.simple_form` layer, but I think here's its grouped for that tangential case where it's nested within our modal.

packages/sage-assets/lib/stylesheets/components/_modal.scss
```diff
@@ -65,6 +65,23 @@ $-modal-fullscreen-top-spacing: rem(104px);
...
+  .sage-modal__container > .simple_form {
+    display: grid;
+    grid-template-rows: auto 1fr auto;
+    max-height: 85vh;
+  }
+
+  .sage-modal__container .sage-modal__content,
```
philschanely, line 78: Given comment above, these might should split up with `__content` moving to group with that element in this stylesheet, but the simple form might should go with the `__container`. Its a grey area for sure since its technically outside of our system.

philschanely (changes requested):
Definitely on the desired track here. I've left some change requests in terms of CSS organization. LMK if I can assist or clarify.

packages/sage-assets/lib/stylesheets/components/_modal.scss
```diff
@@ -65,6 +65,23 @@ $-modal-fullscreen-top-spacing: rem(104px);
...
+  .sage-modal__container > .simple_form {
+    overflow: auto;
```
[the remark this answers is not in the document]
anechol: Thanks for mentioning this, I didn't know we had this mixin.

packages/sage-assets/lib/stylesheets/components/_modal.scss
anechol: Thanks for all the feedback, I've made some changes.

QuintonJason (changes requested):
The `margin-top` from the `.sage-modal__content` is cutting into the scrollable area. Can we move the spacing from the parent to the first child?

Something like:
```
margin-top: 0;

& > :first-child {
  margin-top: sage-spacing(xs);
}
```

QuintonJason (approved):

PR #1322, opened 2022-04-12
Table: Disable row selectable

philschanely:
## Description

This PR adds ability to disable selection on individual rows within a Sage Table (React). 

## Screenshots

Screen Shot 2022-04-12 at 3 09 09 PM

## Testing in `sage-lib`

See http://localhost:4100/?path=/story/sage-table--default

1. (LOW) New props added to TableRow to allow hiding and disabling checkbox on individual rows.


anechol (approved):
LGTM

PR #1326, opened 2022-04-13
[SAGE-440] Foundations link update 

QuintonJason:
## Description

- [x] Reword the heading for the Links documentation page to more closely align with the Design file
- [x] Create a "Default" and "Plain" section on the doc site to align with the designs
- [x] Add a "with icon" variation

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-04-14 at 3 21 35 PM|Screen Shot 2022-04-14 at 3 19 58 PM|

## Testing in `sage-lib`

Verify that the links align with the latest specs:
- [Rails](http://localhost:4000/pages/component/link)
- [React](http://localhost:4100/?path=/story/sage-link--primary)

## Related

Closes [SAGE-440](https://kajabi.atlassian.net/browse/SAGE-440)

### DIFF
--- a/docs/app/views/examples/components/link/_preview.html.erb
+++ b/docs/app/views/examples/components/link/_preview.html.erb
@@ -1,3 +1,55 @@
+<%
+demo_configs_links = [
+  {
+    heading: "Primary",
+    label: "Primary",
+    style: "primary",
+  },{
+    heading: "Secondary",
+    label: "Secondary",
+    style: "secondary",
+  },
+  {
+    heading: "Neutral",
+    label: "Neutral",
+    style: "neutral",
+  },
+  {
+    heading: "Danger",
+    label: "Danger",
+    style: "danger",
+  }
+]
+%>
+
+<h3 class="t-sage-heading-6">Default links</h3>
+<%= sage_component SagePanelStack, {} do %>
+  <% demo_configs_links.each do | config | %>
+    <%= sage_component SageLink, {
+      url: "#",
+      label: config[:heading],
+      external: false,
+      launch: false,
+      help_link: false,
+      show_label: true,
+      style: config[:style],
+    } %>
+  <% end %>
+<% end %>
+<h3 class="t-sage-heading-6">Plain links</h3>
+<%= sage_component SagePanelStack, {} do %>
+  <%= sage_component SageLink, {
+    url: "#",
+    label: "Plain",
+    external: false,
+    launch: false,
+    help_link: false,
+    remove_underline: true,
+    show_label: true,
+    style: "secondary"
+  } %>
+<% end %>
+
 <h3 class="t-sage-heading-6">Standard link</h3>
 <%= sage_component SagePanelStack, { css_classes: "sage-type" } do %>
   <%= sage_component SageLink, {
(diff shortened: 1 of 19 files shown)


QuintonJason:
Screen Shot 2022-04-14 at 3 24 27 PM

There is currently an issue with the icon variations of the link. The underline from the parent link is showing in the child pseudo `::before` element. Since the `::before` must be able to switch sides, it exposed the fact that we'll need a heavy rework to accomplish this.

With this, the issue would not be present if the icon was a svg element instead of a pseudo `::before`. Looking at this, I believe the reason we have a link variant in our buttons was because we need to put the icon on the left.

anechol (changes requested):
Should the focus text color be black?
Screen Shot 2022-04-15 at 12 56 44 PM

monkeypox8:
> There is currently an issue with the icon variations of the link. The underline from the parent link is showing in the child pseudo `::before` element. Since the `::before` must be able to switch sides, it exposed the fact that we'll need a heavy rework to accomplish this.

Short of a full refactor (which I still think may be warranted as a follow-up), and the fact that `.sage-link` already has `position: relative`,  I wonder if we could do something like this for now:

```
[class*="sage-link--icon-left-"] {
  padding-left: calc(#{sage-spacing(sm)} + #{sage-spacing(2xs)}); // icon width + icon margin
}

[class*="sage-link--icon-left-"]::before {
  position: absolute;
  left: 0;
}
```

… and the same but reversed for `sage-link--icon-right-`

monkeypox8:
> Should the focus text color be black? Screen Shot 2022-04-15 at 12 56 44 PM

This does match the current Figma comps, presumably to have a consistent focus ring throughout. Though I agree that out of context it's a little jarring.

monkeypox8 (approved):
👍 LGTM overall pending design resolution of focus outline style on "danger" links. 

The two other items noted in various comments (icon & icon position dropdown and the underline appearing on icons) can be handled as follow-ups.

anechol (approved):
Nice work!! LGTM

QuintonJason:
> > There is currently an issue with the icon variations of the link. The underline from the parent link is showing in the child pseudo `::before` element. Since the `::before` must be able to switch sides, it exposed the fact that we'll need a heavy rework to accomplish this.
> 
> Short of a full refactor (which I still think may be warranted as a follow-up), and with the knowledge that `.sage-link` already sets `position: relative`, I wonder if we could do something like this for now:
> 
> ```
> [class*="sage-link--icon-left-"] {
>   padding-left: sage-spacing(sm) + sage-spacing(2xs); // icon width + icon margin
> }
> 
> [class*="sage-link--icon-left-"]::before {
>   position: absolute;
>   left: 0;
>   height: 1em;  // fix for `small` link variant
>   width: 1em;
> }
> ```
> 
> … and the same but reversed for `sage-link--icon-right-`
> 
> Line breaks may be weird, though. So the SVG approach you'd mentioned would still be favorable in the long run.
> 
> EDIT: just noticed that the icons don't inherit the link size when set to `small` and added an adjustment

Sage Jira ticket to track: [SAGE-447](https://kajabi.atlassian.net/browse/SAGE-447)

PR #1328, opened 2022-04-13
fix(alert): add spacing back in to SageAlert to avoid breaking changes

monicawheeler:
## Description

Return `margin-bottom` to Alerts to avoid breaking `kp`


QuintonJason (approved):
💯

PR #1333, opened 2022-04-15
[Sage-438] Card Highlight - Border Fix

anechol:
## Description

Fixes highlight size and border radius to match new Card specs in Foundations.

## Screenshots

|  Before  |  After  |
|--------|--------|
Screen Shot 2022-04-15 at 3 13 07 PM|Screen Shot 2022-04-15 at 3 12 13 PM

## Testing in `sage-lib`

- View [Card Highlight](http://localhost:4000/pages/component/card_highlight)
- Check that alignment is corrected

(LOW) Fixes highlight size and border radius to match new Card specs in Foundations.

## Related

[Sage-438](https://kajabi.atlassian.net/browse/SAGE-438)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_card_highlight.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_card_highlight.scss
@@ -4,6 +4,9 @@
 /// @group sage
 ////
 
+$-card-highlight-border-radius: sage-border(radius-large);
+$-card-highlight-size: rem(12px);
+
 .sage-card-highlight {
   --color: #{sage-color(primary, 900)};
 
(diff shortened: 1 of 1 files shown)


anechol:
Open to alternatives to fix this issue. The new Card border radius seems to "force" this highlight to be bigger than before in order for it to match the Card's radius.

monkeypox8:
> Open to alternatives to fix this issue. The new Card border radius seems to "force" this highlight to be bigger than before in order for it to match the Card's radius.

As far as I'm aware, there really isn't a great way to handle this without either some intense [ana-tudor-ish style math](https://css-tricks.com/restricting-a-pseudo-element-to-its-parents-border-box/). Alternative would be setting the parent `card` to `overflow: hidden`, but we would then have to assume clipping is okay with any interactive card content, like tooltips or dropdowns.

monkeypox8 (approved):
👍 LGTM. The highlight bleed into the parent Card's border is barely noticeable enough that I think we can give it a pass for now, with a follow-up for later.

The one exception being the "white" highlight card, where it looks like the border is broken.

philschanely (approved):

PR #1334, opened 2022-04-16
New List component

philschanely:
## Description

This PR adds the new List component that will be used for more compositional lists and eventually to replace Sortable, Catalog Item and 

- [x] Add in Rails
- [x] Add in React

This also adds a new PoC for a Reveal utility such as could be used on List rows

- [x] Add Reveal utility

Note that the existing `SageLists` component should be deprecated:
- It is used in only a few spots in `kp` to wrap lists and there are no manual uses of `sage-list`.
- It does not apply any classes or variants directly through the component, so the styles in `_lists.scss` are not actually implements.

## Screenshots

Screen Shot 2022-04-16 at 4 12 52 PM

## Testing in `sage-lib`

- see http://localhost:4000/pages/component/list
- Storybook coming soon

1. (N/A) New List component

## Related

https://kajabi.atlassian.net/browse/SAGE-407


docs/app/views/examples/components/list/_preview.html.erb
```diff
@@ -0,0 +1,120 @@
+<%
+
```
monicawheeler, line 2: Seems to be an extra space here. Was that intentional?

docs/app/views/examples/components/list/_preview.html.erb
```diff
@@ -0,0 +1,120 @@
...
+  {
+    id: 5,
+    name: 'Product vel aliquam',
+  }
+]
+%>
+
+<h3 class="t-sage-heading-6">Default configuration</h3>
+<%= md("
+SageList is implimented most simply by passing the desired contents for the items through the `SageList` `items` property.
```
monicawheeler, line 60:
```suggestion
SageList is implemented most simply by passing the desired contents for the items through the `SageList` `items` property.
```

docs/lib/sage_rails/app/views/sage_components/_sage_list.html.erb
```diff
@@ -0,0 +1,13 @@
+<div
```
monicawheeler, line 1: What are your thoughts around defaulting to a `ul` with an option to set an `ol` since this is called a List component?

packages/sage-assets/lib/stylesheets/dictionary/_tokens.scss
```diff
@@ -3319,3 +3319,4 @@ $sd-sage-content-grid-template: (
   $value: map-get($value, $subkey);
   @return $value;
 }
+
```
monicawheeler, line 3322: An extra space seems to have been added here 🤔

monicawheeler (changes requested):
Looks great! A couple of small notes up for your consideration.

docs/lib/sage_rails/app/views/sage_components/_sage_list.html.erb
philschanely: Love it. Will put in revised PR

PR #1348, opened 2022-04-19
[SAGE-387] Description List foundations update

QuintonJason:
## Description

- [x] update position of the actions
- [x] update `grid-template-columns` and vars to account for going from 3 columns to 2, related to custom title and action widths

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-04-19 at 6 05 22 PM|Screen Shot 2022-04-19 at 6 12 16 PM|

## Testing in `sage-lib`

Visit the description list view and verify alignment with the foundations spec:
- [Rails](http://localhost:4000/pages/component/description)
- [React](http://localhost:4100/?path=/story/sage-description--multiple-items-with-action-button)

## Related

Closes [SAGE-387](https://kajabi.atlassian.net/browse/SAGE-387)

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/themes/next/_sage_description.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/themes/next/_sage_description.html.erb
@@ -38,7 +38,7 @@ end
       end
     end
     %>
-    <div 
+    <div
       class="
         sage-description__term-group
         <%= "sage-description__term-group--hide-title" if item[:hide_title] %>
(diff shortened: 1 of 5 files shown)


pixelflips (approved):
Overall LGTM! 👍🏼  Nice work!

As a side note the arrow-only actions seem a bit out of place in the docs examples (due to new placement), but not a show-stopper for this PR.

docs/lib/sage_rails/app/views/sage_components/themes/next/_sage_description.html.erb
```diff
@@ -38,7 +38,7 @@ end
       end
     end
     %>
-    <div 
+    <div
```
monkeypox8, line 41: Can we make this a `dl` rather than a `div`?

monkeypox8 (changes requested):
One quick update on the parent element, should be good to go otherwise

docs/lib/sage_rails/app/views/sage_components/themes/next/_sage_description.html.erb
monkeypox8: Marking this as resolved because I was mistaken in thinking `div` to be an invalid child of a `dl`, but this is not the case:
> In order to annotate groups with [microdata](https://html.spec.whatwg.org/multipage/microdata.html#microdata) attributes, or other global attributes that apply to whole groups, or just for styling purposes, each group in a [dl](https://html.spec.whatwg.org/multipage/grouping-content.html#the-dl-element) element can be wrapped in a [div](https://html.spec.whatwg.org/multipage/grouping-content.html#the-div-element) element. This does not change the semantics of the [dl](https://html.spec.whatwg.org/multipage/grouping-content.html#the-dl-element) element.

monkeypox8 (approved):
👍 LGTM! I'm uncertain whether the legacy components should also be updated to reflect the updated semantics but that can be handled in a follow-up if needed.

/cc @monicawheeler

PR #1358, opened 2022-04-21
Sync sage-system with `next` modifications

philschanely:
## Description

This PR adds the modifications to `sage-system` that were added in [SAGE-19_EPIC-foundations](https://github.com/Kajabi/sage-lib/pull/1185/files?file-filters%5B%5D=.js&show-deleted-files=true&show-viewed-files=true) so that both "themes" Rails components use most appropriate JS.

See the `.js` files within `sage-system` [here](https://github.com/Kajabi/sage-lib/pull/1185/files?file-filters%5B%5D=.js&show-deleted-files=true&show-viewed-files=true)

## Testing in `sage-lib`

In both themes validate that:

- [Toast](http://localhost:4000/pages/component/form_select) looks as expected in both themes (secondary close button and check icon)
- [Select](http://localhost:4000/pages/component/form_select) gets the dropdown arrow on the right
- [Input](http://localhost:4000/pages/component/form_input) gets appropriately style suffix and prefix

1. (LOW) Patches Rails JS for Next theme

NOTE: This is set to merge to `main` so as to trigger a patch version that can be used in currently running version bump.

### DIFF
--- a/packages/sage-system/lib/inputaffixes.js
+++ b/packages/sage-system/lib/inputaffixes.js
@@ -1,3 +1,5 @@
+import { isNextTheme } from './utils';
+
 Sage.inputaffixes = (() => {
 
   // ==================================================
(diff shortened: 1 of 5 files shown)


monicawheeler (approved):
LGTM

PR #1371, opened 2022-04-27
[SAGE-493] Add generic `.sage-col` when only no column size is specified

QuintonJason:
## Description

- [x] Add generic `.sage-col` when no column size is specified

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-04-27 at 8 51 43 AM|Screen Shot 2022-04-27 at 1 42 21 PM|

## Testing in `sage-lib`

**No visual regression**
Go the the Grid views:
- verify that the `.sage-col` class only shows when no `size`, `small`, `medium`, or `large` prop is set

1. (**HIGH**) Updates classes on the Grid component. Requires QA as this is a high usage component.

## Related

Closes [SAGE-493](https://kajabi.atlassian.net/browse/SAGE-493)

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_grid_col.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_grid_col.html.erb
@@ -3,7 +3,7 @@
 <div class="
   <%= "sage-col" if !has_infix -%>
   <%= "sage-col-#{component.size}" if component.size -%>
-  <%= "sage-col--#{component.breakpoint}-#{component.size}" if component.breakpoint -%>
+  <%= "sage-col--#{component.breakpoint}-#{component.size}" if component.breakpoint && component.size -%>
   <%= "sage-col--sm-#{component.small}" if component.small%>
   <%= "sage-col--md-#{component.medium}" if component.medium%>
   <%= "sage-col--lg-#{component.large}" if component.large%>
--- a/docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_grid_col.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_grid_col.html.erb
@@ -1,10 +1,14 @@
+<% has_infix = component.breakpoint || component.size || component.small || component.medium || component.large %>
+
+
 <div class="
-  <%= "sage-col" if !component.size %>
-  <%= "sage-col-#{component.size}" if component.size %>
-  sage-col--<%= component.breakpoint %>-<%= component.size %> <%= component.generated_css_classes %>
+  <%= "sage-col" if !has_infix -%>
+  <%= "sage-col-#{component.size}" if component.size -%>
+  <%= "sage-col--#{component.breakpoint}-#{component.size}" if component.breakpoint -%>
   <%= "sage-col--sm-#{component.small}" if component.small%>
   <%= "sage-col--md-#{component.medium}" if component.medium%>
   <%= "sage-col--lg-#{component.large}" if component.large%>
+  <%= component.generated_css_classes %>
 " <%= component.generated_html_attributes.html_safe %>>
   <%= component.content %>
 </div>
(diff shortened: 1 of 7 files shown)


docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_grid_col.html.erb
```diff
@@ -1,10 +1,13 @@
+<% has_infix = component.breakpoint || component.size || component.small || component.medium || component.large %>
+
 <div class="
-  <%= "sage-col" if !component.size %>
-  <%= "sage-col-#{component.size}" if component.size %>
-  sage-col--<%= component.breakpoint %>-<%= component.size %> <%= component.generated_css_classes %>
+  <%= "sage-col" if !has_infix -%>
```
philschanely, line 4: I'm unfamiliar with this term. What does it mean? 😄

docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_grid_col.html.erb
```diff
@@ -1,10 +1,13 @@
...
+  <%= "sage-col-#{component.size}" if component.size -%>
+  <%= "sage-col--#{component.breakpoint}-#{component.size}" if component.breakpoint -%>
```
ju-Skinner, line 6: I would add the second check to your if condition `&& component.size`, since there is a dependency on it.

docs/lib/sage_rails/app/views/sage_components/themes/next/_sage_grid_col.html.erb
```diff
@@ -1,10 +1,13 @@
+<% has_infix = component.breakpoint || component.size || component.small || component.medium || component.large %>
+
 <div class="
-  <%= "sage-col" if !component.size %>
-  <%= "sage-col-#{component.size}" if component.size %>
-  sage-col--<%= component.breakpoint %>-<%= component.size %> <%= component.generated_css_classes %>
+  <%= "sage-col" if !has_infix -%>
+  <%= "sage-col-#{component.size}" if component.size -%>
+  <%= "sage-col--#{component.breakpoint}-#{component.size}" if component.breakpoint -%>
```
ju-Skinner, line 6: Same as previous comment.

packages/sage-react/lib/themes/legacy/Grid/GridCol.jsx
```diff
@@ -14,10 +14,12 @@ export const GridCol = ({
   small,
   ...rest
 }) => {
+  const hasInfix = size || small || medium || large;
+
   const classNames = classnames(
     className,
     {
-      'sage-col': !size,
+      'sage-col': !hasInfix,
```
ju-Skinner, line 22: This is more for clarity...the only time this would ever equate to `true` would be if all values are `null` .

ju-Skinner (changes requested):
Requesting changes for the code in the `erb` files for `component.size`. See comments below.

packages/sage-react/lib/themes/legacy/Grid/GridCol.jsx
QuintonJason: That's correct. Did you have a recommendation that could provide more clarity?

ju-Skinner: @QuintonJason I believe it's ok as it is.

ju-Skinner (approved):
Great job.

docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_grid_col.html.erb
QuintonJason: @philschanely Sorry didn't see this but need to call attention to it. From bootstrap to google material design, that's the variable that represent the value string of the breakpoints; like `sm` or `lg` for the column class generation loop. They may have different loops and what they are dependent on but that value is always represented that way.

PR #1381, opened 2022-05-04
[Sage-464] Tokens - Breakpoints Update

anechol:
## Description

Adds map for new Sage Next layout breakpoints. The breakpoints used fall in line with some current breakpoints in `kp` and Sage Legacy styles while adhering to new, future container size updates.

Breakpoints before:
```
xs-max: 544px,
sm-min: 545px,
sm-max: 767px,
md-min: 768px,
md-max: 991px,
lg-min: 992px,
lg-max: 1199px,
xl-min: 1200px,
xxl-min: 1440px,
```

Updates breakpoints to the following values:
```
xs-max: 767px,
sm-min: 768px,
sm-max: 991px,
md-min: 992px,
md-max: 1119px,
lg-min: 1120px,
lg-max: 1279px,
xl-min: 1280px,
xl-max: 1439px,
```

## Testing in `sage-lib`

- Change `sage-breakpoint` function to use `sage-next-breakpoints` variable
- Check all current Sage components
- Check pages in `kp`
- Ensure there are no breaking layout changes.

(LOW) Adds map for new Sage Next layout breakpoints.

## Related

[Sage-464](https://kajabi.atlassian.net/browse/SAGE-464)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/themes/next/tokens/_breakpoints.scss
+++ b/packages/sage-assets/lib/stylesheets/themes/next/tokens/_breakpoints.scss
@@ -9,14 +9,16 @@
 /// Sage breakpoints token
 ///
 $sage-breakpoints: (
-  xs-max: 544px,
-  sm-min: 545px,
-  sm-max: 767px,
-  md-min: 768px,
-  md-max: 991px,
-  lg-min: 992px,
-  lg-max: 1199px,
-  xl-min: 1200px,
+  xxs-max: 699px,
+  xs-min: 700px,
+  xs-max: 799px,
+  sm-min: 800px,
+  sm-max: 989px,
+  md-min: 990px,
+  md-max: 1119px,
+  lg-min: 1120px,
+  lg-max: 1279px,
+  xl-max: 1280px,
   xxl-min: 1440px,
 );
 
--- a/packages/sage-assets/lib/stylesheets/themes/next/tokens/_breakpoints.scss
+++ b/packages/sage-assets/lib/stylesheets/themes/next/tokens/_breakpoints.scss
@@ -9,15 +9,27 @@
 /// Sage breakpoints token
 ///
 $sage-breakpoints: (
-  xs-max: 699px,
-  sm-min: 700px,
-  sm-max: 989px,
-  md-min: 990px,
+  xs-max: 544px,
+  sm-min: 545px,
+  sm-max: 767px,
+  md-min: 768px,
+  md-max: 991px,
+  lg-min: 992px,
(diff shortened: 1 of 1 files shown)


packages/sage-assets/lib/stylesheets/themes/next/tokens/_breakpoints.scss
```diff
@@ -9,14 +9,14 @@
...
-  lg-max: 1199px,
-  xl-min: 1200px,
+  xs-max: 699px,
+  sm-min: 700px,
+  sm-max: 989px,
+  md-min: 990px,
+  md-max: 1119px,
+  lg-min: 1120px,
+  lg-max: 1279px,
+  xl-max: 1280px,
```
philschanely, line 19: Here we should probably keep `min` but perhaps add a `xl-max` at 1439?

philschanely (changes requested):
Looking good. Did this include an audit of uses of the breakpoints to a) know what this impacts and b) potentially adjust any further?

packages/sage-assets/lib/stylesheets/themes/next/tokens/_breakpoints.scss
anechol: Thanks for pointing these out. i've made some updates. Let me know what you think.

monkeypox8 (approved):
👍🏼 LGTM. However, to @philschanely's point, this will have a far-reaching impact, particularly with the (soon to be deprecated) 12-col grid. A usage audit would be highly useful, perhaps as a related follow-up task.

packages/sage-assets/lib/stylesheets/themes/next/tokens/_breakpoints.scss
```diff
@@ -20,6 +20,18 @@ $sage-breakpoints: (
   xxl-min: 1440px,
 );
 
+$sage-next-breakpoints: (
+  xs-max: 767px,
```
philschanely, line 24: This is ok. Since there are a few values in common between the old and new sets, would it make sense to only record the "new" breakpoints? And even then consider whether they are needed here for the whole system versus unique to specific needs in the new app frame? For example, if the new stop and 1120 is only for the change in sidebar size, it may suffice to scope it to that component or name it for that use case?

anechol:
After pairing with @philschanely and further evaluating the potential new breakpoints, the existing breakpoints we have for Sage will suffice for the new changes. Both the sidebar and 3-up layouts can be conformed to the existing 1199-1200px breakpoints. There will be no changes needed at this time so the PR will be closed.

PR #1393, opened 2022-05-09
Ensure consumables don't export CSS

philschanely:
## Description

This PR moves the font size and line height css custom props that were being declared and output in `tokens` files that are included in the `consumables` exports into more appropriate spots. The result is that `consumables` no longer has any output CSS.

This resolves an issue in `kp` where multiples settings for styles were colliding due to unexpected output from `consumables`. 

## Screenshots

CSS output in developer tools:

| Before | After |
|---|---|
| CSS Props output every time consumables are used in addition to initial use with correct theme. | CSS Props only output once for initiaul use in theme. |
| Screen Shot 2022-05-09 at 4 50 30 PM  | Screen Shot 2022-05-09 at 4 48 39 PM |

## Testing in `sage-lib`

See that docs site show now change to font size settings.

1. (**LOW**) Change to where font size and line height CSS props are declared to reduce unexpected output from uses of `_consumables.scss`

## Related

[SAGE-549](https://kajabi.atlassian.net/browse/SAGE-549)

### DIFF
--- a/docs/lib/sage_rails/app/sage_tokens/sage_tokens.rb
+++ b/docs/lib/sage_rails/app/sage_tokens/sage_tokens.rb
@@ -332,7 +332,7 @@ module SageTokens
     "body-xsmall", "body-xsmall-med", "body-xsmall-semi", "body-xsmall-bold",
   ]
 
-  # Aliasses for responsive pairings of size and leading.
+  # Aliases for responsive pairings of size and leading.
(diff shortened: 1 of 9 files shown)


monicawheeler (approved):
Looks great! Would you be able to provide a visual before/after, even if it's code output? Thanks!

philschanely:
@monicawheeler screenshots added.

monkeypox8 (approved):
👍 🔥 LGTM! Font size issue is no longer present when testing with the bridge.

PR #1401, opened 2022-05-11
React breakpoint provider

philschanely:
## Description

This PR adds an experimental breakpoint utility for React uses where contents in `kp` need to change depending on breakpoints. The utility allows such changes to a) sync with official sage breakpoints from within React and b) allows us to track such uses to audit and refactor with better components or tools when they're available.

## Testing in `sage-lib`

N/A

1. (LOW) New React `SageBreakpointProvider`, `SageTokens.BREAKPOINTS` and `SageTokens.BREAKPOINT_QUERIES` for use with provider.


juanviamonte (approved):
Thank you for this. Looks good to me

PR #1409, opened 2022-05-16
Frame: Ensure elements nested within get no default spacing

philschanely:
## Description

This PR adds Frame to the contexts in which elements such as Alert and Form Fields have default margins removed.

1. (**LOW**) Alerts and form fields within Frame no longer have default margins.

## Related

SAGE-542


monkeypox8 (approved):
:shipit:

PR #1416, opened 2022-05-19
Classnames helper

philschanely:
## Description

This PR adds a new helper for working with classname tokens within the system.

Currently we have to do things like this:

```erb
<p class="<%= "#{SageClassnames::TYPE::BODY_SMALL} #{SageClassnames::TYPE_COLOR::CHARCOAL_100}" %>">
  My cool content
</p>
```

But this will allow a little trimming like this:

```erb
<p class="<%= sage_classnames({ type: "BODY_SMALL", color: "CHARCOAL_100" }) %>">
  My cool content
</p>
```

- [x] Adds Rails helper
- [ ] Adds React helper
- [ ] Documents helpers for both platforms

A follow-up will handle populating this helper throughout the docs site and mocks.

## Screenshots

Screen Shot 2022-05-19 at 10 57 48 AM

### DIFF
--- a/docs/lib/sage_rails/app/helpers/sage_classnames_helper.rb
+++ b/docs/lib/sage_rails/app/helpers/sage_classnames_helper.rb
@@ -1,0 +1,110 @@
+module SageClassnamesHelper
+  #
+  # The set of classnames we expose through this helper
+  # pulled from the SageClassnames module
+  #
+  # This leaves `SageClassnames` intact for current uses,
+  # but future deprecate that token and then consolidate definitions
+  # in this file alone.
+  #
+  def sage_classnames_map
+    {
+      # Text color
+      color: module_consts_to_hash(SageClassnames::TYPE_COLORS),
+
+      # Internal grid configurations
+      grid: {
+        CARD: SageClassnames::GRID_CARD,
+        PANEL: SageClassnames::GRID_PANEL,
+      },
+
+      # Internal gap (grid-gap) settings
+      gap: module_consts_to_hash(SageClassnames::GRID_GAP_OPTIONS),
+
+      # Grid templates
+      grid_template: module_consts_to_hash(SageClassnames::GRID_TEMPLATES),
+
+      # Reveal utility classnames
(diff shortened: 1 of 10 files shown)


docs/lib/sage_rails/app/helpers/sage_classnames_helper.rb
```diff
@@ -0,0 +1,110 @@
+module SageClassnamesHelper
+  #
```
philschanely, line 2: Here's the meat of the work here. As noted, this draws from but does not alter the existing `SageClassnames` token module. Hope would be to phase that out in favor of this helper and then consolidate classname definitions. I've tried to make the mapping as automated as practical while also a) not including some classes that are not as useful and b) grouping those that relate such as grids, reveal utils, and text alignment.

docs/lib/sage_rails/app/helpers/sage_misc_helpers.rb
```diff
@@ -0,0 +1,18 @@
+module SageMiscHelpers
+  #
+  # Converts the constants within a module into a has
+  #
```
philschanely, line 4: This may not be needed in the future but was helpful for patching between the existing setup for `SageClassnames` to the hash employed in `sage_classnames_map`

docs/lib/sage_rails/app/helpers/sage_classnames_helper.rb
```diff
@@ -0,0 +1,110 @@
...
+        # Return the only value available if this classname is just a string
+        # and the provided option is truthy
+        if classnames_child.is_a?(String) && value
+          classes.push(classnames_child)
+        else
+          # Output the specified classname value
+          if value.is_a?(Array)
+            # Add all items specified in an array of values
+            value.each do | sub_value |
+              if classnames_child.keys.include? sub_value.to_sym
```
philschanely, line 88: @ju-Skinner is it preferred in Rails to break out into separate methods so as to avoid such deep nesting?

PR #1427, opened 2022-05-25
chore: update ruby dependencies

ju-Skinner:
## Description
Updates the following ruby dependencies, in order to stay in sync with Kajabi products.
* ruby - 2.5.9
* bundler - 2.3.14


.github/workflows/deploy.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     name: Deploy
     runs-on: ubuntu-latest
     container:
-      image: kajabiprodeng/ruby:2.4.9-alpine3.13
+      image: kajabiprodeng/ruby:2.5.9-next-alpine3.13
```
rwc9u, line 9: The `kajabiprodeng/ruby:2.5.9-next-alpine3.13` builds ruby in opt/local/rubies/ruby-2.5.9, so that we could than copy that into the same location in the Dockerfile when we were switching from ruby 2.4.9 to ruby 2.5.9 (i.e. we wanted two rubies in the same image that we could switch between.

I think you can use `ruby:2.5.9-alpine3.13` tag, the community ruby without any customization.

.github/workflows/release-deploy.yml
```diff
@@ -10,10 +10,9 @@ jobs:
     name: Release and Deploy
     runs-on: ubuntu-latest
     container:
-      image: kajabiprodeng/ruby:2.4.9-alpine3.13
+      image: kajabiprodeng/ruby:2.5.9-next-alpine3.13
```
rwc9u, line 13: same: you should be able to use image: `ruby:2.5.9-alpine3.13`

rwc9u (commented):
You might need to add an extra step

```
 - name: Install Bundler 2.3.14
        run: gem install bundler:2.3.14
```

See https://github.com/Kajabi/kajabi-products/pull/23672/files as an example of where I had to do that in our setup.

The other option is to use the default bundler 2 that is installed with ruby 2.5.9 which is 2.2.3.

.github/workflows/deploy.yml
ju-Skinner: Thanks @rwc9u I originally had `ruby:2.5.9-alpine3.13` but changed it once I looking into dockerhub.

rwc9u (approved):

PR #1442, opened 2022-05-31
[SAGE-641] Headings - update color value

QuintonJason:
## Description

- [x] update the color of all heading elements (`h1` - `h6`) to `#040506`
  - currently uses `#202327` inherited from the `body` element

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-05-31 at 2 13 20 PM|Screen Shot 2022-05-31 at 2 13 11 PM|

## Testing in `sage-lib`

Visit the Page Heading element or an element that has a `title` attribute and verify that the color is `#040506`:
- [Page Heading](http://localhost:4000/pages/component/page_heading?tab=preview)
- [Card](http://localhost:4000/pages/component/card?tab=preview)
- [Panel -> Panel Heading](http://localhost:4000/pages/component/panel?tab=preview)

1. (**MEDIUM**) Subtle color update but affects ALL headings, (`h1` - `h6`) so required QE.

## Related

Closes [SAGE-641](https://kajabi.atlassian.net/browse/SAGE-641)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/themes/next/global/_reboot.scss
+++ b/packages/sage-assets/lib/stylesheets/themes/next/global/_reboot.scss
@@ -60,6 +60,7 @@ h5,
 h6 {
   margin-top: 0;
   margin-bottom: 0;
+  color: sage-color(charcoal, 500);
 }
 
 p {
--- a/packages/sage-assets/lib/stylesheets/themes/next/global/_reboot.scss
+++ b/packages/sage-assets/lib/stylesheets/themes/next/global/_reboot.scss
@@ -60,7 +60,6 @@ h5,
 h6 {
   margin-top: 0;
   margin-bottom: 0;
-  color: sage-color(charcoal, 500);
 }
 
 p {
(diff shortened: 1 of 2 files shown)


packages/sage-assets/lib/stylesheets/themes/next/global/_reboot.scss
```diff
@@ -60,6 +60,7 @@ h5,
 h6 {
   margin-top: 0;
   margin-bottom: 0;
+  color: sage-color(charcoal, 500);
```
philschanely, line 63: Consider using `packages/sage-assets/lib/stylesheets/themes/next/core/_typography.scss` instead. There you could add a headings selector group within the `body:not(.sage-excluded), .sage-scoped` selector. This way its a global style yet also prepared to be scoped or excluded when needed.

monkeypox8 (approved):
👍🏼 LGTM, with one caveat: elements using any of the `t-sage-heading-` family of classes have had `color: inherit` set, which in most cases means inheriting from `body`. I'm not certain why that value was redefined rather than relying on the cascade. In any case, if we/design think it's enough of a concern for a follow-up, removing `color: inherit` is probably the best course of action, but requires some extensive testing.

<details>
<summary>inherited color values</summary>

</details>

PR #1444, opened 2022-05-31
Next Modal size

philschanely:
## Description

Per design request (official branch pending) this size adjustment helps to make modals closer to the legacy modal size so that `kp` uses don't appear a little squished. Here we shifted so that the container `md` of `520px` is applied as the inner width (with spacing on outside that) rather than the outer width being this and then constricted by padding. This also updates to latest spec for Modal with smaller padding of `24px`.

## Screenshots

| Before | After |
|---|---|
| Screen Shot 2022-05-31 at 3 54 11 PM | Screen Shot 2022-05-31 at 3 53 32 PM |

## Testing in `sage-lib`

See http://localhost:4000/pages/component/modal?sage_theme=sage_theme_next

1. (**LOW**) Modal size in Next is increased to `520px` from `448px`

## Related

[SAGE 560](https://kajabi.atlassian.net/browse/SAGE-560)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/themes/next/components/_modal.scss
+++ b/packages/sage-assets/lib/stylesheets/themes/next/components/_modal.scss
@@ -5,12 +5,13 @@
 ////
 
 
-$-modal-padding-x: rem(36px);
-$-modal-padding-y: rem(36px);
+$-modal-padding-x: sage-spacer(md);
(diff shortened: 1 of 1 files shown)


pixelflips (approved):
LGTM! 🔥

QuintonJason (approved):

PR #1457, opened 2022-06-06
Hero: Make CTA attributes optionally generate interactive button

philschanely:
## Description

Aligning with latest spec, this PR adjusts Hero to allow both an interactive video CTA button on the artwork as well as just plain artwork based on the `cta_attributes` property.

Similar feature is added in React as well.

## Screenshots

Screen Shot 2022-06-06 at 4 19 22 PM

## Testing in `sage-lib`

See http://localhost:4000/pages/component/hero

1. (**LOW/MEDIUM/HIGH/BREAKING**) Make plain artwork possible within Hero (omit play button effect and CTA)

## Related

[SAGE-630](https://kajabi.atlassian.net/browse/SAGE-630)

### DIFF
--- a/docs/app/views/examples/components/themes/legacy/hero/_preview.html.erb
+++ b/docs/app/views/examples/components/themes/legacy/hero/_preview.html.erb
@@ -1,60 +1,68 @@
+
 <h2 class="t-sage-heading-6">Small</h2>
-<%= sage_component SageCardBlock, {} do %>
-  <%= sage_component SageHero, {
-    alt_text: "Well done alt text",
-    cta_attributes: {
-      "data-js-modaltrigger": "cool-modal",
-      href: "#",
-    },
-    description: "Get early access to new unreleased features and work along side our team by beta testing features before they go live.",
-    image: "card-placeholder-sm.png",
-    size: "small",
-    title: "Be the first to try what Kajabi is building 1",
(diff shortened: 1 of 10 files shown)


docs/app/views/examples/components/themes/legacy/hero/_preview.html.erb
```diff
@@ -1,60 +1,68 @@
...
-    size: "small",
-    title: "Be the first to try what Kajabi is building 1",
-  } %>
-<% end %>
+<%= sage_component SageHero, {
+  alt_text: "Well done alt text",
+  cta_attributes: {
+    "data-js-modaltrigger": "cool-modal",
+    href: "#",
+  },
```
pixelflips, line 8: @philschanely this could be addressed in a follow-up, but the CTA (play icon) and the image seem to behave differently when clicked. I believe they should both popup the modal, but clicking the image does not open the modal right now.

Otherwise looks good so just let me know if you want to address it as a follow-up and I'm ok giving the 👍🏼 on this PR.

philschanely: @pixelflips Yes I'll create a separate ticket for that as it seems some further consideration of the architecture here may be involved. Thanks for catching!

PR #1464, opened 2022-06-07
Adds fallback graphic to replace removed initials in Avatar (Next)

pixelflips:
## Description
React avatar is not loading the default icon when an image is not present.
These updates bring back the check and display the avatar if there is no image.

## Screenshots

|  Before  |  After  |
|--------|--------|
|3c3ce3fb-12d8-4642-967e-6e778f79fcbc|Screen Shot 2022-06-07 at 4 04 53 PM|

## Testing in `sage-lib`
Navigate to http://localhost:4110/?path=/story/sage-avatar--default
Verify default icon displays.

1. (**LOW**) Brings default avatar icon to sage_next avatar when no image present.
   - [ ] People page.

## Related
https://kajabi.atlassian.net/browse/QEINBOX-403

### DIFF
--- a/packages/sage-react/lib/themes/next/Avatar/Avatar.jsx
+++ b/packages/sage-react/lib/themes/next/Avatar/Avatar.jsx
@@ -10,6 +10,7 @@ export const Avatar = ({
   centered,
   color,
   image,
+  useFallbackGraphic,
   size,
   ...rest
 }) => {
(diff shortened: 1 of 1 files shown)


QuintonJason (approved):
Looks and works great

anechol (approved):
Checked Storybook and the Contacts page with the bridge on. Looks good!

PR #1471, opened 2022-06-16
[SAGE-685] Foundations: add tokens and vars for additional border states

QuintonJason:
## Description

- [x] add tokens and vars for additional border states
- [x] update `sage-form-field()` mixin border state values
- [x] ~add docs around use of borders and box-shadows~ [UXD-Handbook#71](https://github.com/Kajabi/UXD-Handbook/pull/71)

**Things to look for while testing:**
- In the form elements, all visual borders should now be `box-shadow`s
- The focus ring **DOES NOT** appear when the form field is hovered

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-06-17 at 8 07 31 AM|Screen Shot 2022-06-17 at 8 05 17 AM|

## Testing in `sage-lib`

Verify the `default`, `:hover`, `:focus`, and `:disabled` states match the spec in the following views:
- Form Input
  - [Rails](http://localhost:4000/pages/component/form_input?tab=preview)
  - [React](http://localhost:4110/?path=/docs/sage-input--default)
- Form Textarea
  - [Rails](http://localhost:4000/pages/component/form_textarea?tab=preview)
  - [React](http://localhost:4110/?path=/docs/sage-textarea--default)
- Form Select
  - [Rails](http://localhost:4000/pages/component/form_select?tab=preview)
  - [React](http://localhost:4110/?path=/docs/sage-select--default)

1. (**HIGH**) Affects high usage form elements. Requires full QE.

## Related

Closes [SAGE-685](https://kajabi.atlassian.net/browse/SAGE-685)

### DIFF
--- a/docs/lib/sage-frontend/stylesheets/docs/themes/next/_token.scss
+++ b/docs/lib/sage-frontend/stylesheets/docs/themes/next/_token.scss
@@ -32,6 +32,14 @@
       content: "#{$token}";
     }
   }
+  &-border-interactive::after {
+    content: "#{sage-border-interactive()}";
+  }
+  @each $name, $token in $sage-borders-interactive {
+    &-border-interactive-#{$name}::after {
+      content: "#{$token}";
+    }
+  }
   &-breakpoint::after {
     content: "#{sage-breakpoint()}";
   }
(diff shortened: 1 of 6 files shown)


monkeypox8 (approved):
👍 LGTM!

philschanely (commented):
Nice work!

Fine as-is, but to me this begins to feel like a token map for which we'd follow the pattern set up for other such tokens (defined in their own file in `tokens/`; define a function for using them, etc.) Any reason we wouldn't go that direction with this map?

Also, do you think it would be worthwhile to also make a note in the doc block above the map regarding when/why these should be used over `sage-border`s? AKA does a future dev know why `box-shadow` vs. standard `border` should be used for a future element's border? And do these also need to be kept in sync with `tokens/_border.scss` such that a cross-reference note between them would be helpful?

PR #1485, opened 2022-06-27
Fix boolean for hasCustomWidth in Frame

darrellrivera:
## Description
I was trying to use the `width` prop on frame with the value of "fill", but I noticed that it still kept adding the class name of `sage-frame--width-custom`, then I noticed that the conditional was using the wrong prop

1. (**LOW**) Fixes the `width` prop on Frame to add the correct class
   - [ ] Add `width` prop with a value of "fill" to a frame component and make sure now that the class matches (`sage-frame--width-fill` and not ``sage-frame--width-custom`)

### DIFF
--- a/packages/sage-react/lib/themes/legacy/Frame/Frame.jsx
+++ b/packages/sage-react/lib/themes/legacy/Frame/Frame.jsx
@@ -34,7 +34,7 @@ export const Frame = ({
   const hasCustomBackground = background
     && !Object.values(SageTokens.COLOR_SLIDERS).includes(background);
   const hasCustomWidth = width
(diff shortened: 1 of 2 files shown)


ju-Skinner:
thanks @darrellrivera...I added another commit to cover `legacy`. Fixed the commit message as well, since we use [conventional commtis](https://www.conventionalcommits.org/en/v1.0.0/#summary)

ju-Skinner (approved):

anechol (approved):
🚀

pixelflips (approved):
LGTM! Nice catch! 👍🏼

PR #1490, opened 2022-06-29
Only specify minimum Node version for packages

systemist:
## Description
Using strict `engines` means consuming applications can't upgrade their stuff, but it makes sense to specify a minimum Node version, so this changes the 5 packages meant to be used by other applications to only specify a minimum Node version in their `engines` and updates the root package to be more specific so developers on this project can specify specific tools to use in development.

## Screenshots
N/A

## Testing in `sage-lib`
No testing required, I don't think.

No testing required, I don't think.

## Related
N/A

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -17,9 +17,7 @@
     "webpack:server": "./bin/webpack-dev-server"
   },
   "engines": {
-    "node": "14.x",
-    "npm": "6.x",
-    "yarn": "1.x"
+    "node": ">=14.0.0"
   },
   "publishConfig": {
     "registry": "https://npm.pkg.github.com/"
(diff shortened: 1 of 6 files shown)


ju-Skinner (approved):
Sorry, don't know how this slipped through the cracks for reviewing. 

I believe this should be fine as well. Honestly, I don't think we need the engines section. From what I can tell we only used or have it, because it was want KP needed to work with the Docker images. 

In our case, we shouldn't be confined to the version constraints of KP, as long as the library can be added as a dependency there things should be fine.

For our reference, should we have to come back to this at a later date. lol
https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines

ju-Skinner:
Closing this. There will be changes made but in a follow up PR once we move to AWS.

PR #1493, opened 2022-06-30
[ SAGE-666] 👿 Sage search reset button misaligned in Toolbar

pixelflips:
## Description
When text has been entered in a Sage Search field nested within a Sage Toolbar, the “reset” button does not appear centered within the input.

These updates correct the alignment.

## Screenshots
|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-06-30 at 12 46 10 PM|Screen Shot 2022-06-30 at 12 47 13 PM|

## Testing in `sage-lib`

- Enable `sage_next` toggle
- Navigate to [Toolbar](http://localhost:4000/pages/component/toolbar?tab=preview)
- Enter text within the search input within a `SageToolbarGroup`
- Verify clear button aligns as expected.

1. (**LOW**) Adjust the alignment of the clear search button within a `SageToolbarGroup`. No effect on KP as the `SageToolbarGroup` component isn't yet used in KP.

## Related
https://kajabi.atlassian.net/browse/SAGE-666

### DIFF
--- a/packages/sage-assets/lib/stylesheets/themes/next/components/_search.scss
+++ b/packages/sage-assets/lib/stylesheets/themes/next/components/_search.scss
@@ -214,4 +214,12 @@ $-search-icon: "::before";
   .sage-search--has-text & {
     visibility: visible;
   }
+
+  .sage-search & {
+    .sage-toolbar__group & {
+      &:last-child {
+        align-self: center;
+      }
+    }
+  }
 }


packages/sage-assets/lib/stylesheets/themes/next/components/_search.scss
```diff
@@ -214,4 +214,13 @@ $-search-icon: "::before";
   .sage-search--has-text & {
     visibility: visible;
   }
+  /* stylelint-disable max-nesting-depth */
+  .sage-search & {
+    .sage-toolbar__group & {
+      &:last-child {
```
philschanely, line 220:
```suggestion
  .sage-toolbar__group .sage-search &:last-child {
```

If this is what you're after is it as effective? Just saying we may not need to use the nested selector per se.

philschanely (approved):
One non-blocking suggestion but devlishly good otherwise.

packages/sage-assets/lib/stylesheets/themes/next/components/_search.scss
pixelflips: @philschanely I am cool with that and can remove the stylelint-disable as well. 🙌🏼 - Incoming update!

monkeypox8 (approved):
👍 ✝️

PR #1498, opened 2022-07-01
[SAGE-710] Update Form Select option attributes

monkeypox8:
## Description
Updates `SageFormSelect` to permit passing `selected` and `disabled` attributes to individual options.

The `selected` change affects Rails only. This is not applicable to the React component, since `value` can already be used to pre-select an item.

- Also corrects several errors in Storybook caused by missing and invalid props
- Documentation updated in Rails and React

## Screenshots
No visual changes

### Rails
1. Navigate to the [Form Select](http://localhost:4000/pages/component/form_select) documentation page in Next or Legacy
2. Locate that the second example, "Pets"
3. Confirm that (None Specified) `option` is selected on page load
4. Confirm that the "Spider" `option` is disabled and cannot be selected 

### React
1. Navigate to [Select in Storybook](http://localhost:4110/?path=/docs/sage-select--default)
2. Verify that the "Disabled Select Field" example has "Option 3" selected on page load
3. Confirm that the "Second Option" in the "Select With Option Disabled" is disabled and cannot be selected

1. (**HIGH**) New attributes for `SageFormSelect` are optional. No expected impact on current use. 

## Related
[SAGE-710](https://kajabi.atlassian.net/browse/SAGE-710)
[SAGE-600](https://kajabi.atlassian.net/browse/SAGE-600)
[SAGE-633](https://github.com/Kajabi/sage-lib/pull/1503)

### DIFF
--- a/docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_form_select.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_form_select.html.erb
@@ -25,5 +25,7 @@
   </select>
   <label class="sage-select__label" for="<%= selectID %>"><%= component.name %></label>
   <i class="sage-select__arrow" aria-hidden="true"></i>
-  <div class="sage-select__message"><%= component.message %></div>
+  <% if component.message.present? %>
+    <div class="sage-select__message"><%= component.message %></div>
+  <% end %>
 </div>
--- a/docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_form_select.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_form_select.html.erb
@@ -14,7 +14,13 @@
     <%= "disabled" if component.disabled %>
   >
     <% component.select_options.each do |option| %>
-      <option value="<%= option[:value] %>"><%= option[:text] %></option>
+      <option
+        value="<%= option[:value] %>"
+        <%= option[:selected] && !option[:disabled] ? " selected=true" : nil -%>
+        <%= option[:disabled] ? " disabled" : nil -%>
+      >
+        <%= option[:text] %>
+      </option>
     <% end %>
   </select>
   <label class="sage-select__label" for="<%= selectID %>"><%= component.name %></label>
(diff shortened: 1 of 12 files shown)


docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_form_select.html.erb
```diff
@@ -14,7 +14,13 @@
     <%= "disabled" if component.disabled %>
   >
     <% component.select_options.each do |option| %>
-      <option value="<%= option[:value] %>"><%= option[:text] %></option>
+      <option
+        value="<%= option[:value] %>"
+        <%= option[:selected] && !option[:disabled] ? " selected=true" : nil -%>
```
pixelflips, line 19: If I am remembering correctly, even though selected is a boolean attribute, I think it should be `selected="selected"`?

pixelflips (approved):
LGTM!! 🔥 - One tiny non-blocking comment on the selected attribute, otherwise no concerns here. Nice update!

docs/lib/sage_rails/app/views/sage_components/themes/legacy/_sage_form_select.html.erb
monkeypox8: Good question! Either way is usually fine, because the browser will interpret an attribute string as truthy. The actual spec [calls for a Boolean](https://html.spec.whatwg.org/multipage/form-elements.html#the-option-element)

pixelflips: Gotcha...and the browser does behave as expected, but the spec also says [true and false are not allowed](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes) - still not a big deal either way.

monkeypox8: Yup! Good call, I'll adjust 👍🏼

packages/sage-react/lib/themes/legacy/Select/Select.jsx
```diff
@@ -77,14 +85,14 @@ Select.defaultProps = {
   message: null,
   onChange: (evt) => evt,
   options: [],
-  value: null,
+  value: '',
```
monkeypox8, line 88: Correction here due to the error thrown in the console for a controlled component

PR #1503, opened 2022-07-06
fix(disabled states): audit, fix, and add stories for disabled states

goodwinchris:
## Description
This PR audits, fixes, and adds stories for all disabled states

## Screenshots

|  Before  |  After  |
|--------|--------|

## Testing in `sage-lib`
[Rails foo](http://localhost:4000/pages/component/form_input?tab=preview)
[React foo](http://localhost:4100/?path=/docs/sage-input--default)

(LOW) Adjusts form `select`'s `option`s to support disabled options, and adds disabled stories for a few undocumented states

## Related
[SAGE-633 Form element: disabled use should be consistent](https://kajabi.atlassian.net/browse/SAGE-633)


ju-Skinner (approved):
LGTM...My question does NOT block my approval.

docs/app/views/examples/components/themes/next/form_select/_preview.html.erb
```diff
@@ -36,7 +36,7 @@
 } %>
 
 <%= sage_component SageFormSelect, {
-  name: "Pets",
+  name: "Pets (with disabled option)",
```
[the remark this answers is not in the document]
goodwinchris: I can certainly add one, I figured since it was just one option that we could just disable it in the same story. Happy to do whatever the team thinks is best though!

packages/sage-react/lib/themes/next/Select/Select.jsx
```diff
@@ -40,17 +40,22 @@ export const Select = ({
 
         {options.map((option, i) => {
           let optionLabel,
-            optionValue;
+            optionValue,
+            optionDisabled;
           if (typeof option === 'string') {
             optionLabel = option;
             optionValue = option;
+            // optionDisabled = option;
```
QuintonJason, line 48: Is this uncommented code still needed?

goodwinchris: Good call, pulling out now

monkeypox8 (approved):
👍🏼 LGTM!

Not a blocker: thoughts on providing a toggle-able `disabled` option for the default story? If so, there would be a couple of proptypes in `Textarea.jsx` to update as well.

QuintonJason (approved):
looking good

packages/sage-react/lib/Textarea/Textarea.story.jsx
```diff
@@ -15,6 +15,14 @@ const Template = (args) => <Textarea {...args} />;
 
 export const Default = Template.bind({});
```
monkeypox8, line 16: Not a blocker, but there's a warning in the console because we aren't currently passing an `id` in the arguments for the default example.

monkeypox8 (approved):
👍 LGTM! One minor cleanup item to consider for a warning I'm seeing in the console (not due to this PR).

packages/sage-react/lib/Textarea/Textarea.story.jsx
```diff
@@ -12,8 +12,17 @@ export default {
 };
 
 const Template = (args) => <Textarea {...args} />;
-
 export const Default = Template.bind({});
+Default.args = {
+  id: 'field-1',
```
monkeypox8, line 17: This works well. You could also consider adding the `id` to the default args object on line 7-11; keeps it a bit more streamlined.

PR #1535, opened 2022-07-27
[SAGE-765] - fix(sage-react): input was using an invalid prop 'inputType'

ju-Skinner:
Prop should have been 'type'

## Description
It was called out that the React component `Input` was raising a warning about an invalid prop name `inputType`. 

## Screenshots

|  Before  |  After  |
|--------|--------|
|image||

## Testing in `sage-lib`
1. Navigate to Storybook `input-email`...[here](http://localhost:4100/?path=/docs/sage-input--input-email)
2. Enter an invalid value and you should be greeted with a tooltip similar to the one below. 
image
3. Should not see any warnings related to Hero or `suppressDefaultClass` in the console.


packages/sage-react/lib/Hero/Hero.story.jsx
```diff
@@ -23,14 +23,14 @@ export default {
         <Button
           className=""
           color="primary"
-          raised={false}
+        // raised={false}
```
pixelflips, line 26: Could this be completely removed?

packages/sage-react/lib/Hero/Hero.story.jsx
```diff
@@ -23,14 +23,14 @@ export default {
...
         >
           Get started
         </Button>
         <Button
           className=""
           color="primary"
-          raised={false}
+          // raised={false}
```
pixelflips, line 33: Could this be completely removed?

philschanely (approved):
Fix looks good. 🙌 

Not a blocker but what do you think about moving the quotation fixes (unrelated to this?) to a different PR?

pixelflips (approved):
Minor comments but not show-stoppers. LGTM! 🔥 - Thanks for the additional syntax cleanup as well! 🙌🏼

packages/sage-react/lib/Hero/Hero.story.jsx
ju-Skinner: Actually, I forgot about that change...That actually is something that needs a discussion.  I don't see it being used anywhere, so I was wondering what was the intent or use case.

pixelflips: I believe the raised variant of buttons was indeed removed, but might be worth double checking.

ju-Skinner:
> Fix looks good. 🙌
> 
> Not a blocker but what do you think about moving the quotation fixes (unrelated to this?) to a different PR?

@philschanely Thanks, the quotations are an artifact of the deprecation changes for the import statement. Based on my eslint and prettier configs is what changed the single to double quotes. <- Knowledge sharing not justifying. 

The main change for those were going from `'@storybook/addon-docs/blocks'` to `"@storybook/addon-docs"`.

ju-Skinner:
@philschanely Took your advice and removed all unrelated changes. I will create a separate PR to remove warning and deprecations.

QuintonJason (approved):

PR #1537, opened 2022-07-28
[Sage-761] Card - Add Interactivity

anechol:
## Description

Adds new Card prop for interactivity.

The main/primary link or button in the Card should make the entire card clickable. When a secondary link is present, the `z-index` should elevate it above the primary link's hitbox to make it clickable.

[Figma](https://www.figma.com/file/sMbtLUHSt2vfKgKjyQ3052/Sage-components?node-id=1667%3A23273)

## Screenshots

https://user-images.githubusercontent.com/14791307/185470449-e009f561-404a-4bdf-9e11-8cd27006caf3.mov

## Testing in `sage-lib`

- View [Card](http://localhost:4000/pages/component/card?tab=preview#)
- Check that entire Interactive Card is clickable with proper focus and hover states
- Add a second SageLink to `docs/app/views/examples/components/card/_preview.html.erb` line:56 and change style to `secondary`
- Ensure that secondary link is clickable

(**LOW**) Adds new Card prop for interactivity.

## Related

[Sage-761](https://kajabi.atlassian.net/browse/SAGE-761)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_card.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_card.scss
@@ -46,6 +46,17 @@
 
 .sage-card--interactive {
   border: 0;
+
+  a:not(.sage-link--primary),
+  button:not(.sage-link--primary) {
+    z-index: sage-z-index(default, 2);
+    position: relative;
+
+    &::before,
+    &:hover::before {
+      box-shadow: none;
+    }
+  }
 }
 
 .sage-card__footer {
--- a/packages/sage-assets/lib/stylesheets/components/_card.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_card.scss
@@ -47,8 +47,18 @@
 .sage-card--interactive {
   border: 0;
 
-  a:not(.sage-link--primary),
-  button:not(.sage-link--primary) {
+  .sage-link.sage-card__link--primary {
+    // Remove when Link component is updated with new styles
+    color: sage-color(charcoal, 400);
+    text-decoration: none;
+
+    &:hover {
+      color: sage-color(charcoal, 500);
+    }
+  }
+
+  a:not(.sage-card__link--primary),
+  button:not(.sage-card__link--primary) {
     z-index: sage-z-index(default, 2);
     position: relative;
 
--- a/packages/sage-assets/lib/stylesheets/components/_card.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_card.scss
@@ -12,6 +12,24 @@
   width: 100%;
 }
 
+.sage-card--border-dashed {
+  border-color: sage-color(white);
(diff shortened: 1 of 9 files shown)


packages/sage-assets/lib/stylesheets/components/_card.scss
```diff
@@ -26,22 +44,19 @@
...
-  padding-top: 0;
-}
+.sage-card--interactive {
+  border: 0;
 
-.sage-card--clear-padding-bottom,
-.sage-card--clear-padding-both {
-  padding-bottom: 0;
-}
+  a:not(.sage-link--primary),
```
QuintonJason, line 50: Can the class be something tied to the card, like `.sage-card__link--primary`? Currently `sage-link--primary` can be used outside of the card and would reduce any edge cases.

packages/sage-assets/lib/stylesheets/components/_link.scss
```diff
@@ -98,6 +99,45 @@ $-link-base-styles: (
...
+
+    &:invalid::before {
+      box-shadow: sage-border-interactive(error);
+    }
+
+    &:focus:invalid::before {
+      box-shadow: sage-border-interactive(error-focus);
+    }
+
+    &:focus::after {
```
QuintonJason, line 126: This may also need to be a `box-shadow` as this is currently showing two borders.

QuintonJason (changes requested):
Screen_Shot_2022-08-15_at_10_49_38_AM
1. Design may need to provide feedback, but the primary target in an interactive card looks similar to the primary target for a Catalog Item or a Table Row. The treatment seems weight related.
2. When the Card is `:active` and `:focus`ed two rings are showing. Does an active card need a focus state as well? This may be something design has to weigh in on

packages/sage-assets/lib/stylesheets/components/_link.scss
anechol: Addressed this. Removed the shadow on the `after` and placed it on the `before`.

anechol:
Updated to make the Sage Card primary action resemble primary links in tables and lists.

goodwinchris:
I see your most recent comment about updating link styles. Just want to confirm (I'm still learning here) the expected visual is how its shown in my attached screenshot and not what's shown in the PR description?
CleanShot 2022-08-16 at 14 32 51@2x

anechol:
> I see your most recent comment about updating link styles. Just want to confirm (I'm still learning here) the expected visual is how its shown in my attached screenshot and not what's shown in the PR description? CleanShot 2022-08-16 at 14 32 51@2x

Thanks for calling that out @goodwinchris, I've replaced the video in the description.

ju-Skinner (approved):
LGTM

QuintonJason (approved):

PR #1539, opened 2022-08-01
[CHORE] - Some code cleanup for console warnings

ju-Skinner:
## Description
There have been numerous warnings that have been showing up in our console warnings. 
* suppressDialogClass: will no longer be passed to an html element (button)
* raised: has been deprecated with the release of Sage Next (v5)
* Storybook: Deprecation of library `@storybook/addon-docs/blocks`, replaced with `@storybook/addon-docs`

1. Turn on the bridge
2. Confirm that there are no errors, but warnings will still be preset with Buttons that still may be using `raised` prop. This could be in both rails and react.
 - CreationWizard uses the React button with the `raised` prop. <- This will still display warning in console,  due to the use in KP.
 - Affiliate  Users index pages uses the Rails button.

### DIFF
--- a/docs/lib/sage_rails/app/sage_tokens/sage_schemas.rb
+++ b/docs/lib/sage_rails/app/sage_tokens/sage_schemas.rb
@@ -51,7 +51,6 @@ module SageSchemas
     value: [:optional, String],
     # TODO: Deprecations in Next
     subtle: [:optional, NilClass, TrueClass],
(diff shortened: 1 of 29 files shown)


pixelflips (approved):
Thanks for the clarification on the warning. I checked with the bridge enabled and all LGTM! 👍🏼

philschanely (approved):

PR #1541, opened 2022-08-01
List: Sortable list interactions update

philschanely:
## Description

This PR updates styling for interactions with sortable List.

## Screenshots

[Figma spec](https://www.figma.com/file/sMbtLUHSt2vfKgKjyQ3052/Sage-components?node-id=12501%3A23675)

| Before | After |
|---|---|
|  |  |

## Testing in `sage-lib`

- See Rails http://localhost:4000/pages/component/list?tab=preview
- See React http://localhost:4100/?path=/story/sage-list--sortable-list

1. (LOW) Updates interactive styling for the List (sortable) component in Rails and React.

## Related

[SAGE-535](https://kajabi.atlassian.net/browse/SAGE-535)


philschanely:
FFR: Scheduled a design review with @cameronsimony for later today.

docs/lib/sage_rails/app/views/sage_components/_sage_list_item.html.erb
```diff
@@ -19,7 +19,12 @@ end
 >
   <% if component.sortable %>
     <div class="sage-list__item-sortable-handle">
-      <%= sage_component SageIcon, { icon: "handle" } %>
+      <%= sage_component SageButton, {
```
pixelflips, line 22: Curious about the icon vs. button here? Are there any benefits? I see we do get the focus state, but would love to understand further.

philschanely: yes, I'm going to check on this with Cam during design review. IconButton was used in the Figma specs and think there is some interest in revising it so that it can only be reordered by dragging on this affordance. Def need Cam to look at it in the browser and ensure he's happy with it.

pixelflips (approved):
Looks and works great! LGTM! Nice work! 👍🏼

docs/lib/sage_rails/app/views/sage_components/_sage_list_item.html.erb
pixelflips: @philschanely would love to hear how the design review goes. I am also curious about the grab vs. move cursor, but my thoughts may become mute depending on what Cam comes back with.

goodwinchris:
Actually clicking and dragging on the drag icon doesn't work for me in FF

monkeypox8:
> Actually clicking and dragging on the drag icon doesn't work for me in FF
> 
> Does work in Chrome

As Chris noted, the `button` in Firefox (and to a lesser extent, Safari when "holding" the button for a second or so) does not propagate the parent list item's event. 

- Whether or not the icon ends up being the drag "trigger" rather than the list item, we should consider changing it back an icon rather than a `button` element. This would restore the draggable behavior in FF/Safari and also be somewhat less confusing to AT users, since the label indicates that the button allows sorting
- Related to that, the accessible capabilities of sortable [appear to be lacking](https://github.com/SortableJS/Sortable/issues/1951). I'm wondering if we'll want to replace it with a custom solution using the draggable API as a future follow-up. As it currently stands, the functionality is not keyboard accessible

docs/lib/sage_rails/app/views/sage_components/_sage_list_item.html.erb
monkeypox8: > Actually clicking and dragging on the drag icon doesn't work for me in FF
> 
> Does work in Chrome

As Chris noted, the `button` in Firefox (and to a lesser extent, Safari when "holding" the button for a second or so) does not propagate the parent list item's event. 

Whether or not the icon ends up being the drag "trigger" rather than the list item, we should consider changing it back an icon rather than a `button` element. This would restore the draggable behavior in FF/Safari and also be somewhat less confusing to AT users, since the label indicates that the button allows sorting, but it's only possible with a mouse.

monkeypox8 (approved):
👍🏼 As it currently stands, the functionality is not keyboard accessible, but this is part of the existing behavior, and not necessarily a blocker.

On a related note, the accessible capabilities of sortable [appear to be lacking](https://github.com/SortableJS/Sortable/issues/1951). I'm wondering if we can replace it with a custom solution using the draggable API for a future follow-up. Or maybe leverage [something like this as a shortcut](https://github.com/schne324/dragon-drop).

Last thing… the `sortable_update_url` config in the Rails preview throws an error when `onEnd` fires. If we're not storing that we can probably just remove it for now.

cameronsimony (approved):
paired with @philschanely for design QA on this & it looks great! a big 👍 from me!

philschanely:
@goodwinchris does this work better/consistently for you now?

@monkeypox8 are the changes here to your liking as well?

philschanely:
@monkeypox8 @pixelflips requesting a fresh review after having updated this PR.

pixelflips (commented):
A couple of minor non-blocking comments. In Safari, when dragging an item in the top section, items in the bottom section get highlighted/a background color applied.

https://user-images.githubusercontent.com/1175111/185443370-23d7ca6d-a613-4ffe-a76d-5920e5e2f0cf.mov

pixelflips (approved):
Rechecked and am no longer seeing the ID issue as we discussed. LGTM! 👍🏼

monkeypox8 (approved):
👍 LGTM, nice work! Thanks for getting this… sorted out 😝

docs/lib/sage_rails/app/views/sage_components/_sage_list.html.erb
```diff
@@ -5,16 +5,16 @@ tag = component.tag.present? ? component.tag : "ul"
   class="
     sage-list
     <%= component.generated_css_classes %>
-    <%= "sage-list--hide-first-border" if component.hide_first_border %>
+    <%= "sage-list--all-draggable" if component.sortable_configs && component.sortable_configs[:handle] == false %>
```
ju-Skinner, line 8: ruby tip you can change this to `component.dig(:sortable_configs, :handle).blank?`. This will account for it either being `nil` or `false`'. Handle being set to `true` will pass the `blank?` check.

```
[7] pry(main)> component = {sortable_configs: {handle: false}}
=> {:sortable_configs=>{:handle=>false}}
[8] pry(main)> component.dig(:sortable_configs, :handle).blank?
=> true
[9] pry(main)> component.dig(:sortable_configs, :handle)
=> false
[10] pry(main)> component = {sortable_configs: {handle: true}}
=> {:sortable_configs=>{:handle=>true}}
[11] pry(main)> component.dig(:sortable_configs, :handle).blank?
=> false
[12] pry(main)>
```

ju-Skinner (approved):

ju-Skinner:
Lint failure 

```
packages/sage-assets/lib/stylesheets/components/_table.scss
 35:5  ⚠  Expected indentation of 2 spaces   indentation
 ```

**Comment by philschanely** [comment]
@monkeypox8 @pixelflips and @ju-Skinner since last review I made the following updates to further hone how this component works:

- I abandoned the `sortableConfigs/sortable_configs` prop that was initially meant to allow the user to pass configs down to the `SortableJS` inside the component. On React, I instead surfaced only the props we intend to expose at this time: `setList`, `onStart`, and `onEnd`. On Rails this is not practical at this time (and not attempted in previous Sortable component either).
- I added a `dragHandleType/drag_handle_type` prop that allows user to choose between `"default"` and `"row"`. The default is to only allow the end user to sort/drag by the drag handle, but `"row"` makes the whole row draggable.
- I moved `sortable<boolean>` as a prop up from the children Items to the parent List.

**Review by pixelflips — APPROVED** [review]
Rechecked all main browsers and all still LGTM! 👍🏼

File: docs/app/views/examples/components/list/_preview.html.erb [hunk 3]
```diff
@@ -55,41 +55,76 @@ sample_products = [
...
+
+<h3 class="<%= SageClassnames::TYPE::HEADING_6 %>">Sortable configuration</h3>
+<%= md("
+Sortable lists can be created with `SageList` by adding `sortable: true` and a `sortable_resource`.
+You can also pass a `sortable_update_url` for sorting callbacks to items
+that will be called after the user finishes sorting an item.
+", use_sage_type: true) %>
+
+<%= sage_component SageList, {
+  sortable: true,
```
monkeypox8, line 79: For the sortable examples, thoughts on adding `tag: "ol"` to drive home the ordinal values?

I suppose we could also be more heavy-handed and change the default to an `ol` when setting `sortable: true` in the component, but that's beyond the scope of this work.

packages/sage-react/lib/List/List.story.jsx
```diff
@@ -8,7 +9,11 @@ import {
...
   title: 'Sage/List',
   component: List,
-  argTypes: {},
+  argTypes: {
+    ...selectArgs({
+      dragHandleType: List.DRAG_HANDLE_TYPES,
+    }),
+  },
   args: {
     items: [],
```
monkeypox8, line 18: I'm getting an error when attempting to toggle the base list to "sortable" in Storybook. Seems it's because we're not defining `setList` for the default config?

monkeypox8 (approved):
👍🏼 LGTM! Two final items to consider; both are documentation-related so not considered as blockers for the merge.

Noting here also that I'm not able to get the bridge running due to a `bundler` issue.

packages/sage-react/lib/List/List.story.jsx
philschanely: That's right. The setList needs to be set as well so it's not as straightforward to switch props in storybook.

pixelflips (commented):
On drag-n-drop I am seeing the following error in the console. I only point it out as I am seeing this popping up in [DataDog UX Monitory Errors](https://app.datadoghq.com/rum/explorer?query=%40issue.id%3A110c80f2-ca7b-11eb-aa20-da7ad0900002%20%40type%3Aerror&cols=&tab=error&from_ts=1660000432258&to_ts=1661210032258&live=false) on the KP side as well.

Screen Shot 2022-08-22 at 4 15 01 PM


Definitely not a blocker, since it looks like it's been hanging around for a while.

pixelflips (commented):
It may or may not be related to these updates, but I seem to be getting some odd behavior when clicking the dropdown in the fully draggable section. After opening the dropdown, clicking out of it kinda blinks and doesn't close smoothly if at all.

https://user-images.githubusercontent.com/1175111/186040123-e50a16bf-4415-4f57-b94e-416af60b3743.mov

docs/app/views/examples/components/list/_preview.html.erb
```diff
@@ -55,41 +55,76 @@ sample_products = [
...
       sortable_update_url: "#sortable-update-url?item=#{item[:id]}",
     }
-  }
+  },
+} %>
+
+<h3 class="<%= SageClassnames::TYPE::HEADING_6 %>">Fully draggable row</h3>
+<%= md("
+By default only the drag handle is active for dragging/sorting a row.
+However, `draggable_anywhere` can be set to `true` in order to allow the whole row to be draggable instead.
```
ju-Skinner, line 93: I believe you need to change `draggable_anywhere` to "`drag_handle_type` can be set to `row`" here.

docs/lib/sage_rails/app/views/sage_components/_sage_list.html.erb
```diff
@@ -1,11 +1,13 @@
 <%
 tag = component.tag.present? ? component.tag : "ul"
+drag_handle_type = component.tag.present? ? component.tag : "default"
```
ju-Skinner, line 3: This may have been an oversight or a misunderstanding from me.  `drag_handle_type` accepts two options based on the code above `default` and `row`. So if  `component.tag` is anything other than those two values, what would be the expectation here?

Is this `tag: [:optional, NilClass, Set.new(["ul", "ol"])],` the same `component.tag` values?

packages/sage-react/lib/List/List.jsx
```diff
@@ -60,42 +85,30 @@ export const List = ({
...
 
 List.defaultProps = {
   children: null,
   className: null,
-  hideFirstBorder: false,
   items: [],
   itemRenderer: null,
-  sortableConfigs: null,
+  dragHandleType: List.DRAG_HANDLE_TYPES.DEFAULT,
+  onEnd: null,
```
ju-Skinner, line 96: for functions it is recommended to use `noop` vs `null`

ju-Skinner (changes requested):
see comments. the request changes is for documentation purposes in `docs/app/views/examples/components/list/_preview.html.erb`.  There are a couple of other callouts as well.

packages/sage-react/lib/List/List.jsx
philschanely: TIL!

philschanely: Hmm... validation does not like that... is it an accepted keyword in `js`?

philschanely: I set these up as `() => null` but lmk if that's not what you were looking for.

ju-Skinner: Sorry, I made a bad assumption. So [noop](https://www.geeksforgeeks.org/lodash-_-noop-method/) is a lodash function, that essentially returns `undefined` which is completely different than the value of `null`. 

Examples taken from console....created a `noop` function to mimic lodash to show difference. 
```
null === undefined
false
function noop() { return undefined }
undefined
null === noop()
false
undefined === noop()
true
```

ju-Skinner (changes requested):

docs/lib/sage_rails/app/views/sage_components/_sage_list.html.erb
```diff
@@ -1,13 +1,16 @@
 <%
-tag = component.tag.present? ? component.tag : "ul"
-drag_handle_type = component.tag.present? ? component.tag : "default"
+tag = component.sortable ? "ol" : "ul"
+if component.tag.present?
```
ju-Skinner, line 3: Note: tag = component.tag || tag

ju-Skinner (approved):
Great job. :shipit:

PR #1561, opened 2022-08-17
Pagination: (React) Adjust default value of `pageCount` for improved use

philschanely:
## Description

Per SAGE-751 description, React Pagination defaults `pageCount` to `1` which makes page calculation moot. By setting it to default of `null` it becomes dependent on the user to provide `pageCount` for a specific number of pages or to provide `itemsTotalCount` and `pageSize` for the page count to be calculated better.

## Screenshots

No visual changes.

## Testing in `sage-lib`

See the following for improved stories showing the two configuration options.

- http://localhost:4100/?path=/story/sage-pagination--from-page-count 
- http://localhost:4100/?path=/story/sage-pagination--from-calculated-pages

1. (**LOW**) Adjust default value for `pageCount` in React Pagination for easier configuration. Also audited existing uses of Pagination to ensure no breaking changes or impact on uses. 👍 

## Related

https://kajabi.atlassian.net/browse/SAGE-751

### DIFF
--- a/packages/sage-react/lib/Pagination/Pagination.jsx
+++ b/packages/sage-react/lib/Pagination/Pagination.jsx
@@ -29,17 +29,17 @@ export const Pagination = ({
   pageSize,
   pageURLFn,
 }) => {
-  // Get a safe page count if possible:
+  // Get a safe page count:
   let localPageCount;
   switch (true) {
-    // We favor `pageCount` if it is not null.
-    case pageCount:
+    // We favor `pageCount` if it is a positive value.
+    case pageCount > 0:
       localPageCount = pageCount;
       break;
 
-    // Or, if both `itemsTotalCount` and `pageSize` are not null
+    // Or, if both `itemsTotalCount` and `pageSize` are positive values
     // we can calculate the page count.
-    case itemsTotalCount && pageSize:
+    case itemsTotalCount > 0 && pageSize > 0:
       localPageCount = Math.ceil(itemsTotalCount / pageSize);
       break;
 
--- a/packages/sage-react/lib/Pagination/Pagination.jsx
+++ b/packages/sage-react/lib/Pagination/Pagination.jsx
@@ -212,7 +212,7 @@ Pagination.defaultProps = {
   labels: null,
   maxPageButtons: 4,
   onClickPage: null,
-  pageCount: 1,
+  pageCount: null,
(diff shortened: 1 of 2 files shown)


QuintonJason (changes requested):
Please consider rebasing as this contains other PR work

pixelflips (approved):

packages/sage-react/lib/Pagination/Pagination.jsx
```diff
@@ -212,7 +212,7 @@ Pagination.defaultProps = {
   labels: null,
   maxPageButtons: 4,
   onClickPage: null,
-  pageCount: 1,
+  pageCount: null,
```
ju-Skinner, line 215: Based on our conversation via huddle...It looks like we can clean up the logic that uses the property. With programming there are two ways to  clean this up, neither are wrong.

**first**
```
  let localPageCount = canCalculatePageRange 
    ? Math.ceil(itemsTotalCount / pageSize) 
    : 1;
  if (pageCount) {
    localPageCount = pageCount;
  }
  ```
**second**
```
  switch (true) {
    case pageCount:
      localPageCount = pageCount;
      break;
    case canCalculatePageRange:
      localPageCount = Math.ceil(itemsTotalCount / pageSize);
      break;
    default:
      localPageCount = 1;
      break;
  }
  ```

packages/sage-react/lib/Pagination/Pagination.jsx
```diff
@@ -29,14 +29,24 @@ export const Pagination = ({
...
+    // Or, if both `itemsTotalCount` and `pageSize` are positive values
+    // we can calculate the page count.
+    case itemsTotalCount > 0 && pageSize > 0:
+      localPageCount = Math.ceil(itemsTotalCount / pageSize);
+      break;
+
+    // But if all else fails we set page count to `1` .
+    default:
+      localPageCount = 1;
+      break;
```
ju-Skinner, line 49: **Note**:  Since the primary expression `switch(true)` is always true, the case statements themselves need to be an expression or act as an `if` statement in other words. 

excerpt taken from [switch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch)
>switch (true) pattern as an alternative to if...else

ju-Skinner (approved):

PR #1562, opened 2022-08-18
PageHeading: Adjust React proptypes

philschanely:
## Description

This PR adjusts the proptypes for actionItems and toolbarItems to be just nodes rather than an array of nodes in keeping with a more conventional approach to such props in React.

## Screenshots

No visual impact.

## Testing in `sage-lib`

See http://localhost:4100/?path=/story/sage-pageheading--default and other related stories to see no visual change in output.

1. (**LOW**) Adjust proptype setting for toolbarItems and actionItems to expect just a node rather than an array of nodes. This change does not create a breaking change as React still handles the array-based nodes, but [this corresponding PR](https://github.com/Kajabi/kajabi-products/pull/24747) patches related aspects in `kp` to avoid any console warnings or errors with mismatching proptypes.

## Related

[SAGE-752](https://kajabi.atlassian.net/browse/SAGE-752)

### DIFF
--- a/packages/sage-react/lib/PageHeading/PageHeading.jsx
+++ b/packages/sage-react/lib/PageHeading/PageHeading.jsx
@@ -4,14 +4,14 @@ import classnames from 'classnames';
 import { Breadcrumbs } from '../Breadcrumbs';
 
 export const PageHeading = ({
-  actionItems,
+  actions,
   breadcrumbs,
   className,
   children,
   help,
   image,
   introText,
-  toolbarItems,
+  toolbar,
   secondaryText,
   ...rest
 }) => (
@@ -51,15 +51,15 @@ export const PageHeading = ({
         <img alt={image.alt || ''} src={image.src} />
       </div>
     )}
-    {toolbarItems && (
+    {toolbar && (
       <div className="sage-page-heading__toolbar">
-        {toolbarItems}
+        {toolbar}
       </div>
     )}
-    {actionItems && (
+    {actions && (
       <div className="sage-page-heading__actions">
         <div className="sage-page-heading__actions-inner">
-          {actionItems}
+          {actions}
         </div>
       </div>
     )}
@@ -72,18 +72,18 @@ export const PageHeading = ({
 );
 
 PageHeading.defaultProps = {
-  actionItems: null,
+  actions: null,
   breadcrumbs: null,
   className: '',
   help: null,
   image: {},
   introText: null,
-  toolbarItems: null,
+  toolbar: null,
   secondaryText: null,
 };
 
 PageHeading.propTypes = {
-  actionItems: PropTypes.node,
+  actions: PropTypes.node,
   breadcrumbs: PropTypes.arrayOf(Breadcrumbs.itemPropTypes),
   children: PropTypes.node.isRequired,
   className: PropTypes.string,
@@ -94,5 +94,5 @@ PageHeading.propTypes = {
   }),
   introText: PropTypes.string,
   secondaryText: PropTypes.string,
-  toolbarItems: PropTypes.node,
+  toolbar: PropTypes.node,
 };
--- a/packages/sage-react/lib/PageHeading/PageHeading.jsx
+++ b/packages/sage-react/lib/PageHeading/PageHeading.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import uuid from 'react-uuid';
 import classnames from 'classnames';
 import { Breadcrumbs } from '../Breadcrumbs';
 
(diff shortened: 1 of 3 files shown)


packages/sage-react/lib/PageHeading/PageHeading.jsx
```diff
@@ -84,7 +83,7 @@ PageHeading.defaultProps = {
 };
 
 PageHeading.propTypes = {
-  actionItems: PropTypes.arrayOf(PropTypes.node),
+  actionItems: PropTypes.node,
```
ju-Skinner, line 86: Could you provide more context around the reasoning for the change here? Both prop names `actionItems` and `toolbarItems` are named in a way (plural naming) that indicate they are to contain multiple items. This seems to indicate better use for an array over node usage.

Also, looking at the ticket you referenced  [SAGE-752](https://kajabi.atlassian.net/browse/SAGE-752), there is no context as to the "why" for this change. To help me, and possibly others, better review this, could you add more context and details to the change? This can be done either here via PR review or an ad-hoc zoom meeting.

packages/sage-react/lib/PageHeading/PageHeading.story.jsx
```diff
@@ -69,29 +71,31 @@ PageHeadingWithThumbnail.args = {
...
+      </Button>
+      <Button
+        color={Button.COLORS.SECONDARY}
+        icon={SageTokens.ICONS.GEAR}
+        subtle={true}
+      >
+        Settings
+      </Button>
+    </>
+  ),
```
ju-Skinner, line 98: Same comment applies here as well, but in regards to `toolbarItems`.

packages/sage-react/lib/PageHeading/PageHeading.story.jsx
```diff
@@ -13,23 +13,25 @@ export default {
...
+          color={Button.COLORS.PRIMARY}
+          icon={SageTokens.ICONS.CART}
+        >
+          Create
+        </Button>
+      </>
+    ),
     toolbarItems: null,
-  }
+  },
```
ju-Skinner, line 34: To piggyback off of my above comments, you can see here in the story, that we are still listing multiple "action items"; in this case `buttons`.

ju-Skinner (changes requested):
See comments.

packages/sage-react/lib/PageHeading/PageHeading.jsx
philschanely: Great points, @ju-Skinner ! 🙌 and thanks for the quick zoom sync too. I have updated the PR description to call out more clearly:

- we have an opportunity to align better with how these similar aspects are named in Rails, and
- we have an opportunity to align better with convention we have in React for how slots like this are typed and rendered.

Also given these changes I'm updating the classification of this change to be a breaking change. The PR on the `kp` side is also updated to match.

packages/sage-react/lib/PageHeading/PageHeading.jsx
```diff
@@ -1,18 +1,17 @@
...
-  actionItems,
+  actions,
   breadcrumbs,
   className,
   children,
   help,
   image,
   introText,
-  toolbarItems,
+  toolbar,
```
ju-Skinner, line 14: Seeing this property makes me believe that we are to pass a Toolbar component. I know you we somewhat opposed to using `toolbarContent` but I believe the more clear we can be with the use or placement the easier it is for devs.

ju-Skinner (changes requested):
See most recent comment regarding `toolbar` param

packages/sage-react/lib/PageHeading/PageHeading.jsx
philschanely: @ju-Skinner Thanks! I think there are multiple correct/good ways to look at it, so I'd like to see what some other colleagues think. Do you feel the same way about `actions` which is also a content slot here?

Def open to going with a team consensus so looking forward to see what others think 🙌

pixelflips: I can understand both points here, but it almost seems like to-may-to vs to-mah-to. IMHO, it seems like we need to decide if our priority at the moment is semantics or consistency.

While both are viable options for sure, I'd lean toward consistency right now, and if we want to think more about semantics (eg. prop naming) that could be addressed in a larger follow-up. I feel like we'd want to keep those aligned in both areas even if we do decide to change the naming. 

Overall not sure if that helps or not, but just thoughts since it was requested.

pixelflips (approved):

packages/sage-react/lib/PageHeading/PageHeading.jsx
philschanely: @ju-Skinner it sounds like we're open overall here but consensus leans to `toolbar` for now. Is this important enough to you that we should change to `toolbarContent` now?

ju-Skinner: @philschanely Overall, I still think the naming of these should be changed, but for the sake of consistency between rails and react, we can go with `toolbar` and `actions`. 

I agree with both @monkeypox8 and @pixelflips with having a larger initiative for these to follow a standard,  which then they would be documented.

ju-Skinner (approved):
Approved, but see comment.

PR #1566, opened 2022-08-22
[SAGE-780] `SageGrid` column class fix

monkeypox8:
## Description

[SAGE-780](https://kajabi.atlassian.net/browse/SAGE-780)

Addresses a bug in the 12-column grid when using Rails `SageGridCol` component by removing the duplicate `sage-col` class when a responsive breakpoint is provided. In certain cases, this caused irregular sizing issues due to conflicting media queries.

Most easily noticeable when the `lg` breakpoint is in use. For example, open the "Help & Feedback" modal in the KP sidebar, and note that the `sage-col--lg-6` columns follow the `sage-col-6` media query when resized.

## Screenshots

|  Before  |  After  |
|--------|--------|
|before-grid-col|after-grid-col|

## Testing in `sage-lib`
1. Navigate to the [Sage docs homepage](http://localhost:4000/pages/index)
2. Inspect any of the content columns (within `.docs-section`)
3. Confirm the grid columns are only rendering responsive grid column classes (ex. `sage-col--md-4`)

1. (**HIGH**) Modifies output of `SageGridCol` component
   - [ ] Website -> Blog posts -> Edit post 
   - [ ] Sidebar -> "Help & Feedback" modal

### DIFF
--- a/docs/lib/sage_rails/app/helpers/sage_grid_helper.rb
+++ b/docs/lib/sage_rails/app/helpers/sage_grid_helper.rb
@@ -1,0 +1,29 @@
+module SageGridHelper
+  def grid_column_classes(col)
+    column_sizes = ["xlarge", "large", "medium", "small"]
+    column_classes = ""
+
+    if col.breakpoint.present? && col.size.present?
+      column_classes << "sage-col--#{col.breakpoint}-#{col.size}"
+    elsif !col.breakpoint.present? && col.size.present?
+      column_classes << "sage-col-#{col.size}"
+    # elsif !col.breakpoint.present? && !col.size.present? && grid_col_legacy(column_sizes)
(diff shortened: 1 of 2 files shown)


QuintonJason (approved):
Looks and works great!

philschanely (approved):
Love it! Nice work. Not a blocker, but I wonder if this could/should be a method on the Grid component itself rather than in a helper?

pixelflips (approved):

monkeypox8:
> Love it! Nice work. Not a blocker, but I wonder if this could/should be a method on the Grid component itself rather than in a helper?

That's a great point, separating out into a helper is probably unnecessary. Since this was intended as a somewhat temporary patch until the 12-col grid situation is fleshed out, I'm inclined to wait to fold it in until we have a chance to refactor the component.

PR #1573, opened 2022-08-29
[SAGE-757] Table - responsive updates

QuintonJason:
## Description
Repurposes table markup for a "stacked" visual layout on small screens.

- Table structure semantics are affected when setting `display`. Each default `role` has been added back in with JS.
- We're using JS to duplicate table header (`th`) text into each individual cell `td` as `span` elements. The actual table headers are visually hidden but remain readable for screen readers, keeping existing semantics/relationship with cells intact.

### Requirements:
- `responsive_stack` property must be enabled
- Existing `is_responsive` property can still be set independently, and will take effect on larger screen sizes **OR** when mobile stack content overflows its container
- JS runs on **page load/init only** to minimize reflow. The expectation being that users will not switch between multiple viewports, so adding a listener for resize should not be needed

### Current assumptions:
- headers are _required_ for tables
- cells will have a **max** of 4 children
  - example: multiple tags/badges within a cell
- truncated text is disabled on mobile
  - `max-width` set on truncated text causes issue with cell width layout
- some specific content will always be hidden when "stacked"
  - avatars: use varying nested containers

#### TODOs
- [ ] Design: confirm moving row selection checkbox (in table header) to bulk actions
- [x] Design: confirm patterns:
  - [x] does this apply to *all* tables, or only special cases?
  - [x] if all tables, strict rules will be needed to enforce markup order/structure
  - [x] verify [overflow approach](https://github.com/Kajabi/sage-lib/pull/1573/files#r1040334638)
- [ ] account for React `headers` schema override 
- [ ] investigate simplified use of dropdowns with `sage_table_for` - requires nested string(s)/schema
- [ ] investigate "action" column container for checkbox, avatar, options content

## Screenshots

|  |Before  |  After  |
|-----|---|--------|
|Standard table|rails|after-table-stack|
|Using `sage_table_for`|before-table-for|after-table-for-stack|

## Testing in `sage-lib`

Visit the table view and verify that the mobile view aligns with the design spec
- [Rails](http://localhost:4000/pages/component/table?tab=preview)
- [React](http://localhost:4100/?path=/docs/sage-table--default)

⚠️ Note: when testing the responsive view using browser devtools, you must refresh the page after resizing (see description above, under Requirements)

1. (**HIGH**) Update mobile styles for Table.
   - [ ] People Page

## Related

Closes [SAGE-757](https://kajabi.atlassian.net/browse/SAGE-757)

### DIFF
--- a/docs/lib/sage_rails/app/helpers/sage_table_helper.rb
+++ b/docs/lib/sage_rails/app/helpers/sage_table_helper.rb
@@ -215,6 +215,56 @@ module SageTableHelper
     table.contents
   end
 
+  def sage_table_classes(table)
+    table_classlist = "sage-table"
+
+    table_classlist << " sage-table--selectable" if table.selectable
+    table_classlist << " sage-table--condensed" if table.condensed
+    table_classlist << " sage-table--has-leading-input" if table.has_leading_input
+    table_classlist << " sage-table--has-menu-options" if table.has_menu_options
+
+    return table_classlist
+  end
+
+  def sage_table_wrapper_classes(table, is_responsive)
+    table_wrapper_classlist = "sage-table-wrapper"
+    table_wrapper_classlist << " sage-table-wrapper--reset-above" if table.reset_above
+    table_wrapper_classlist << " sage-table-wrapper--reset-below" if table.reset_below
+    table_wrapper_classlist << " sage-table-wrapper--scroll" if is_responsive
+
+    return table_wrapper_classlist
+  end
+
+  def sage_table_caption_classes(table)
+    table_caption_classlist = "sage-table__caption"
+
+    if table.caption_side
+      table_caption_classlist << " sage-table__caption--#{table.caption_side}"
+    end
+
+    return table_caption_classlist
+  end
+
+  def sage_table_row_classes(table)
+    table_row_classlist = ""
+
+    if table.selectable
+      table_row_classlist << "sage-table__row--selectable"
+    end
+
+    return table_row_classlist
+  end
+
+  def sage_table_cell_classes(table)
+    table_cell_classlist = ""
+
+    if table.has_borders
+      table_cell_classlist << "sage-table-cell--borders"
+    end
+
+    return table_cell_classlist
+  end
+
   def sage_table_sortable_header_link(label, attribute)
     current = attribute.to_s == sort_column
     asc = sort_direction == "asc"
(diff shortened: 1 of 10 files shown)


docs/lib/sage_rails/app/helpers/sage_table_helper.rb
```diff
@@ -215,6 +217,55 @@ def sage_table_for(collection, opts={})
...
   end
 
+  def sage_table_classes(table)
+    table_classlist = "sage-table"
+
+    table_classlist << " sage-table--selectable" if table.selectable
+    table_classlist << " sage-table--condensed" if table.condensed
+    table_classlist << " sage-table--has-leading-input" if table.has_leading_input
+    table_classlist << " sage-table--has-menu-options" if table.has_menu_options
+    table_classlist << " sage-table--stack" if table.responsive_stack
```
monkeypox8, line 227: Will need to clarify with design here:

I've created a new prop `responsive_stack` rather than repurposing `is_responsive`. This is all under the assumption that we want to maintain maximum backwards compatibility with overflow/scrolling tables since the responsive stack layout is much more stringent. Retrofitting all current table use is not likely to be a trivial effort.

packages/sage-system/lib/table.js
```diff
@@ -24,6 +30,88 @@ Sage.table = (function() {
...
+      const rows = table.querySelectorAll('tbody tr');
+      const tableHeadings = [];
+
+      // populate an array with each table's headers
+      headers.forEach(header => {
+        const label = header.textContent.trim();
+        tableHeadings.push(label);
+      })
+
+      rows.forEach(row => {
```
monkeypox8, line 60: Using multiple nested loops to dig into the table markup. Probably not ideal, but I'm not sure we have much of an alternative. Performance impact should be minimal with the majority of cases rendering paginated results.

docs/lib/sage_rails/app/views/sage_components/_sage_table.html.erb
```diff
@@ -1,32 +1,24 @@
...
     <div class="sage-table-wrapper__overflow">
   <% end %>
 
-  <table class="
-      sage-table
-      <%= "sage-table--selectable" if component.selectable %>
-      <%= "sage-table--condensed" if component.condensed %>
-    "
+  <table
+    class="<%= sage_table_classes(component) %>"
```
monkeypox8, line 16: Moved CSS class output logic to `sage_table_helper`

PR #1577, opened 2022-09-06
[SAGE-144]: Add NilClass to all optional props (Rails)

pixelflips:
## Description
When iterating through creating components within a loop in Rails optional properties benefit from including nil as a possible value. This way a default of nil can be provided while also allowing additional values to be set conditionally within the loop.

This PR adds `NilClass` to all optional props for the rails components.

## Screenshots
No visual changes.

## Testing in `sage-lib`
Check files changed and then try passing `nil` to any of the updated properties in a preview file

1. (**LOW**) Adds `NilClass` to all optional props. No effect is expected on KP.

## Related
https://kajabi.atlassian.net/browse/SAGE-144


philschanely (approved):

ju-Skinner (approved):

docs/lib/sage_rails/app/sage_components/sage_card_list.rb
```diff
@@ -1,6 +1,6 @@
 class SageCardList < SageComponent
   set_attribute_schema({
-    block_spacing: [:optional, Set.new(["sm"])],
-    hide_first_border: [:optional, TrueClass, String],
+    block_spacing: [:optional, NilClass, Set.new(["sm"])],
+    hide_first_border: [:optional, NilClass, TrueClass],
```
goodwinchris, line 4: Lost `String` here - is that intended?

pixelflips: @goodwinchris great catch and not intended. will bring back. thanks!

pixelflips: @goodwinchris fixed. thanks again!

goodwinchris (approved):

PR #1582, opened 2022-09-09
Deprecate `NullView` component and `EmptyState`'s `compact` property

philschanely:
## Description

See https://kajabi.atlassian.net/browse/SAGE-63. This PR removes all the artifacts for `NullView` (only present in Rails) and removes the deprecated `compact<boolean>` property from `EmptyState` (both React and Rails).

## Testing in `sage-lib`

See EmptyState that now fully replaces NullView and no longer contains the `compact` property.

1. (**BREAKING**) Fully remove NullView (replaced by EmptyState) and EmptyState's `compact` property (now use `scope`). NOTE: All uses of these features are already removed from `kajabi-products` so this is not a blocking breaking change.

## Related

[SAGE-63](https://kajabi.atlassian.net/browse/SAGE-63)


pixelflips (approved):
Gotta love some good 'ol cleanup. Great work! 👍🏼

monkeypox8 (approved):
👍 LGTM, nice work!

PR #1584, opened 2022-09-12
Modal: Background controls and other improvements

philschanely:
## Description

This PR improves the Modal by adjusting how the current `disable_background_dismiss` affordance works:

- [x] Add a new affordance specifically for disabling the background screen overall: with this activated one can still click outside the Modal to dismiss it but there is no longer a visible screen.
- [x] Revise what is affected by the existing `disable_background_dismiss` affordance so that this does exactly (and only) what the name indicates: with this activated then clicking the space outside the modal will no longer dismiss it. This can now be used on its own or in conjunction with `disable_background` to make the area outside the modal able to be interacted with.
- [x] Keep the `disable_background_blur` intact but update the classes used for consistency.
- [x] Update how these properties are used within Drawer to ensure the same default behavior but also the ability to adjust these settings.
- [x] Improve documentation of Modal to better explain its features and display the variants.

## Screenshots

<a href="https://www.loom.com/share/9020edb66cf546be965b3160d5e3ec83">
    <p>Modal background options demo - Watch Video</p>
    
  </a>

## Testing in `sage-lib`

See http://localhost:4000/pages/component/modal
See http://localhost:4100/?path=/story/sage-modal--wired

1. (**BREAKING**) Update affordances on Modal for controlling background and dismiss functionality.
   - Existing property `disable_background_dismiss` now only disables the background dismiss functionality and no longer also makes the background screen disappear.
   - New property `disabled_background` can now be used specifically to hide the background screen itself.
   - NOTE: No uses of these properties currently in `kp` are adversely affected by this breaking change. However, [this PR](https://github.com/Kajabi/kajabi-products/pull/25046) removes some styling that is no longer needed as a result of this change. It does not need to block be this version bump and can be merged afterwards.

## Related

https://kajabi.atlassian.net/browse/SAGE-746


QuintonJason (changes requested):
I didn’t understand some of the changes in the PR based on Darrell’s initial ask and needed some help with clarity. Initially there seemed to be confusion around what `disableBackgroundDismiss` did in the React component as he was looking to control when the modal would close.

We currently have some props, `onExit` and `active`  coupled with `useEffect` and `useState` that handles the state of the modal being open regardless of what’s happening inside the modal. 

I was also curious about the changes to the Drawer component as it extends from the Modal. There are changes related to click behavior of the modal/drawer that weren’t highlighted in the original ask. 

Can you me understand the need for the changes based on the ask?

packages/sage-react/lib/Drawer/Drawer.jsx
```diff
@@ -16,6 +16,8 @@ export const Drawer = ({
   children,
   className,
   customHeader,
+  disableBackground,
```
philschanely, line 19: This changes sync up Drawer with the latest affordances available from Modal.

packages/sage-react/lib/Drawer/Drawer.jsx
```diff
@@ -36,7 +38,7 @@ export const Drawer = ({
         onExpandChange(expanded ? DRAWER_END_EXPAND : DRAWER_END_COLLAPSE);
       }, 750);
     }
-  }, [expanded]);
+  }, [expanded, onExpandChange]);
```
philschanely, line 41: This change fixed a linting issue related to the dependencies here.

packages/sage-react/lib/Drawer/Drawer.story.jsx
```diff
@@ -65,27 +65,7 @@ export const WiredExample = () => {
...
-        Lorem ipsum dolor sit amet consectetur adipisicing elit.
-        Eum in deleniti doloribus sequi ipsam unde pariatur perspiciatis
-        ex molestias temporibus accusantium fuga debitis,
-        voluptates hic eos ab perferendis ad quo.
-      </p>
-    </>
-  );
-
-  const respondToDrawerExpandChange = (status) => {
+  const respondToDrawerExpandChange = useCallback((status) => {
```
philschanely, line 68: `useCallback` here was necessary when the dependencies were updated on `packages/sage-react/lib/Drawer/Drawer.jsx:41`. This helps prevent too many re-renders.

packages/sage-react/lib/Drawer/Drawer.story.jsx
```diff
@@ -120,9 +114,14 @@ export const WiredExample = () => {
       <Drawer
         active={drawerActive}
         expanded={drawerExpanded}
-        onExit={() => setDrawerActive(false)}
+        onExit={() => {
```
philschanely, line 117: This just cleaned up how the Drawer demo was behaving to ensure it always re-opens in the collapsed form.

packages/sage-react/lib/Drawer/Drawer.story.jsx
```diff
@@ -120,9 +114,14 @@ export const WiredExample = () => {
...
+          setDrawerExpanded(false);
+          setDrawerActive(false);
+        }}
+        id="1234jkhkqd"
         onExpandChange={respondToDrawerExpandChange}
         customHeader={drawerCustomHeader}
+        disableBackground={!drawerExpanded}
```
philschanely, line 124: This demonstrates how the background can be toggled on and off in sync with the expanded/collapsed drawer if desired.

docs/lib/sage_rails/app/sage_components/sage_drawer.rb
```diff
@@ -1,6 +1,8 @@
 class SageDrawer < SageComponent
   set_attribute_schema({
     active: [:optional, NilClass, TrueClass],
+    disable_background: [:optional, NilClass, TrueClass],
```
philschanely, line 4: These props just ensure that Drawer can control the nested Modal when desired; sync with latest from Modal.

philschanely:
Hi @QuintonJason ! Sorry for the confusion!

Darrell's context needed the Modal's feature of dismissing the modal when clicking outside of it to be disabled, yet the current setup for that `disable_background_dismiss` also would remove the background screen (which Darrell did not want). I believe we'd originally added this affordance when we were building Drawer so that we could allow users to interact outside the "modal" when the Drawer/modal is open.

So the goal was to separate those two affordances so that they could be toggled independently. As a result, I have added `disable_background` specifically for hiding the screen when desired, and I modified `disable_background_dismiss` to only disable the dismiss action when desired.

These two features can also be combined to allow interactions outside of the "modal" which is what was needed as the default setup for Drawer. So I carried these new and revised properties over to Drawer to ensure it continues to behave as it did before while also allowing more control when needed. The story for Drawer was updated accordingly.

I have added comments on changes aspects to further point out what and why. 

Did this help clarify? Would you like to sync to explore more? Looking forward to your feedback.

QuintonJason (changes requested):
The creation of `disable_background_dismiss` was added to the Drawer as it was merely supposed to extend the Modal, but it now seems to be influencing Modal decisions. This is a breaking change that is influencing existing modal click behavior, so we'd need to consult about this as a team to collectively provide the best experience for all consumers of sage. 

The naming of the various `no-background` classes are very similar and cause confusion as they do different things, but have the same prefix.

Going back to the initial ask, it seems the main confusion resides around the use of the class name and expected behavior. My concern is that instead of clarifying the prop name and/or adding an event to hook into, there is an addition of props that could lead to further confusion.

philschanely:
Thanks @QuintonJason . I'll look to schedule a team discussion around this. 🙌

philschanely:
Thanks to @QuintonJason 's feedback we've relayed the option to Darrell to address the initial need by using `active` and `onExit`. We decided to close this PR and its related Jira ticket that will look into reducing the dependency between Drawer and Modal if not separate them altogether.

PR #1598, opened 2022-09-23
[DSS-147] Update Property component

monkeypox8:
## Description
Aligns the Property component to current [Figma specs](https://www.figma.com/file/sMbtLUHSt2vfKgKjyQ3052/%F0%9F%A7%A9-Sage-components?node-id=4583%3A21358):
- increase property text style from `400` to `500`
- property color change from `charcoal-200` to `charcoal-300`

Additional requirements:
- properties are read-only and are not interactive
  - verified in KP that current `Property` use is non-interactive
- max of 6 properties in a property group
  - consulted with Cam; we'll treat this as a recommendation for Design rather than a limitation in code

## Screenshots

|  Before  |  After  |
|--------|--------|
|property-before|property-after|

## Testing in `sage-lib`
1. Navigate to the `Property` component in the [Rails docs](http://localhost:4000/pages/component/property) and in Storybook
2. Inspect the element on the page using your browser devtools, and confirm the specs match Figma (`500` font-weight, `charcoal-300` color)

1. (**MED**) Property component style updates. CSS only; no impact expected in KP.
   - [ ] New Product Modal: product category features list (see screenshot in PR)

### DIFF
--- a/docs/app/helpers/components_helper.rb
+++ b/docs/app/helpers/components_helper.rb
@@ -663,7 +663,7 @@ module ComponentsHelper
       },
       {
         title: "property",
-        description: "TO BE NAMED – Displays a metadata item for a resource.",
+        description: "Displays a metadata item for a resource.",
         scss: "done",
         docs: "todo",
         rails: "done",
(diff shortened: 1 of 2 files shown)


ju-Skinner (approved):

pixelflips (approved):
No concerns here, nice work! :shipit:

PR #1605, opened 2022-09-29
[DSS-157] Avatar - Custom Icon and Color

anechol:
## Description

Adds props to allow for background and foreground (icon stroke/fill) color for badges. Attempting to provide a backwards compatible solution to not disturb what's already been used in `kp`.

## Screenshots

Screen Shot 2022-10-06 at 10 18 11 AM

https://user-images.githubusercontent.com/14791307/195883012-d0d0ecb1-656f-400e-9679-b6940c037789.mov

### Rails
- View Avatar in [Rails](http://localhost:4000/pages/component/avatar?tab=preview) and [React](http://localhost:4100/?path=/docs/sage-avatar--custom-badge)
- Check that new props work as expected

(**LOW**) Adds props to allow for background and foreground (icon stroke/fill) color for badges.

## Related

[DSS-157](https://kajabi.atlassian.net/browse/DSS-157)

### DIFF
--- a/packages/sage-react/lib/Avatar/Avatar.jsx
+++ b/packages/sage-react/lib/Avatar/Avatar.jsx
@@ -70,8 +70,8 @@ export const Avatar = ({
       {badge && (
         <div
           className={badgeClassnames}
-          style={(badgeBackgroundColor && badgeBackgroundColor !== '') && ({
-            '--badge-custom-bg-color': badgeBackgroundColor
+          style={({
+            '--badge-custom-bg-color': badgeBackgroundColor || ''
           })}
         >
           <Icon
--- a/packages/sage-react/lib/Avatar/Avatar.jsx
+++ b/packages/sage-react/lib/Avatar/Avatar.jsx
@@ -7,7 +7,8 @@ import { AVATAR_COLORS } from './configs';
 
 export const Avatar = ({
   badge,
-  badgeColor,
+  badgeBackgroundColor,
+  badgeForegroundColor,
   badgeIcon,
   className,
   centered,
@@ -22,8 +23,16 @@ export const Avatar = ({
     className,
     {
       'sage-avatar--centered': centered,
-      'sage-avatar--custom-badge': badgeColor || badgeIcon,
+      'sage-avatar--custom-badge': badgeBackgroundColor || badgeIcon || badgeForegroundColor,
       [`sage-avatar--${color}`]: color,
+    }
+  );
+
+  const badgeClassnames = classnames(
+    'sage-avatar__badge',
+    className,
+    {
+      'sage-avatar__badge--custom-bg': badgeBackgroundColor
     }
   );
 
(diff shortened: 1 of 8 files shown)


ju-Skinner (approved):

philschanely (approved):
Looks great!

Not blocking, but what do you think about restructuring so that rather than `badge`, `badgeColor`, and `badgeIcon` we just have `badge` but it can accept just `true` (for default badge) or a hash/object with optional `icon` and `color` properties for encapsulating the custom settings?

pixelflips (commented):
I can't quite work out if it's a Storybook issue, but when I set badge to true and adjust the badge color and icon, it changes the icon color and not the background color.

Happy to pair if it's not obvious.

Screen Shot 2022-10-06 at 8 34 37 AM

monkeypox8 (approved):
👍 LGTM, with one note for the badge appearance:

Requirements for the issue are unclear on this, but the Figma mocks are displaying various social media logo badges with varying background color fills.

The issue that @pixelflips is seeing is because the icon examples in Rails are using `-filled` variants. If the intent will be to allow the `badge_color` to affect the background, then we'll have to ask Design to add those versions of the icons as well. Otherwise, it's working as intended.

anechol:
@pixelflips @monkeypox8 Thanks for the feedback here. I just spoke to Quinton about this issue and yes, it looks like this is going to have to be brought up to design. Filled icons are working as intended, but those with only strokes are going to look awkward. Going to bring this up in the team sync for ideas.

anechol:
Per feedback from the team and design, I'll be implementing a solution for both a foreground (icon stroke/fill) and background color.

pixelflips (changes requested):
Code looks great, but I am seeing an issue with Storybook. When setting Badge to `true` and then trying to add a `string` for the `badgeBackgroundColor` the screen goes blank and an error shows in the console.


https://user-images.githubusercontent.com/1175111/195393080-a29cb9ad-f369-45d6-b648-eaf9069b505e.mov

anechol:
> Code looks great, but I am seeing an issue with Storybook. When setting Badge to `true` and then trying to add a `string` for the `badgeBackgroundColor` the screen goes blank and an error shows in the console.
> 
>  Screen.Recording.2022-10-12.at.9.02.01.AM.mov

Yeah @pixelflips this is a weird issue, and I think it's Storybook related, but can't confirm entirely. If you set the background color first and then set badge to true, then it works as inspected.

pixelflips:
> Yeah @pixelflips this is a weird issue, and I think it's Storybook related, but can't confirm entirely. If you set the background color first and then set badge to true, then it works as inspected.

@anechol it is weird for sure, maybe @ju-Skinner or @philschanely have thoughts on the error I mentioned above. 🤞🏼

anechol:
@monkeypox8 Thanks for the pointer. I adjusted the conditional so this should work fine in storybook now.

anechol:
> While continuing to review and changing up the controls, also noticed that the icons seem to break out in some cases. Will drop a screenshot so you can see the controls.
> 
> Screen Shot 2022-10-12 at 10 16 19 AM

Ahh thanks for pointing this out @pixelflips . I might have to look into this further as I'm not sure what's a good solution to handle every use case for every icon in the library 🤔

packages/sage-react/lib/Avatar/Avatar.jsx
```diff
@@ -55,10 +68,15 @@ export const Avatar = ({
   return (
     <div className={classNames} style={style} {...rest}>
       {badge && (
-        <div className="sage-avatar__badge">
+        <div
+          className={badgeClassnames}
+          style={({
```
monkeypox8, line 73: (non-blocking) You can also drop the enclosing parentheses `()` here now that the conditional has been removed.

monkeypox8 (approved):
👍 LGTM! Left a note on one cleanup item to consider.

monkeypox8:
> I might have to look into this further as I'm not sure what's a good solution to handle every use case for every icon in the library 🤔

Agreed; this is going to be pretty difficult to solve without a change to our icon system sizing, so I'm inclined to leave it for another day if the group's on board. Right now the use case for custom badges is limited. But if it does become an issue, we could always apply a 90% scale transform or something similar, though it'd have to be across the board.

ju-Skinner (approved):
great job. Thanks for adding specs.

cameronsimony (approved):
👍 👍

PR #1612, opened 2022-10-03
[DSS-183]: Storybook Components Not Displaying Code Examples

pixelflips:
## Description
Code examples for the following components do not display properly in Storybook.
Most are returning `<No Display Name />` instead of markup.

- FormSection
- IconCard
- Loader
- Popover
- Select
- Table
- Tooltip
- Topbar
- Type
- UploadCard

Updates included using the `{Story()}` function when using decorators in the component story.

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2022-10-03 at 4 01 03 PM|Screen Shot 2022-10-03 at 4 00 33 PM|

## Testing in `sage-lib`
Navigate to [Storybook](http://localhost:4100/)
Visit each component listed above.
Verify code examples now display on Docs tab.

1. (**LOW**) Fixes code examples not displaying in Storybook. Documentation update only, no effect on KP expected.

## Related
https://kajabi.atlassian.net/browse/DSS-183

### DIFF
--- a/packages/sage-react/lib/FormSection/FormSection.story.jsx
+++ b/packages/sage-react/lib/FormSection/FormSection.story.jsx
@@ -25,10 +25,8 @@ const Template = (args) => <FormSection {...args} />;
 export const Default = Template.bind({});
 Default.decorators = [
   (Story) => (
-    <>
-      <div style={{ width: '75%', marginLeft: 'auto', marginRight: 'auto' }}>
-        <Story />
-      </div>
-    </>
+    <div style={{ width: '75%', marginLeft: 'auto', marginRight: 'auto' }}>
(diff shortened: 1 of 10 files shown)


anechol (approved):
LGTM! Just one non-blocking change: the Popover story is gigantic. Can we remove the defined height in that story?

pixelflips:
> LGTM! Just one non-blocking change: the Popover story is gigantic. Can we remove the defined height in that story?

@anechol If we completely remove it, the popover gets cut off when it's opened. I'll find a height that still allows it to display properly. Right now it's 100vh, I'll find a better value. Thanks for pointing that out!

PR #1619, opened 2022-10-06
Table: Add sortable headers in React

philschanely:
## Description

(WIP)

This PR adds sortable heading styling for React Tables.

- [x] Add React sub component for presenting an intentionally styled sortable heading
- [x] Add an example story that shows this sub component in use
- [ ] Seek early feedback on direction

If direction is approved, the following tasks are also in mind:

- [ ] Add documentation for the new sub component
- [x] Update styling to latest design specs
- [ ] Create follow-up tickets for Rails side

## Screenshots

|  Before  |  After  |
|--------|--------|
|n/a|table-styles|

## Testing in `sage-lib`

- See http://localhost:4100/?path=/docs/sage-table--table-with-sortable-headers

1. (**LOW/MEDIUM/HIGH/BREAKING**) TBD

## Related

https://kajabi.atlassian.net/browse/DSS-131


monkeypox8:
Looking into potential items for labeling sort order once applied. Ideally we'd be able to make use of [`aria-sort` on the table headers](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort)

packages/sage-assets/lib/stylesheets/components/_table.scss
```diff
@@ -136,49 +138,40 @@ $-table-avatar-width: rem(32px);
...
+    &:focus {
+      color: $-table-heading-font-hover-color;
     }
 
-    &:hover,
-    &:focus,
+    &:focus-visible,
     &:active {
-      color: $-table-cell-focus;
+      @include sage-focus-outline--update-color(sage-color(primary, 200));
```
monkeypox8, line 161: Had a question from Cam on whether we can/should apply the focus ring only on keyboard  activity. Now that `focus-visible` has greater adoption this should be possible, which wasn't the case with legacy Sage. For the sake of system consistency, I'm leaving it in for now until we can have a larger discussion.

PR #1637, opened 2022-11-17
[DSS-225] Popover - add caret to popover

QuintonJason:
## Description
- [ ] add caret to all popover instances
- [ ] update the horizontal position based on presence of tooltip
- [ ] update react demos

## Screenshots

|  Before  |  After  |
|--------|--------|
|||

## Testing in `sage-lib`

- [Rails](http://localhost:4000/pages/component/popover?tab=preview)
- React

1. (**LOW**) Add caret to `Popover`. Minimal visual change

## Related

Closes [DSS-225](https://kajabi.atlassian.net/browse/DSS-225)


packages/sage-assets/lib/stylesheets/components/_popover.scss
```diff
@@ -49,38 +54,97 @@ $-popover-panel-offset: sage-spacing(sm);
...
     bottom: calc(100% + #{$-popover-panel-offset});
     left: 0;
+
+    &::after {
+      left: $-popover-tooltip-offset;
+      top: 100%;
+      transform: translate3d(-50%, 0, 0);
+      border-left: $-popover-tooltip-inactive;
+      border-right: $-popover-tooltip-inactive;
+      border-top: $-popover-tooltip-active;
```
goodwinchris, line 76: Feels like most of this could move up to line 60 so we don't have to repeat it for each modifier, right?

QuintonJason: Much of this code isn't repeating as each modifier has different positional properties and 3 border sides updated. The only common thing I was able to pull out was https://github.com/Kajabi/sage-lib/blob/8e10bacc1391e7a92d246ba5031cd069b7332ab2/packages/sage-assets/lib/stylesheets/components/_popover.scss#L57 through 60. Is that what you're referring to?

goodwinchris: Oh I see, is it possible to make the shape with the same three border declarations and then rotate it correctly or does that get funky with alignment?

packages/sage-assets/lib/stylesheets/components/_popover.scss
```diff
@@ -49,38 +54,97 @@ $-popover-panel-offset: sage-spacing(sm);
...
+  &::after {
+    content: "";
+    position: absolute;
+  }
+
   .sage-popover--is-expanded & {
     visibility: visible;
   }
 
   .sage-popover--top & {
```
monkeypox8, line 66: I believe this has been brought up before since it's an existing issue, but popovers on mobile have a high potential to break layouts. Though the caret addition helps with orientation, we should also consider standardizing display of all popovers on mobile, regardless of specified position. 

Since we're only affecting the visibility on the popover, and its default position is right-aligned to its toggle/trigger, the popover's content has a high chance of overflowing the parent container with smaller screens. This gets worse with left-positioned popovers, which can't be viewed/scrolled at all due to the overlay lock.

<details>
<summary>popover mobile overflow</summary>

</details>

<details>
<summary>popover truncated</summary>

</details>

Design will want to weigh in here, but treating popover content in the same way as a `modal` (centered on screen, fixed position) might be an option.

packages/sage-assets/lib/stylesheets/components/_popover.scss
monkeypox8: > Oh I see, is it possible to make the shape with the same three border declarations and then rotate it correctly or does that get funky with alignment?

Love this (assuming the alignment/position doesn't get all out of whack with the transform).

monkeypox8 (approved):
👍🏼 Looks and works great on larger screens. On mobile there's an existing issue with position that will require further discussion/follow-up.

packages/sage-assets/lib/stylesheets/components/_popover.scss
QuintonJason: @goodwinchris I created this task for that work. The tooltip could also benefit from such change as well, so created a ticket for a shared function that both tooltip and popover would consume: https://kajabi.atlassian.net/browse/DSS-226

goodwinchris (approved):

packages/sage-assets/lib/stylesheets/components/_popover.scss
QuintonJason: @monkeypox8 create [DSS-227](https://kajabi.atlassian.net/browse/DSS-227) for tracking this work

PR #1640, opened 2022-11-29
Dockerfiles

voodooGQ:
## Description
This PR introduces Dockerfile's for the Storybook, SASS-Docs, and primary Documentation sites. These will be used in creating AWS instances of the sites to replace Heroku

https://kajabi.atlassian.net/browse/DSS-228


ju-Skinner (approved):

voodooGQ:
@ju-Skinner @rwc9u are we sure we want to yank all the engines? I know they've been a bit of a headache in the past, but I'm a bit paranoid that something may break.

monicawheeler:
@voodooGQ - Excited to see this work! I was curious, I see some updates to the README's and wanted to clarify, is there any additional steps devs need to do to work locally? As an example, when this work is merged do we need to build docker images to view local environments? I want to ensure that if this work impacts day-to-day development, it's documented. Thank you!

voodooGQ:
@monicawheeler README info is purely for building/running the dockerimage. At the moment that has no baring on the development environment, would be purely for PE's benefit when setting up the main sites.

voodooGQ:
@ju-Skinner Sure, we can certainly get by without them, but the `engine` field is used specifically to ensure users don't run into a version that is not supported by the system. I'm less worried about it breaking on  us and more worried on it breaking with anything that uses it. If anything we should probably add new versions that we know Sage is compatible with to the list `node: "14.x" | "16.x" | "18.x"` ect.

Anyways, that's just my soapbox rant. We can move forward if ya'all agree, but we might want to discuss it a little more before we pull the trigger.

ju-Skinner:
@voodooGQ If you currently have reservations about the `engines` node, then we should add it back. We or I can look into testing it without it at a later date. No sense on holding  things up over something that may or may not have impact. 

cc: @rwc9u

rwc9u:
I'd say add it back and make sure the list contains the values used in the Dockerfiles

rwc9u:
I think it also allows '> X.X' syntax so maybe just set a lower boundary?

voodooGQ:
@ju-Skinner @rwc9u, updated. Made it so it takes anything above what we had set before. Take a peek and see if it's time to :shipit:

ju-Skinner:
:shipit:

prsimp (approved):

PR #1641, opened 2022-11-30
Update ruby [2.7.7] / rails [6.0.6] version

ju-Skinner:
## Description
In order to stay current with the current versions being used by Kajabi Products, we are updating both the ruby and rails version for the Sage docs and components.

## Notes
1. You will need to install ruby `2.7.7`. You can run the following based on which version manager you are using
  -  `rbenv install 2.7.7` // rbenv
  -  `ruby-install 2.7.7` // Chruby
  -  `rvm install 2.7.7` // rvm

2. Once you have the latest installed, you will need to run `yarn setup`


pixelflips (approved):

rwc9u (approved):
LGTM 🚢

monkeypox8 (approved):
👍🏼 Thanks for this! Should fix the version mismatch console errors between sage and KP when running the bridge

PR #1645, opened 2022-12-01
[DSS-216] - Storybook Dropdown - options menu not responding

ju-Skinner:
## Jira Ticket
https://kajabi.atlassian.net/browse/DSS-216

## Description
The template that was being used for the story failed to add the spreading of args `{...args}`.  Fixed a few small things
* RichTextEditorIconOnly decorator was using OptionMenu decorator, renamed to define its own.
* Moved OptionDropdown out into its own Story file. This solves seeing properties that are not in use with this component. @goodwinchris this should address your other concern regarding, changing `icon`, `label`, etc. 

## Screenshots
https://www.loom.com/share/e733d646a3784c3aaa5d6386af2c719d

### DIFF
--- a/packages/sage-react/lib/Dropdown/Dropdown.story.jsx
+++ b/packages/sage-react/lib/Dropdown/Dropdown.story.jsx
@@ -94,7 +94,7 @@ MenuWithArrow.decorators = [
 ];
 
 export const OptionMenu = (args) => (
-  <OptionsDropdown options={defaultOptionsItems} isPinned={args.isPinned} />
+  <OptionsDropdown options={defaultOptionsItems} isPinned={args.isPinned} {...args} />
 );
 OptionMenu.args = {
   isPinned: false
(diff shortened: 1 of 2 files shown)


packages/sage-react/lib/Dropdown/Dropdown.story.jsx
```diff
@@ -94,7 +94,7 @@ MenuWithArrow.decorators = [
 ];
 
 export const OptionMenu = (args) => (
-  <OptionsDropdown options={defaultOptionsItems} isPinned={args.isPinned} />
+  <OptionsDropdown options={defaultOptionsItems} isPinned={args.isPinned} {...args} />
```
ju-Skinner: the addition of the spreading of args `{...args}` was the solution.

goodwinchris:
I'm not seeing an option to pass `disabled` into the new options dropdown story. Is that expected?

ju-Skinner:
> I'm not seeing an option to pass `disabled` into the new options dropdown story. Is that expected?
@goodwinchris Based on the Props that were expected in the component, I don't think that was ever an option. 
[OptionsDropdown](https://github.com/Kajabi/sage-lib/blob/develop/packages/sage-react/lib/Dropdown/OptionsDropdown.jsx#L8-L17)

goodwinchris (approved):

PR #1667, opened 2023-01-06
fix(grid): add vertical alignment to individual sage-col elements

goodwinchris:
## Screenshots

|  Before  |  After  |
|--------|--------|
|||

### DIFF
--- a/docs/app/views/pages/grid.html.erb
+++ b/docs/app/views/pages/grid.html.erb
@@ -349,10 +349,10 @@
 </div>
 
 <div class="sage-type">
-  <h2 id="grid-alignment">Alignment</h2>
+  <h2 id="grid-alignment">Row Alignment</h2>
   <h3 class="t-sage-heading-5" id="grid-alignment-horizontal">Horizontal Alignment</h3>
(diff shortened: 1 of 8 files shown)


docs/app/views/pages/grid.html.erb
```diff
@@ -429,32 +427,78 @@
...
+      vertical_alignment: "baseline"
     } do %>
-      <div class="grid-item">Tall column with multiple lines</div>
+      <div class="grid-item">Baseline</div>
     <% end %>
   <% end %>
 
-  <pre class="prettyprint"><code>&lt;%= sage_component SageGridRow, {
-  vertical_alignment: &quot;center&quot;
+    <pre class="prettyprint"><code>&lt;%= sage_component SageGridRow, {
```
ju-Skinner, line 496: Do we need to pass the empty object `{}`?

ju-Skinner (approved):
LGTM...Comment left is not a blocker.

docs/app/views/pages/grid.html.erb
goodwinchris: From what I'm seeing in other examples with nothing in the object, we still need to pass one but I've cleaned up this code a bit so its all on one line now

PR #1671, opened 2023-01-10
fix(dropdown): calculate the position of the dropdown panel

ju-Skinner:
## JIRA
https://kajabi.atlassian.net/browse/DSS-60

## Description
The initial problem was when a dropdown would be invoked and render outside the viewport. This PR addresses that issue, by determining the available viewport space and renders accordingly.

## Screenshots

|  Before  |  After  |
|--------|--------|
image|image|

Loom: [here](https://www.loom.com/share/55ff74d24f134ac59d51f334cb53d146)

1. Turn the bridge on for Sage Lib to KP
2. Navigate to a few locations that use dropdown (a few examples below)
  - Products Index
  - Websites
3. Confirm behavior shown in loom video.

### DIFF
--- a/packages/sage-system/lib/dropdown.js
+++ b/packages/sage-system/lib/dropdown.js
@@ -167,8 +167,63 @@ Sage.dropdown = (function () {
     }
   }
 
+  function getHeight(el) {
+    const styles = window.getComputedStyle(el);
+    const height = el.offsetHeight;
+    const borderTopWidth = parseFloat(styles.borderTopWidth);
+    const borderBottomWidth = parseFloat(styles.borderBottomWidth);
+    const paddingTop = parseFloat(styles.paddingTop);
+    const paddingBottom = parseFloat(styles.paddingBottom);
+
+    return height - borderBottomWidth - borderTopWidth - paddingTop - paddingBottom;
+  }
+
+  function positionElement(el) {
+    let direction = null;
+
+    // Elements
+    const panel = el.lastElementChild
+    const win = panel.ownerDocument.defaultView;
+    const button = document.activeElement;
+    const docEl = window.document.documentElement;
+
+    panel.style.top = ''; // resets the style
+
+    // Dimensions
+    const buttonDimensions = button.getBoundingClientRect();
+    const panelDimensions = panel.getBoundingClientRect();
+
+    const panelNewLoc = {
+      top: buttonDimensions.height + panelDimensions.height
+    }
+
+    const viewport = {
+      top: docEl.scrollTop,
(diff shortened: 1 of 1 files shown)


packages/sage-system/lib/dropdown.js
```diff
@@ -167,8 +167,63 @@ Sage.dropdown = (function () {
...
+    }
+
+    const offset = {
+      top: panelDimensions.top + win.pageYOffset,
+      left: panelDimensions.left + win.pageXOffset,
+      bottom: (panelDimensions.top + win.pageYOffset)
+    };
+
+    const enoughSpaceAbove = viewport.top < ( offset.top + getHeight(panel));
+    const enoughSpaceBelow = viewport.bottom > (offset.bottom + getHeight(panel));
```
QuintonJason, line 212: Regarding the positioning, should `left` and `right` also be considered, or is this something that is out of scope for this work?

ju-Skinner: My opinion is out of scope...But I am open to hear others as well. 

cc: @monicawheeler

pixelflips (approved):
Unable to check with the bridge but seems to be working fantastic on the docs site as well. Nice work!

packages/sage-system/lib/dropdown.js
```diff
@@ -167,8 +167,63 @@ Sage.dropdown = (function () {
...
+    const docEl = window.document.documentElement;
+
+    panel.style.top = ''; // resets the style
+
+    // Dimensions
+    const buttonDimensions = button.getBoundingClientRect();
+    const panelDimensions = panel.getBoundingClientRect();
+
+    const panelNewLoc = {
+      top: buttonDimensions.height + panelDimensions.height
```
monkeypox8, line 197: Since we're applying this to all dropdown instances, thoughts on accounting for the bottom `margin` in the calculation here? We'd be able to more closely match the default position.

margin

packages/sage-system/lib/dropdown.js
monkeypox8: I'm in favor of leaving it for a follow-up. Seems useful for views using the Drawer component but not an immediate need for this fix.

monkeypox8 (approved):
👍 LGTM! Noted one item to consider re: the `top` position calculation

packages/sage-system/lib/dropdown.js
QuintonJason: Then you're good to go

QuintonJason (approved):
Great addition

PR #1694, opened 2023-02-23
[DSS-300] Button and Toolbar style corrections

monkeypox8:
## Description
Addresses misalignment of Toolbar child elements and associated specs 

[Related KP PR](https://github.com/Kajabi/kajabi-products/pull/27262) modifies shims to match these updates.

General (Button and Toolbar)
- Realigns base Button `line-height` and `padding` to Figma specs (see [DSS-351](https://kajabi.atlassian.net/browse/DSS-351))
- Remove patch values (KP areas to be addressed separately) 
- Button base padding values moved to SCSS variable
- Increases overall Toolbar `border-radius` to match `10px` Figma value (previously grouped with Panel Controls)

Grouped Toolbar search fields
- Apply hover and focus states to match grouped Button styles

Grouped Toolbar input fields
- Corrects `z-index` hover issue
- Adjusts button absolute position offset
- Default form input labels in Toolbar as visually hidden (enforces vertical alignment)

Toolbar Variant: Rich Text Editor
- Scoped SCSS var for RTE Toolbar Button padding values

## Screenshots

|  Before  |  After  |
|--------|--------|
|toolbar-before|toolbar-after|

## Testing in `sage-lib`
1. Navigate to the Toolbar component in [Rails docs](http://localhost:4000/pages/component/toolbar?tab=preview) and [Storybook](http://localhost:4100/?path=/docs/sage-toolbar--default)
2. Verify all child components are in vertical alignment and display equal heights

1. (**HIGH**) https://github.com/Kajabi/sage-lib/pull/1694 Modifies Toolbar and Button line-height and padding values
   - [ ] General: confirm existing Sage buttons align with any associated form fields
   - [ ] Contacts list Search/Filter toolbar: verify Toolbar buttons and form fields align vertically with equal height(s)*

[*] Bulk selection checkbox/dropdown is part of Panel Controls; to be corrected in DSS-353

## Related

- DSS-300
- DSS-351
- DSS-353

[DSS-351]: https://kajabi.atlassian.net/browse/DSS-351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_form_input.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_form_input.scss
@@ -33,8 +33,6 @@ $-input-text-height: sage-font-size(body);
   .sage-toolbar__group & {
     &:first-child .sage-input__field {
       border-radius: sage-border(radius-medium);
-      border-top-right-radius: 0;
-      border-bottom-right-radius: 0;
     }
 
     &:last-child .sage-input__field {
@@ -106,6 +104,11 @@ $-input-text-height: sage-font-size(body);
 
   grid-area: label;
   margin-bottom: sage-spacing(xs);
+
+  // input labels within toolbar groups are visually hidden to maintain alignment
+  .sage-toolbar__group & {
+    @include visually-hidden();
+  }
 }
 .sage-input__field-wrapper {
   position: relative;
@@ -135,7 +138,7 @@ $-input-text-height: sage-font-size(body);
     &:hover,
     &:focus,
     &:active {
-      z-index: sage-z-index(default, 1);
+      z-index: unset;
     }
   }
 }
--- a/packages/sage-assets/lib/stylesheets/components/_form_input.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_form_input.scss
@@ -31,8 +31,10 @@ $-input-text-height: sage-font-size(body);
   }
 
   .sage-toolbar__group & {
-    &:first-child .sage-input__field {
+    &:first-child:not(:only-child) .sage-input__field {
       border-radius: sage-border(radius-medium);
+      border-top-right-radius: 0;
+      border-bottom-right-radius: 0;
     }
 
     &:last-child .sage-input__field {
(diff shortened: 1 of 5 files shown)


pixelflips (approved):
Very nice update and fix on this one! No concerns here. LGTM! 👍🏼 

I found no usage outside of the documentation but I just wanted to note that in the  `SageToolbarGroup` examples the sign up button overlaps the input. We may need to set padding on the input like in the input group component. 

Definitely for a future follow-up only (or may documentation update) and does not block this one.

Screenshot 2023-02-27 at 12 09 21 PM

Great work!

packages/sage-assets/lib/stylesheets/components/_form_input.scss
```diff
@@ -135,7 +138,7 @@ $-input-text-height: sage-font-size(body);
     &:hover,
     &:focus,
     &:active {
-      z-index: sage-z-index(default, 1);
+      z-index: unset;
```
monkeypox8, line 141: This pattern is not in use anywhere except documentation. 

However, there's a related issue where the `hover` state overrides `focus` styles in any non-search input form fields used in Toolbar (See the `Mixed content toolbar in a SageToolbarGroup` in Rails for an example). 

Seems to be due to an overall specificity/ordering issue in Form Input from the compiled SCSS. Holding off on a change here since there aren't any use cases so far (the more widely used Search input applies separate styles and doesn't have the same issue).

monkeypox8:
> I found no usage outside of the documentation but I just wanted to note that in the `SageToolbarGroup` examples the sign up button overlaps the input. We may need to set padding on the input like in the input group component.

Great catch! I'm actually not sure where the original absolute position for the button came from, might have been inherited from Panel Controls. The Figma mockup shows the input field as self-contained, rather than overlapping the button like with Input Group, so I'll remove the absolute positioning to match.

ju-Skinner (approved):

PR #1703, opened 2023-03-06
[DSS-352]: Replace Help link focus state legacy styles

pixelflips:
## Description
Updates legacy focus-related styles applied to help icon link.

## Screenshots

||  Before  |  After  |
|--------|--------|--------|
|Docs|Screenshot 2023-03-06 at 10 09 59 AM|Screenshot 2023-03-06 at 10 10 12 AM|
|KP|Screenshot 2023-03-06 at 9 31 23 AM|Screenshot 2023-03-06 at 9 30 45 AM|

## Testing in `sage-lib`

- Navigate to Page Heading
- Apply focus to help link in the "Page Heading with Help Link" section.
- Verify styling is updated and correct.

1. (**LOW**) Updates focus styling on help icon links. Styling only updates.
   - [ ] Products > All Products
   - [ ] Coupon > Edit Coupon

## Related
DSS-352

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_link.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_link.scss
@@ -8,6 +8,7 @@ $-link-color: sage-color(primary, 300);
 $-link-color-hover: sage-color(primary, 400);
 $-link-subtext-color: sage-color(charcoal, 200);
 $-link-subtext-color-hover: sage-color(charcoal, 400);
+$-link-help-icon-only-size: rem(20px);
 
 $-link-base-styles: (
   primary:(
@@ -198,8 +199,14 @@ $-link-base-styles: (
 }
 
 .sage-link--help-icon-only {
-  @include sage-focus-outline($outline-offset-inline: 4px, $outline-offset-block: 2px);
-  @include sage-focus-outline--update-color($-link-color);
+  @include sage-focus-outline(
+    $outline-offset-inline: 4px,
+    $outline-offset-block: 4px,
+    $outline-border-radius: sage-border(radius-round),
+  );
+  @include sage-focus-outline--update-color(sage-color(primary, 200));
+  width: $-link-help-icon-only-size;
+  height: $-link-help-icon-only-size;
 }
 
 .sage-link--remove-underline {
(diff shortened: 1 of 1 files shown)


packages/sage-assets/lib/stylesheets/components/_link.scss
```diff
@@ -8,6 +8,7 @@ $-link-color: sage-color(primary, 300);
 $-link-color-hover: sage-color(primary, 400);
 $-link-subtext-color: sage-color(charcoal, 200);
 $-link-subtext-color-hover: sage-color(charcoal, 400);
+$-link-help-icon-only-size: rem(20px);
```
monkeypox8, line 11: So, this is an existing issue, but one that gets more noticeable if we decide to set a link size. Since help links are mainly used inline, their outline will overlap smaller text. Coupons and the Products & Offers Automations panel are two examples that come to mind. We may want to consider inheriting the parent font size or applying an offset margin, though I suspect design may also have an opinion.

<details>
<summary>Coupon detail</summary>

</details>

pixelflips: > So, this is an existing issue, but one that gets more noticeable if we decide to set a link size. Since help links are mainly used inline, their outline will overlap smaller text. Coupons and the Products & Offers Automations panel are two examples that come to mind. We may want to consider inheriting the parent font size or applying an offset margin, though I suspect design may also have an opinion.
> 

@monkeypox8 Great callout and thanks for the heads up. Will ping design and get their input. Stay tuned.

pixelflips:
Spoke w/ @cameronsimony on the issue brought up by @monkeypox8 - for this round we will just add a margin that will help prevent the overlap when the focus state is applied.

Code adjustments are incoming!

packages/sage-assets/lib/stylesheets/components/_link.scss
pixelflips: @monkeypox8 spoke w/ Cam. He's fine with the margin. Leaving an additional comment below.

ju-Skinner (approved):

monkeypox8 (approved):
👍 LGTM

PR #1706, opened 2023-03-08
[DSS-281] - React Accordion has misaligned arrow

goodwinchris:
## Description
Changes to the `.sage-loader` button broke accordions. I've rescoped our `position: relative;` on the `.sage-loader` styles to ensure any time the loader is `.visually-hidden` it defers to `position: absolute;`. 

Note: The Linter also did its thing and cleaned up this file a bit. 

## Screenshots

|  Before  |  After  |
|--------|--------|
CleanShot 2023-03-08 at 09 57 57@2x|CleanShot 2023-03-08 at 09 57 08@2x

## Testing in `sage-lib`
1. [Open Storybook and navigate to the Accordion component ](http://localhost:4100/?path=/story/sage-accordion--single-panel)
2. Confirm other instances of `.sage-loader` haven't regressed

1. (**MEDIUM**) Change should be scoped to only when loader is expected to be visually hidden but due to the many uses of `sage-loader` throughout KP there could be adverse effects. A few places to check the loader is performing as expected: 
   - [ ] Loading contacts list on Contacts page
   - [ ] Loading recent transactions on Analytics page

## Related
[DSS-281: React Accordion has misaligned arrow](https://kajabi.atlassian.net/browse/DSS-281)

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_loader.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_loader.scss
@@ -4,7 +4,6 @@
 /// @group sage
 ////
 
-
 $-loader-bar-bg-color: sage-color(grey, 500);
 $-loader-bar-height: rem(6px);
 $-loader-bar-speed: 1.2s;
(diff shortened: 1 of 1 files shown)


pixelflips (approved):
LGTM! All seems to be in working order, nice job! 👍🏼 

Note: Since it's a medium change, it may be helpful in the "testing in KP" section to list a specific location(s) for QE to test/verify. That way they don't have to go fishing or ping us for an existing use case that they can check.

ju-Skinner (approved):

monkeypox8 (approved):
👍🏼 LGTM. afaik the Sage Accordion isn't used in KP. Worth double-checking to be certain, but this should be safe to note as docs-only.

PR #1701, opened 2023-03-03
[DSS-356] - Fix dropdown logic for safari

ju-Skinner:
## Description
It was discovered that the `sage-dropdown` was broken when using Safari. 

## Screenshots

|  Before  |  After  |
|--------|--------|
|https://www.loom.com/share/6c862d9f2fe942f787eab8ad3a5fe075|https://www.loom.com/share/e17f6684dcac41809c5317d694c4fb98|

** NOTE:
updated After loom video with the changes made in the [commit](https://github.com/Kajabi/sage-lib/pull/1701/commits/5565cc69e0d416e1838ab7b7f00219e1a562c6a0), which addresses @monkeypox8 concern.

## Testing in `sage-lib`
1. Start up the library by running `yarn start`
2. Open up Safari 
3. Go to the doc site by navigating to `localhost:4000`
4. Navigate to `Components -> Dropdown` or click [here](http://localhost:4000/pages/component/dropdown?tab=preview)
5. Scroll down to get a button to be at the bottom of the viewport. 
6. Click the button, dropdown should show the panel above the trigger. 

## Testing in `kajabi-products`
1. Navigate to a page e.g Products. 
2. Have enough products on the page so that a dropdown would be forced to render above the trigger.

PR #1712, opened 2023-03-09
[DSS-359] Updated React Dropdown positioning

QuintonJason:
## Description

Update the positioning of the dropdown component to be aligned with the work performed on the Rails side in [PR 1701](https://github.com/Kajabi/sage-lib/pull/1701)

#### `sage-lib`
|  Before  |  After  |
|--------|--------|
|reactTooltipPositioningBefore|reactTooltipPositioning|

#### `kp`
|  Before  |  After  |
|--------|--------|
|reactTooltipPositioningBeforeKp|reactTooltipPositioningAfterKp|

## Testing in `sage-lib`

Go to the React Dropdown page
- Scroll down to get a button to be at the bottom of the viewport.
- Click the button, dropdown should show the panel above the trigger.

1. (**MEDIUM**) Updates the React Dropdown content positioning.
   - Scroll down to get a button to be at the bottom of the viewport.
   - Click the button, dropdown should show the panel above the trigger. 

## Related

Closes [DSS-359](https://kajabi.atlassian.net/browse/DSS-359)

[DSS-359]: https://kajabi.atlassian.net/browse/DSS-359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cif

### DIFF
--- a/packages/sage-react/lib/Dropdown/Dropdown.jsx
+++ b/packages/sage-react/lib/Dropdown/Dropdown.jsx
@@ -91,7 +91,7 @@ export const Dropdown = ({
 
     // Elements
     const button = el;
-    const panel = el.lastElementChild
+    const panel = el.lastElementChild;
     const win = panel.ownerDocument.defaultView;
     const docEl = window.document.documentElement;
 
@@ -102,7 +102,7 @@ export const Dropdown = ({
     const panelDimensions = panel.getBoundingClientRect();
 
     const panelNewLoc = {
-      top: (buttonDimensions.height / 2 ) + panelDimensions.height
+      top: (buttonDimensions.height / 2) + panelDimensions.height
     };
 
     const viewport = {
@@ -117,7 +117,7 @@ export const Dropdown = ({
     };
 
     const panelHeight = getHeight(panel);
-    const enoughSpaceAbove = viewport.top < ( offset.top + panelHeight);
+    const enoughSpaceAbove = viewport.top < (offset.top + panelHeight);
     const enoughSpaceBelow = viewport.bottom > (offset.bottom + panelHeight);
 
     if (!enoughSpaceBelow && enoughSpaceAbove) {
--- a/packages/sage-react/lib/Dropdown/Dropdown.jsx
+++ b/packages/sage-react/lib/Dropdown/Dropdown.jsx
@@ -73,7 +73,7 @@ export const Dropdown = ({
 
     // Elements
     const button = el;
-    const panel = el.lastElementChild
+    const panel = el.lastElementChild;
     const win = panel.ownerDocument.defaultView;
     const docEl = window.document.documentElement;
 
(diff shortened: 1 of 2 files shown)


packages/sage-react/lib/Dropdown/Dropdown.jsx
```diff
@@ -55,41 +51,90 @@ export const Dropdown = ({
...
-      left: align !== 'right'
-        ? rect.left + inlineBoxOffset
-        : 'intitial',
-      right: align === 'right'
-        ? window.innerWidth - rect.right + inlineBoxOffset
-        : 'intitial',
-    });
+  const onUpdate = useCallback(debounce(() => setPanelCoords(), 20), []); // eslint-disable-line
+
+  // NOTE: getHeight and positionElement must be kept in alignment
```
QuintonJason, line 56: @Kajabi/ux-devs Added a NOTE stating to keep the two functions in those files in sync since we don't currently have a way to share assets between rails and react.

packages/sage-system/lib/dropdown.js
```diff
@@ -167,6 +167,10 @@ Sage.dropdown = (function () {
     }
   }
 
+
```
QuintonJason, line 170: Added a NOTE stating to keep the two functions in those files in sync since we don't currently have a way to share assets between rails and react.

ju-Skinner (approved):
Looks good to me...Here is a [video](https://www.loom.com/share/6bea742a04854806898ed477c18d62e9) where I tested locally with the bridge on.

packages/sage-react/lib/Dropdown/Dropdown.jsx
```diff
@@ -71,6 +71,66 @@ export const Dropdown = ({
...
+    const borderTopWidth = parseFloat(styles.borderTopWidth);
+    const borderBottomWidth = parseFloat(styles.borderBottomWidth);
+    const paddingTop = parseFloat(styles.paddingTop);
+    const paddingBottom = parseFloat(styles.paddingBottom);
+
+    return height - borderBottomWidth - borderTopWidth - paddingTop - paddingBottom;
+  };
+
+  const positionElement = (el) => {
+    let direction = null;
```
monkeypox8, line 90: Chrome/Firefox look great, Safari seems to have a strange hiccup when calculating the offset. I'm not sure if this would help, but would setting this to `direction = 'below'` (instead of `null`) as the default decrease the timing?

monkeypox8 (approved):
👍 LGTM, had one question about the default for `direction` re: Safari, but not a blocker. Nice work!

packages/sage-react/lib/Dropdown/Dropdown.jsx
monkeypox8: Think I answered my own question. Setting the default direction doesn't seem to change anything with Safari, so I think we're good to go

PR #1715, opened 2023-03-13
update SageComponent rendering and ProgressBar properties to support Turbo Stream rendering

kellyredding:
#### fix(sage component): explicit force html format when rendering Sage components
    
    When attempting to render a component from a Turbo Stream response
    format an `ActionView::MissingTemplate` error is raised, e.g.
    
    Given (in e.g. `/.../show.turbo_stream.erb`):
    ```html
    <turbo-stream action="update" target="progress-bar">
      <template>
        <%= sage_component SageProgressBar, {
          percent: @view.progress_percent,
          label: "Progress",
        } %>
      </template>
    </turbo-stream>
    ```
    
    Raises:
    ```
    ActionView::MissingTemplate - Missing partial
    sage_components/_sage_progress_bar with
    {
      :locale=>[:en],
      :formats=>[:turbo_stream],
      :variants=>[],
      :handlers=>[:raw, :erb, :html, :builder, :ruby, :jbuilder]
    }. Searched in:
      * "/Users/kellyredding/projects/kajabi/kajabi-products/app/views"
      * "/Users/kellyredding/.rubies/2.7.7/lib/ruby/gems/bundler/
         gems/sage-lib-1a13d8b0a7ca/docs/lib/sage_rails/app/views"
      * ...
    ```
    
    This switches to forcing the `formats:` directive on the render
    call when rendering Sage components (since all of the component
    templates are html). I got this from the Rails docs here:
    https://guides.rubyonrails.org/layouts_and_rendering.html#the-formats-option

#### feat(progress bar): add `animate` property to ProgressBar animation on/off
    
    When rendering progress bar updates using Turbo Streams, we are
    doing server side rendering. This means we re-render the progress
    bar component with a new percentage each time. This works fine
    except that the animation (that was previously always on) would
    run and slide the bar from zero to percentage each time it was
    re-rendered. This made for jarring progress bar updates b/c it
    looked like each update had to restart the progress from zero.
    
    I reached out to #sage-support and ended up pairing with Julian
    on adding a new `animate` boolean property to ProgressBar. This
    property (`nil` but "on" by default for backwards compatibility)
    controls whether the animation CSS is applied. If created with
    `animate: false`, the animation will be disabled and any render
    or re-render will just show the progress bar and the given percent
    without animation.
    
    This covers the both the Rails and React component adjustments to
    support the new property. This also updates the docs site to show
    the new property. Finally, I tested the property in both the Rails
    docs site page and the React storybook.

## Screenshots
Default animate "on" behavior still exists (backwards compatible change):
sage-animate-by-default

Animate false updating via Turbo Streams (via my spike):
sage-no-animate

Animate on/off in React storybook:
sage-react-demo

## Testing in `sage-lib`
1. Do the yarn setup/start stuff to run all the processes locally.
2. Fire up http://localhost:4000/pages/component/progress_bar?tab=preview and verify animate is still on by default (you should see the progress bars animate to the percentage
3. Fire up http://localhost:4100/?path=/story/sage-progressbar--default&args=percent:89;animate:true and toggle the animate switch from true to false and see the animation turn on/off on change.

Impact level: LOW
I was spiking a test to make sure we could update progress bar percents from Turbo Stream renders and I noticed these issues. There is no place in kajabi-products where this is used (yet) - I was just spiking a new usage. I've attached a gif above from my spike showing it working. The change is backwards compatible so it shouldn't affect any previous usage of progress bars. Let me know if you need more for this section.

## Related
N/A - I don't have a usage PR yet as I was just spiking for now.


pixelflips (approved):
All seems to be working great locally. Nice work and thanks for jumping on this one!! 👍🏼 LGTM!

ju-Skinner (approved):

kellyredding:
@ju-Skinner what else is needed on this to merge it? Are we good to go?

ju-Skinner:
> @ju-Skinner what else is needed on this to merge it? Are we good to go?

I believe we are waiting on one more review from the team. Since I helped contribute, my review didn't count 😉

anechol (approved):
🚀

PR #1725, opened 2023-03-28
[DSS-363]: Carousel buttons non-functional in React

pixelflips:
## Description
The Previous/Next button action in the React version of the Carousel does not function.

## Screenshots

|Carousel|
|--------|
||

## Testing in `sage-lib`
Navigate to [Carousel](http://localhost:4100/?path=/docs/sage-carousel--default)
Verify next and prev arrow now function

1. (**LOW**) Bug fix for Carousel arrows in React component.

## Related
DSS-363

### DIFF
--- a/packages/sage-react/lib/Carousel/Carousel.jsx
+++ b/packages/sage-react/lib/Carousel/Carousel.jsx
@@ -84,7 +84,7 @@ export const Carousel = ({
           disabled={arrowPrevDisabled}
           icon="caret-left"
           id="prev"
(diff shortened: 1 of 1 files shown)


QuintonJason (approved):
Thanks for updating this

pixelflips:
> Here is another approach and a little more common use of calling functions.

Appreciate the suggestion! Works great and I've updated it to use that approach. Thanks again!

ju-Skinner:
:shipit:

PR #1687, opened 2023-02-07
PRODENG-1577: Kube deployment

raynorg:
## Screenshots

|  Before  |  After  |
|--------|--------|
|||

## Testing in `kajabi-products`

1. (**LOW/MEDIUM/HIGH/BREAKING**) Description of the change and its impact with QA as the audience.
   - [ ] One more examples of the component in use to either test the change or verify the change has not had adverse effects.

PR #1737, opened 2023-05-02
ci: add/update environments for use of EKS

ju-Skinner:
Each environment will now have its own Documentation Site and Storybook URLs

PR that will leverage these new environments
https://github.com/Kajabi/sage-lib/pull/1687


docs/config/environments/staging.rb
rwc9u, line 1: The duplication is fine. One thing that kajabi-products does is require another environments settings and then overrride just the ones from that environment that are changing, so
qa < staging < production
Just another option to consider.
```ruby
require Rails.root.join("config/environments/production")

Rails.application.configure do
  AppConfig.primary_domain           = ENV.fetch("PRIMARY_DOMAIN", "newkajabi-staging.com")
  AppConfig.primary_domain_alternate = ENV.fetch("PRIMARY_DOMAIN_ALTERNATE", "newkajabi-staging-alternate.com")
# ... more overrides
end

rwc9u (approved):

docs/config/environments/staging.rb
ju-Skinner: I like that. It's much cleaner.

PR #1740, opened 2023-05-09
fix(react-button): suppressDefaultClass warning

ju-Skinner:
## Description
Warning occurs when the property was being set on a HTML button

 The conditional fixes that by checking if it's a Link component

fix: DSS-374


ju-Skinner:
I know we closed this issue, but I forgot I already had a branch with the fix. If we still feel like we want to leave it to rest easy, then we can. 

cc: @monicawheeler

monkeypox8 (approved):
🔥 🪓 🥇 you've defeated the kraken! they'll sing of this feat in valhalla

PR #1741, opened 2023-05-15
fix(alert): grid template areas

ju-Skinner:
Jira Ticket:  https://kajabi.atlassian.net/browse/DSS-405

## Description
When the Alert has no actions, the additional row was still being displayed. This is not the intended behavior. 

## Screenshots

|  Before  |  After  |
|--------|--------|
image|
image

## Testing in `sage-lib`
1. Navigate to Storybook -> Alert
2. Verify that there is no additional row added when the alert has no Actions. 

I need to find a location in the app that has an Alert with no action buttons. At this time I'm unaware of it. I'll try to update prior to merging.

### DIFF
--- a/packages/sage-react/lib/Alert/Alert.jsx
+++ b/packages/sage-react/lib/Alert/Alert.jsx
@@ -28,6 +28,7 @@ export const Alert = ({
       [`sage-alert--${color}`]: color,
       'sage-alert--dismissable': dismissable,
       'sage-alert--small': small,
+      'sage-alert--actions': actions !== null,
     }
   );
 
--- a/docs/lib/sage_rails/app/views/sage_components/_sage_alert.html.erb
+++ b/docs/lib/sage_rails/app/views/sage_components/_sage_alert.html.erb
@@ -4,6 +4,7 @@
     <%= "sage-alert--#{component.color}" if component.color %>
     <%= "sage-alert--dismissable" if component.dismissable %>
     <%= "sage-alert--small" if component.small %>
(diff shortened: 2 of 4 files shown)


packages/sage-react/lib/Alert/Alert.jsx
```diff
@@ -28,6 +28,7 @@ export const Alert = ({
       [`sage-alert--${color}`]: color,
       'sage-alert--dismissable': dismissable,
       'sage-alert--small': small,
+      'sage-alert--actions': actions !== null,
```
pixelflips, line 31: I believe this class will need to be added to the rails component as well. It's currently missing and breaking the layout.

pixelflips, line 31: I believe this class will need to be added to the rails component as well. It's currently missing and breaking the layout of the actions and dismiss icon.

Screenshot 2023-05-15 at 10 56 27 AM

pixelflips (changes requested):
Seeing some minor layout issues on the Rails side and the linter is yelling a bit. After those are addressed, it should be good to go.

packages/sage-react/lib/Alert/Alert.jsx
ju-Skinner: Yeah, I failed to do the rails component. Glad things worked on the React side, so it's just a matter of doing the same for Rails. Thanks

pixelflips (approved):
I'd recommend checking w/ Design on the missing title by default, but definitely not a show-stopper. Even if it's just to alert them 🤣 that we now support that use case.

LGTM! 👍🏼

QuintonJason (approved):
Looks good to me

PR #1762, opened 2023-07-13
fix(react-dropdowns): add `...rest` to allow additional props

ju-Skinner:
fixes: https://kajabi.atlassian.net/browse/DSS-433

## Description
The `SelectDropdown` did not allow different properties to be used. This was in part of not accepting `...rest`. So we couldn't use properties such as `isPinned`, `customTrigger`, to name a few. In order to fix this, I elected to add `...rest` to the list of props to `SelectDropdown`, `OptionsDropdown`, and `ToolbarDropdown`. This should prevent any future issues with props missing.

## Screenshots

|  Before  |  After  |
|--------|--------|
|image|image|

1. Check out this branch `DSS-433/sage-react-dropdown-position-fix` in KP
2. Navigate to `/admin/sites/1/contacts?`
3. Make your viewport narrow so that it would attempt to render the dropdown on the right side. It should technically be impossible to replicate the `before` picture.

### DIFF
--- a/packages/sage-react/lib/Dropdown/OptionsDropdown.jsx
+++ b/packages/sage-react/lib/Dropdown/OptionsDropdown.jsx
@@ -15,6 +15,7 @@ export const OptionsDropdown = ({
   panelSize,
   options,
   triggerButtonSubtle,
+  ...rest // Used as an option to pass other props to Dropdown (base) component
 }) => (
   <Dropdown
     align={align}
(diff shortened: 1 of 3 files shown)


pixelflips (approved):
LGTM! Thanks for taking these on!!

QuintonJason (approved):
Looks and works great!

anechol (approved):
LGTM

PR #1770, opened 2023-07-31
feat: add component attributes to sage_form_select component

ju-Skinner:
## Description
Per discussion in Slack, see [here](https://kajabi.slack.com/archives/C01A424HY8Y/p1690384266736959) for more details; it was requested to have the ability to add or subscribe to `onChange` events for the `sage_form_select` component in Rails. 

fixes: https://kajabi.atlassian.net/browse/DSS-444

## Testing in `sage-lib`
1. Navigate to `Form Select` in your local instance of the Doc Site e.g  http://localhost:4000/pages/component/form_select?tab=preview
2. Scroll to the bottom and interact with the select titled `Example on how to pass additional attributes to the component`
3. Click on the Code Tab to see how to implement it.

PR #1769, opened 2023-07-25
[DSS-441] Expandable Card Update

pixelflips:
## Description
Updates Expandable Card component to allow for the following:

- Adds an align trigger prop (moves trigger to allow for custom header)
- Adds a slot to allow for custom header content
- Documentation/Notes in storybook
- Documentaiton/Notes in docs site

## Screenshots

|  Demo of aligning trigger + custom header  |
|--------|
||

## Testing in `sage-lib`

Rails:

- Navigate to [Expandable Card](http://localhost:4000/pages/component/expandable_card?tab=preview)
- See the Custom Header section and verify it functions as expected.

React:
- Navigate to [Expandable Card](http://localhost:4100/?path=/docs/sage-expandablecard--default)
- See the Custom Header story and verify it functions as expected.

## Testing in `kajabi-products`

1. (**LOW**) Updates expandable card to support custom header content and trigger alignment.
   - [ ] These are new props and slot and should not have any effect. Quick sanity of areas using Accordian or Expandable card should be checked.

## Related
https://kajabi.atlassian.net/browse/DSS-441

PR #1772, opened 2023-07-31
Next Version Bump

ju-Skinner:
1. (**LOW**) - #1770 - Allows the ability to add additional attributes to the Components' element.
2. (**LOW**) - #1769 - Updates expandable card to support custom header content and trigger alignment.
   - [ ] These are new props and slot and should not have any effect. Quick sanity of areas using Accordian or Expandable card should be checked.


anechol (approved):

pixelflips (approved):
:shipit:

PR #1773, opened 2023-08-01
docs(sage-component): add to the list of components on the rails doc site

ju-Skinner:
## Description
With the addition of a new prop to `SageComponent`, we need the ability to share the details with our customers. This will add the base `SageComponent` to the Rails doc site. 

## Screenshots
image

## Testing in `sage-lib`
1. Navigate to the Rails doc site e.g http://localhost:4000
2. Click on "Components", Sage Component should be the first component in the list. This was done intentionally since it serves as the base component for all Rails Sage components. 
3. Click on the `Props` tab, you should see the same details as the screenshot above.

No Impact

### DIFF
--- a/docs/app/helpers/components_helper.rb
+++ b/docs/app/helpers/components_helper.rb
@@ -13,6 +13,12 @@ module ComponentsHelper
     [
       # Sage Generated Components
       {
+        title: "sage_component",
+        description: "This serves as the base component for all Rails components",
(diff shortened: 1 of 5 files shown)


pixelflips (approved):
👍🏼 LGTM! This is a great addition, thanks for adding!

QuintonJason (approved):
Great Additions

PR #1803, opened 2023-09-26
fix(react-input): allow ref to be forwarded to component

ju-Skinner:
## Description
In a support request, it was discovered the `ref` was not working as expected. This solves the issue now. 

# Implemenation example.
```
import { Input } from '@kajabi/sage-react'

const inputRef = React.useRef(null);

React.useEffect(() => {
  if (inputRef.current) {
    inputRef.current.focus();
  }
}, []);

return (
  <Input onFocus={onFocus} ref={inputRef} />
);
```

Video Reference:  Hopefully the audio was recorded. 
https://www.loom.com/share/6ceb6bb9b0284072b147a5c442e09191

## Testing in `sage-lib`
1. You can load storybook locally and navigate to `http://localhost:4100/?path=/story/sage-input--input-focus-test`
2. Result...input should be focused. Clicking off it will show the placeholder text.

### DIFF
--- a/packages/sage-react/lib/Input/Input.jsx
+++ b/packages/sage-react/lib/Input/Input.jsx
@@ -8,7 +8,7 @@ import {
   INPUT_TYPE
 } from './configs';
 
-export const Input = ({
+export const Input = React.forwardRef(({
   autocomplete,
   className,
   disabled,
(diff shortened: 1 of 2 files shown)


safa-kajabi (approved):
Looks good to me! -- Approved!

pixelflips (approved):
LGTM! Thanks so much for jumping on this one!

PR #1857, opened 2024-01-11
STANDARD: Node 18

voodooGQ:
## Description
This PR integrates Node18 as the primary node version for the monorepo. Support for 16 is still active in the CI lint-test-build workflow as Kajabi Products is still on Node16 for awhile.

Communities reached out to @Kajabi/standard-model because they were getting warnings in Netlify about the support for 16. Although these are just warnings they may turn into hard blockers over time. In addition, we're on track to get Node18 into the Monolith with-in the next couple months. Assuming `pine` is not ready and available by that time this effort would be needed regardless to support the monolith.

> NOTE: In order for Node18 to work with the current dependencies `NODE_OPTIONS=--openssl-legacy-provider"` must be passed. This was added to the `.env-dist`, but DSS will have to update their `.env` with the value OR pass it along when running commands. Unfortunately, because this puts it on an unsupported openssl schema, `vscode` (and possibly other editors) will error out with `--openssl-legacy-provider is not allowed in NODE_OPTIONS`. In order to address this I simply ran `unset NODE_OPTIONS && code .` when opening from my terminal. Not a great solution, may be a better path, but seems to be the community accepted answer for the time being until we get everything else updated.

Also removed all CircleCI references in the docs / integrations since this is no longer used.

## Testing in `sage-lib`
1) Ran `yarn start`, verified all documentation nodes were loading as appropriate
2) Ran `yarn test` and allowed the test suite to complete successfully
  - There are very limited tests, this concerns me a bit that something may have "broken", but that's what QE is for right? :)
3) Updated `lint-test-build.yaml` workflow to run a matrix of 16 and 18 Node versions, verified a pass.

This changes the underlying Node version that the suite runs on. In theory, because the build/test automations all worked, we can assume that the underlying components are working as well. That being said, wouldn't hurt to do a "sweep" to verify we're not seeing anything funky.

## Related
https://kajabi.slack.com/archives/C031S5GTM63/p1704895683581299


.github/workflows/lint-test-build.yml
```diff
@@ -6,6 +6,9 @@ jobs:
   lint_test_build:
     name: Lint, Test, Build
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [16, 18]
```
ju-Skinner, line 11: You can put this in a repo variable and have it read from there. This way it doesn't need to be another PR to update the versions. I would even go as far as to make all uses to read from the variable and pass it along to the composite workflows. 

You can look [here](https://github.com/Kajabi/pine/blob/next/.github/workflows/build.yml#L15-L26) for examples.

ju-Skinner (approved):

safa-kajabi (approved):
This works well and I don't have anything blocking outside of things Julian already commented -- Approved 🙌

voodooGQ:
@ju-Skinner, I updated the workflow to give a different command when using Node 18 (specifically adding the `NODE_OPTIONS` as discussed above). I _think_ everything is good-to-go now, but can you go through the [Node 16](https://github.com/Kajabi/sage-lib/actions/runs/7493524337/job/20399413577?pr=1857) and [Node 18](https://github.com/Kajabi/sage-lib/actions/runs/7493524337/job/20399413913?pr=1857) workflow runs and just see if anything catches your eye? 

If not I think we can get it merged.

ju-Skinner:
> @ju-Skinner, I updated the workflow to give a different command when using Node 18 (specifically adding the `NODE_OPTIONS` as discussed above). I _think_ everything is good-to-go now, but can you go through the [Node 16](https://github.com/Kajabi/sage-lib/actions/runs/7493524337/job/20399413577?pr=1857) and [Node 18](https://github.com/Kajabi/sage-lib/actions/runs/7493524337/job/20399413913?pr=1857) workflow runs and just see if anything catches your eye?
> 
> If not I think we can get it merged.

Looks good to me. :shipit:

voodooGQ:
100% totally forgot about this. Will set some time aside this week to get it in play

PR #1858, opened 2024-01-16
fix(popover): Add "active" prop 

pixelflips:
## Description

Adds `active` prop to component to allow to show and hide.

## Screenshots
No visual changes are expected with these updates.

## Testing in `sage-lib`

- Navigate to [popover](http://localhost:4100/?path=/docs/sage-popover--default)
- Verify `active` prop exists and functions as expected.
- Verify popover still functions as expected when interacted with.

1. (**LOW**) Adds new active prop to Popover (React).

## Related
https://kajabi.atlassian.net/browse/DSS-501

### DIFF
--- a/packages/sage-react/lib/Popover/Popover.jsx
+++ b/packages/sage-react/lib/Popover/Popover.jsx
@@ -7,6 +7,7 @@ import { SageTokens } from '../configs';
 import { POPOVER_POSITIONS } from './configs';
 
 export const Popover = ({
+  active,
   className,
   children,
   customContentClassName,
@@ -20,7 +21,7 @@ export const Popover = ({
   title,
   ...rest
 }) => {
-  const [selfActive, setSelfActive] = useState(false);
+  const [selfActive, setSelfActive] = useState(active);
 
   const handleExpandClick = () => {
     setSelfActive(true);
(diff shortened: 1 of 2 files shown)


packages/sage-react/lib/Popover/Popover.jsx
```diff
@@ -20,7 +21,7 @@ export const Popover = ({
   title,
   ...rest
 }) => {
-  const [selfActive, setSelfActive] = useState(false);
+  const [selfActive, setSelfActive] = useState(active);
```
ju-Skinner, line 24: I know you didn't write the initial line of code, but we should change the `selfActive` name to `isActive`. We should follow the same for `setSelfActive`, and rename it to `setIsActive`.

pixelflips: @ju-Skinner Updated. Great call, that does read much cleaner!

packages/sage-react/lib/Popover/Popover.jsx
```diff
@@ -20,27 +21,27 @@ export const Popover = ({
   title,
   ...rest
 }) => {
-  const [selfActive, setSelfActive] = useState(false);
+  const [isActive, setisActive] = useState(active);
```
ju-Skinner, line 24: `setisActive` should be camel-cased `setIsActive`.

pixelflips: doh! all lowercase is a hard habit to break. Should be good now, thanks for catching it!

ju-Skinner (approved):
Great job. Thanks for taking this on.

PR #1921, opened 2024-08-01
revert(select): removes size prop from pds-icon

ju-Skinner:
## Description
Removes using the size attribute to resolve the pds-icon overwritten using the internal style attribute.

## Screenshots
No Visual Changes

## Testing in `sage-lib`
1. Go to Components -> Form Select in Rails Doc. 
2. Navigate to storybook -> Select 
3. In both steps, verify the icon is aligned.


pixelflips (approved):
LGTM!

PR #1945, opened 2024-08-13
chore: update node to v18.20.4

ju-Skinner:
## Description
* Removes node 16
* Updates to the latest stable version of node 18.

## TODO
Update Action matrix variables to match `package.json:engine`, when merged

https://nodejs.org/en/about/previous-releases

## Dev Instructions
* install version 18.20.4 via your package manager

### ASDF
`asdf install nodejs 18.20.4`

### NVM
`nvm install 18.20.4`


voodooGQ (approved):
Probably arbitrary, but we have the monolith on `18.19.1`. Obviously you don't need to be in parity with that if you don't want, but just want to pass along in case.

Looks good!

pixelflips (approved):

PR #1953, opened 2024-08-20
Update the workflow/Dockerfile for Storybook

voodooGQ:
@ju-Skinner and I spent many...many...many hours trying to get the Dockerfile to properly build-out the Storybook site. Despite our efforts we consistently ended up with missing files for seemingly no reason and therefore would get errors. Sage is becoming legacy, and junking a bunch more time into getting it to work "right" is counter-intuitive. Instead what we're going to do is build on the Github Actions runner and then just copy over the file to the Docker image as we know the "local builds" are working correctly.

This still has to be tested, but should accomplish our goals.


ju-Skinner (approved):
:shipit:

PR #2012, opened 2024-10-18
fix: tooltip custom class (react)

pixelflips:
## Description
Currently, there is no hook into the react tooltip to allow custom adjustments.
This adds a `tooltipCustomClass` prop to allow devs a hook.

This comes as a request from ds-support.

## Screenshots
Screenshot 2024-10-18 at 3 13 41 PM

## Testing in `sage-lib`

- Navigate to [tooltip](http://localhost:4100/?path=/docs/sage-tooltip--default)
- Verify custom class prop is available
- Verify story is correct

1. (**LOW**) Adds custom class prop to react tooltip.

## Related
TBD

### DIFF
--- a/packages/sage-react/lib/Tooltip/Tooltip.jsx
+++ b/packages/sage-react/lib/Tooltip/Tooltip.jsx
@@ -2,14 +2,13 @@ import React, { Children, useState, cloneElement } from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import { TooltipElement } from './TooltipElement';
-import {
-  TOOLTIP_POSITIONS
-} from './configs';
+import { TOOLTIP_POSITIONS } from './configs';
 
 export const Tooltip = ({
   children,
   content,
   position,
+  tooltipCustomClass,
   ...rest
 }) => {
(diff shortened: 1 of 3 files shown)


packages/sage-react/lib/Tooltip/Tooltip.jsx
```diff
@@ -2,14 +2,13 @@ import React, { Children, useState, cloneElement } from 'react';
...
-import {
-  TOOLTIP_POSITIONS
-} from './configs';
+import { TOOLTIP_POSITIONS } from './configs';
 
 export const Tooltip = ({
   children,
   content,
   position,
+  tooltipCustomClass,
```
pixelflips, line 11: def open to naming suggestions, but I notice `tabChoiceCustomClass` and went with a similar approach.

pixelflips, line 11: I'm definitely open to naming suggestions. I noticed `tabChoiceCustomClass` and went with a similar approach.

ju-Skinner (approved):

monicawheeler (approved):
👍🏻 Appreciate the warning in the tooltip example!

PR #2013, opened 2024-10-22
style: removes underline on tabs filter variant

anechol:
## Description

Removes underline on tabs filter variant.

## Screenshots

|  Before  |  After  |
|--------|--------|
|||
Screenshot 2024-10-22 at 4 44 00 PM|Screenshot 2024-10-22 at 4 42 32 PM

## Testing in `sage-lib`

- Navigate to [Tabs](http://localhost:4000/pages/component/tabs?tab=preview)
- For the filter variant, click and drag to activate the focus state, or activate focus state in the inspector tools
- Check that there's no mercury outline on the element

1. (**LOW**) Removes underline on tabs filter variant.

## Related

[DSS-1109](https://kajabi.atlassian.net/browse/DSS-1109)

[DSS-1109]: https://kajabi.atlassian.net/browse/DSS-1109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_tab.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_tab.scss
@@ -31,6 +31,10 @@ $-tab-color-disabled: map-get($sage-tab-colors, disabled);
     background-color: sage-color(grey, 200);
     border-radius: sage-border(radius-x-large);
     @extend %t-sage-body-small-semi;
+
+    &::after {
+      display: none;
+    }
   }
 
   @extend %t-sage-body-semi;


packages/sage-assets/lib/stylesheets/components/_tab.scss
```diff
@@ -31,6 +31,10 @@ $-tab-color-disabled: map-get($sage-tab-colors, disabled);
     background-color: sage-color(grey, 200);
     border-radius: sage-border(radius-x-large);
     @extend %t-sage-body-small-semi;
+
+    &::after {
+      display: none;
+    }
```
pixelflips, line 37: This works and is not a blocker but could/should it live nested in the focus around line 49? I think that's where is was before, but focus changed to focus-visible.

pixelflips (approved):
On non-blocking comment, otherwise LGTM!

monicawheeler (approved):

packages/sage-assets/lib/stylesheets/components/_tab.scss
anechol: I considered that initially but chose those route to remove the after tag wholesale instead of just on focus.

QuintonJason (approved):

PR #2046, opened 2025-01-30
style: add selector for icon-only buttons to not rely on icon name

pixelflips:
## Description
Since we no longer update the icon list in Sage, this has caused a padding issue on the icon-only buttons when new icons are added to the system.

Currently, the generated class names include the icon name. We should adjust the style selector so the icon name is unnecessary since we no longer update the list of icons in Sage.

This PR adjusts the CSS selector so the icon name is not required.

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screenshot 2025-01-30 at 12 06 54 PM|Screenshot 2025-01-30 at 12 06 38 PM|

## Testing in `sage-lib`

- Navigate to docs site
- On one of the icon-only buttons change the class name to an icon that does not exist (`sage-btn--icon-only-file-search`)
- Verify that the padding is correct on the button and it remains round.

1. (**LOW**) style fix for icon-only buttons that have new icons.

## Related
https://kajabi.atlassian.net/browse/DSS-1249

### DIFF
--- a/packages/sage-assets/lib/stylesheets/core/mixins/_sage.scss
+++ b/packages/sage-assets/lib/stylesheets/core/mixins/_sage.scss
@@ -113,7 +113,8 @@
 
   // --- Standalone Icon Buttons
   @else if $direction == only {
-    .sage-btn--icon-only-#{$icon-name} {
+    .sage-btn--icon-only-#{$icon-name},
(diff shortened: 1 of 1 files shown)


ju-Skinner (approved):
🔥

anechol (approved):

PR #2055, opened 2025-02-05
style: update textarea tokens to pine

pixelflips:
## Description
Converts sage tokens to pine tokens

## Testing in `sage-lib`
Navigate to Form textarea
Verify token updates

1. (**LOW**) Converts textarea tokens to Pine

## Related
https://kajabi.atlassian.net/browse/DSS-1270

### DIFF
--- a/packages/sage-assets/lib/stylesheets/components/_form_textarea.scss
+++ b/packages/sage-assets/lib/stylesheets/components/_form_textarea.scss
@@ -5,21 +5,17 @@
 ////
 
 // Border
-$-textarea-border-width: rem(1px);
+$-textarea-border-width: var(--pine-border-width);
 
 // Margin
-$-textarea-label-margin: sage-spacing(xs);
+$-textarea-label-margin: var(--pine-dimension-xs);
 
 // Padding
-$-textarea-label-padding: rem(3px);
-$-textarea-padding: sage-spacing(sm);
+$-textarea-padding: var(--pine-dimension-sm);
 
 // Sizing
 $-textarea-height: 100%;
-$-textarea-min-height: rem(90px);
-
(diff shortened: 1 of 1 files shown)


anechol:
This is kind of a weird one with the style mixin (sage-form-field) used. I can see a case for removing the mixin and updating the necessary styles to match so that nothing else is affected as this overall effort entails. **But** if we do that, then we wouldn't be reusing common styles across our form field components... 🙃 
What do you think about this?

pixelflips:
@anechol Honestly, I am not sure there are so many form mixins that I closed the file and ran. There are a lot of those mixins related to various input-related elements. I also don't know how those all play with things like simple form and the like. 

These types of cases are where I honestly would prefer updating the mixins vs component by component but without equivalent values it's tough. I suppose we can separate them all, but that feels like a pretty large effort along with a lot of QE testing. Let me know if you have other ideas.

anechol:
@pixelflips I'll throw the question in Slack again to discuss on Monday, since this particular case feels especially egregious.

anechol:
@pixelflips I spoke with @QuintonJason about this and he's updated the visual styles in the form field mixin for the Input, which will also touch this component. Since the mixin isn't being used directly in `kp`, I think we'll be okay going with that approach for this scenario.

anechol (approved):

pixelflips:
> @pixelflips I spoke with @QuintonJason about this and he's updated the visual styles in the form field mixin for the Input, which will also touch this component. Since the mixin isn't being used directly in `kp`, I think we'll be okay going with that approach for this scenario.

@anechol thanks for the info and checking into this.

QuintonJason (approved):

PR #2082, opened 2025-02-26
fix(banner): align icons and styles to updated design spec

pixelflips:
## Description
Banners have not been updated to match the latest design spec. A request was received and this PR make the following adjustments to match.

- Icons
- Typograpy
- Background colors

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screenshot 2025-02-25 at 5 07 08 PM|Screenshot 2025-02-25 at 5 06 51 PM|

## Testing in `sage-lib`
Navigate to Banner
Verify icons, typography, and colors are updated to match design spec.

1. (**LOW**) Updates banner icons and styling to match design spec.

## Related
https://kajabi.atlassian.net/browse/DSS-1301


packages/sage-assets/lib/stylesheets/components/_banner.scss
```diff
@@ -8,28 +8,32 @@
 $-banner-height-ladera: map-get($sage-banner-heights, ladera);
 $-banner-colors: (
   default: (
-    background: sage-color(primary, 300),
+    background: sage-color(purple, 600),
```
ju-Skinner, line 11: should these be converted to use the pine variable?

ju-Skinner (approved):
The comment that was left does NOT block my approval.

packages/sage-assets/lib/stylesheets/components/_banner.scss
pixelflips: Possibly, but then I wonder what the approach should be. Should I convert the entire CSS file or just the sections where the changes apply? I initially decided to update only the values for this PR and not convert them, but this is a good point to bring up.

@Kajabi/dss-devs, any additional thoughts on how we should handle these use cases where component updates are needed but fall outside the core token conversions we implemented?

Maybe it's a case-by-case basis based on effort or lift at the time of update?

I think I will convert the banner here (low lift), but did still want to this up with the group.

pixelflips:
@ju-Skinner since I included the var updates here. Going to request another review since the scope has changed. tia!

packages/sage-assets/lib/stylesheets/components/_banner.scss
anechol: I would say if the component needs to be updated, then it wouldn't hurt to update the affected tokens with it. It would save us headaches in the future.

anechol (approved):

ju-Skinner (approved):

PR #2089, opened 2025-03-12
chore: add environment to KP when Bridge is enabled

ju-Skinner:
## Description

The Sage Bridge was not consistently working. After moving in with @pixelflips, I finally found a solution to the problem. The environment variable is used to control which features we need to enable in Vite.


pixelflips (approved):
Thanks, roommate! LGTM!

monicawheeler (approved):

voodooGQ (approved):

PR #2121, opened 2025-05-06
refactor(inputaffixes): streamline affix handling and correct padding

pixelflips:
## Description

Form inputs with prefixes/suffixes that are hidden on page load (e.g., a hidden expandable div) receive incorrect padding when they become visible. This happens because their dimensions can't be measured when hidden, causing text overlap or misalignment once they become visible.

The `inputaffixes.js` script will now use an `IntersectionObserver` to calculate padding when an input becomes visible, ensuring correct dimensions. Padding will also now utilize CSS logical properties (paddingInlineStart/paddingInlineEnd) for automatic LTR/RTL support.

## Screenshots

|  Before  |  After  |
|--------|--------|
|||

## Testing in `sage-lib`
Navigate to the form input
Verify the prefix and suffix functions as expected

1. (**LOW**) Refactor input prefix/suffix padding JS. 
   - [ ] Sanity check any inputs with prefix/suffix

## Related
https://kajabi.atlassian.net/browse/DSS-1274

### DIFF
--- a/packages/sage-system/lib/inputaffixes.js
+++ b/packages/sage-system/lib/inputaffixes.js
@@ -11,47 +11,12 @@ Sage.inputaffixes = (() => {
   const valueClass = "sage-input__affix-value";
   const valueElement = "span";
   const inputPaddingOffset = 16;
+  const observers = []; // To store IntersectionObserver instances for cleanup
 
 
   // ==================================================
   // Functions
   // ==================================================
-
-  // Adds affix content to an element
-  const addAffixes = (el) => {
-    const elRoot = el;
-    const elInput = el.querySelector(`.${fieldClass}`);
-
-    // Toggle off the affixed class
-    elRoot.classList.remove(affixRootClass);
-
-    if (elRoot.dataset.jsInputPrefix) {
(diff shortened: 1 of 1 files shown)


QuintonJason:
It's good with me but I'll wait for the group. I had the same issue where the page loads and I need to position something based on an element, but the width and height are 0 because it comes in late. I chose to go the [requestAnimationFrame() route in my pine input prefix/suffix PR](https://github.com/Kajabi/pine/blob/beef6b5a487be65db4e1c4fee3b46984e05a31f1/libs/core/src/components/pds-input/pds-input.tsx#L162) so interested to see other solutions to the "late loading" issue and what is most agreed upon.

pixelflips:
@QuintonJason I ran it by the bots and got the following reply:

`IntersectionObserver` is preferred over `requestAnimationFrame` (rAF) for this fix because it's specifically designed for efficiently detecting when elements become visible. This event-driven approach is more performant than using rAF, which would necessitate continuously polling element properties (like offsetWidth) up to 60 times per second until visibility is confirmed. `IntersectionObserver` also leads to simpler, more declarative code for this visibility-dependent styling, avoiding manual state and loop management.

anechol (approved):
LGTM

PR #2136, opened 2025-07-16
fix(button): update XS group gap option to xs instead of null

pixelflips:
## Description

An issue with the Button Group XS gap not working as expected in the React component, and it appears that the token may be set to `null` in the `packages/sage-react/lib/Button/configs.js` file. 

`XS` token should be set to `xs` in config

## Screenshots

|  Before  |  After  |
|--------|--------|
|Screenshot 2025-07-16 at 1 00 22 PM|Screenshot 2025-07-16 at 1 01 18 PM|

## Testing in `sage-lib`
Navigate to [button group](http://localhost:4100/?path=/docs/sage-button-group--default)
Verify setting gap to XS works.

1. (**LOW**) Sets button group xs gap to correct "xs" value.

## Related
https://kajabi.atlassian.net/browse/DSS-1482

### DIFF
--- a/packages/sage-react/lib/Button/configs.js
+++ b/packages/sage-react/lib/Button/configs.js
@@ -11,7 +11,7 @@ export const BUTTON_ICON_POSITIONS = {
 };
 
 export const BUTTON_GROUP_GAP_OPTIONS = {
-  XS: null,
+  XS: 'xs',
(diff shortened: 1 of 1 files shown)


QuintonJason (approved):
🔬

PR #2199, opened 2026-06-02
ci: PE-3158 Depot vs GitHub runner A/B comparison

allantaylor8907:
## Summary
- Adds `.github/workflows/runner-comparison.yml`, a spike workflow that runs the lint/test/build pipeline twice in parallel — once on `ubuntu-24.04` (GitHub-hosted) and once on `depot-ubuntu-24.04` (Depot.dev) — and emits a side-by-side report in the job summary plus Datadog metrics.
- Triggers: `workflow_dispatch` (manual, with node version input) and `pull_request` **scoped to the workflow file itself** so it runs automatically on this PR but won't burn CI on every future PR after merge.
- Linear: [PE-3158](https://linear.app/kajabi/issue/PE-3158/spike-evaluate-depotdev-as-github-actions-replacement)

## What gets measured
Per runner: total wall-clock, `yarn install` time, estimated per-run cost.
Aggregated: % delta + cost savings, plus the metrics `ci.runner.duration_ms`, `ci.runner.install_time_ms`, `ci.runner.estimated_cost_usd`, `ci.runner.depot_improvement_pct` in Datadog.

## To make the Depot job actually run (needed before this can produce data)
The `depot-runner` job will sit `Queued` until the org is connected to Depot. None of this is in the PR diff:

1. **Create the Depot org account** → https://depot.dev, sign in with GitHub, create org for `Kajabi`.
2. **Install the Depot GitHub app on the Kajabi org** and grant it access to (at minimum) `Kajabi/sage-lib`. That's what makes `runs-on: depot-ubuntu-24.04` resolve.
3. **(Optional) DD_API_KEY repo secret** — without it the Datadog steps no-op, the job summary still works. If you want metrics flowing, set `DD_API_KEY` in repo settings → Secrets and variables → Actions.
4. **Pick a plan** — Developer ($20/mo, 2000 min) is enough for the spike; can downgrade/cancel after.

Once #1 + #2 are done, re-run this workflow (PR → Checks → "Runner Comparison" → Re-run, or `gh workflow run runner-comparison.yml`) and both columns of the comparison should populate.

## Test plan
- [ ] PR opens, `Runner Comparison` workflow triggers automatically
- [ ] `github-runner` job completes green and posts duration in the summary
- [ ] After Depot org setup: `depot-runner` job completes green
- [ ] `compare` job posts the side-by-side table to the job summary
- [ ] Datadog receives `ci.runner.*` metrics tagged `runner:github` and `runner:depot` (only if `DD_API_KEY` is set)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/actions/runner-benchmark/action.yml
+++ b/.github/actions/runner-benchmark/action.yml
@@ -1,0 +1,103 @@
+# Shared lint/test/build steps used by runner-comparison.yml so the
+# GitHub-hosted and Depot-hosted jobs stay byte-identical.
+#
+# Inputs cover everything the steps need; the only difference between
+# callers is the runs-on label and the cost rate (handled in the
+# calling workflow).
+
+name: 'Runner Benchmark Steps'
+description: 'Cold yarn install + lint/test/build, capturing install time'
+
+inputs:
+  node-version:
+    description: 'Node major version (18/20/22)'
+    required: true
+  access-token:
+    description: 'GitHub Packages token (passed as a secret from the caller)'
+    required: true
+
+outputs:
+  install_time_ms:
+    description: 'Wall-clock time for `yarn install` in milliseconds'
+    value: ${{ steps.install_time.outputs.duration_ms }}
+
+runs:
+  using: composite
+  steps:
+    - name: Setup Node
+      uses: actions/setup-node@v4
+      with:
+        node-version: ${{ inputs.node-version }}
+
+    - name: Set NPM Config
+      shell: bash
+      run: npm config set "//npm.pkg.github.com/:_authToken=${{ inputs.access-token }}"
+
+    - name: Install Yarn
+      shell: bash
+      env:
+        NODE_VERSION: ${{ inputs.node-version }}
+      run: |
+        if (( NODE_VERSION >= 18 )); then
+          NODE_OPTIONS=--openssl-legacy-provider npm install -g yarn
+        else
+          npm install -g yarn
+        fi
+
+    - name: Record Install Start
+      id: install_start
+      shell: bash
+      run: echo "time=$(date +%s%3N)" >> "$GITHUB_OUTPUT"
+
+    - name: Yarn Install
+      shell: bash
+      env:
+        NODE_VERSION: ${{ inputs.node-version }}
+      run: |
+        if (( NODE_VERSION >= 18 )); then
+          NODE_OPTIONS=--openssl-legacy-provider yarn install
+        else
+          yarn install
+        fi
+
+    - name: Record Install Time
+      id: install_time
+      shell: bash
+      env:
+        INSTALL_START: ${{ steps.install_start.outputs.time }}
+      run: |
+        end=$(date +%s%3N)
+        echo "duration_ms=$((end - INSTALL_START))" >> "$GITHUB_OUTPUT"
+
+    - name: Yarn Lint
+      shell: bash
+      env:
+        NODE_VERSION: ${{ inputs.node-version }}
+      run: |
+        if (( NODE_VERSION >= 18 )); then
+          NODE_OPTIONS=--openssl-legacy-provider yarn lint
+        else
+          yarn lint
+        fi
+
+    - name: Yarn Test
+      shell: bash
+      env:
+        NODE_VERSION: ${{ inputs.node-version }}
+      run: |
+        if (( NODE_VERSION >= 18 )); then
+          NODE_OPTIONS=--openssl-legacy-provider yarn test
+        else
+          yarn test
+        fi
+
+    - name: Yarn Build
+      shell: bash
+      env:
+        NODE_VERSION: ${{ inputs.node-version }}
+      run: |
+        if (( NODE_VERSION >= 18 )); then
+          NODE_OPTIONS=--openssl-legacy-provider yarn build
+        else
+          yarn build
+        fi
--- a/.github/workflows/runner-comparison.yml
+++ b/.github/workflows/runner-comparison.yml
@@ -1,14 +1,9 @@
-# Runner Comparison Workflow for sage-lib
-# Runs identical jobs on GitHub and Depot runners to benchmark performance
+# A/B benchmark: GitHub-hosted vs Depot-hosted runners for the
+# lint/test/build pipeline. Linear: PE-3158.
 #
-# Usage:
-#   1. Add this file to .github/workflows/runner-comparison.yml
-#   2. Ensure DD_API_KEY secret is set in repo settings
-#   3. Trigger manually via workflow_dispatch or on PR
-#
-# After Depot account setup:
-#   - Connect Kajabi org to Depot at https://depot.dev
-#   - The depot-ubuntu-24.04 runner will automatically work
+# The shared per-runner steps live in
+# .github/actions/runner-benchmark/action.yml — this file just orchestrates
+# the two callers and the comparison summary.
 
 name: Runner Comparison (GitHub vs Depot)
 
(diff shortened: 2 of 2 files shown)


prsimp (changes requested):
**Verdict: Request changes.** The workflow does not currently run at all — it fails GitHub's workflow-file validation, so neither the comparison nor any single job has ever executed on this PR. There's also a guaranteed-invalid JSON payload in the cost metric.

## How this was verified

- `actionlint` on the file at the PR head commit (`41a6b9903b`).
- Branch run history: the `runner-comparison.yml` run **failed in 0s with 0 jobs** (`"This run likely failed because of a workflow file issue"`), and there is **no "Runner Comparison" check** on the PR — only `Lint-Test-Build`, labeler, and Aikido. The Test-plan items ("workflow triggers automatically", "github-runner completes green") are therefore unmet; the workflow never compiled.

---

## 🔴 BLOCKER — `secrets` context is not allowed in a step `if:` (workflow won't compile)

`runner-comparison.yml:124`, `:241`, `:329`

```yaml
if: ${{ vars.DD_API_KEY != '' || secrets.DD_API_KEY != '' }}
```

`secrets` is not an available context inside an `if:` expression. actionlint reports it exactly:

```
context "secrets" is not allowed here. available contexts are
"env", "github", "inputs", "job", "matrix", "needs", "runner", "steps", "strategy", "vars".
```

This is a hard parse error → GitHub rejects the entire file → the 0-second/0-job failure above. Nothing in this workflow runs until this is fixed, including the green `github-runner` job the description promises.

The standard fix is to surface the secret through `env` (which *is* allowed in `if:`) at the **job** level — a *step-level* `env` is not reliably visible to that same step's own `if`, so it has to be job- or workflow-level:

```yaml
jobs:
  github-runner:
    runs-on: ubuntu-24.04
    env:
      DD_API_KEY: ${{ secrets.DD_API_KEY }}
    steps:
      ...
      - name: Send Metrics to Datadog
        if: ${{ env.DD_API_KEY != '' }}
        run: ...
```

Same fix in all three jobs. This also lets the now-redundant `vars.DD_API_KEY` half go away (the description documents `DD_API_KEY` as a **secret**, not a repo var, so `vars.DD_API_KEY` is always empty).

---

## 🟠 MAJOR — the `estimated_cost_usd` Datadog metric always sends invalid JSON

`runner-comparison.yml:148` and `:265`

```yaml
"value": '$(echo "scale=6; ${{ steps.timer.outputs.duration_ms }} / 60000 * 0.006" | bc)'
```

`bc` drops the leading zero for values < 1, and CI cost-per-run is always well under $1:

```
$ echo "scale=6; 300000 / 60000 * 0.006" | bc   # a 5-min run
.030000
```

`"value": .030000` is not valid JSON (numbers can't start with `.`), so the Datadog `/api/v2/series` call returns 400 and that metric never lands — for every run, on both runners. The `duration_ms` / `install_time_ms` metrics are integers and are fine; only the cost metric is affected.

Fix: force a leading zero, e.g.

```bash
cost=$(awk "BEGIN{printf \"%.6f\", $DURATION/60000*0.006}")
```

The cost in the `compare` job's summary table (`:312`–`:313`) renders as `$.0300` for the same reason — cosmetic there, but worth fixing alongside.

---

## 🟡 MINOR

- **Cost rates look off, which undermines the spike's purpose.** `:148`/`:302` use `$0.006/min` for GitHub-hosted Linux; GitHub's standard 2-core Linux rate is `$0.008/min`. If that's the real rate, the comparison understates GitHub's cost and biases the "cost savings" conclusion. Worth confirming both per-minute figures (GitHub `0.006`, Depot `0.004`) against current pricing and citing the source in a comment, since cost is the deliverable.

- **The two runner jobs are ~100 near-identical lines each** (`github-runner` / `depot-runner` differ only in `runs-on`, the summary label, and the `0.006`/`0.004` rate). `lint-test-build.yml` already uses a matrix; a `strategy.matrix.include` over `{runner, tag, cost_rate}` here would halve the file and prevent the two copies from drifting.

- **`compare` will stay pending indefinitely until Depot resolves.** `:276` `needs: [github-runner, depot-runner]`, and `depot-ubuntu-24.04` (`:159`) won't schedule until the org is connected (actionlint flags it as an unknown label, expected here). With no `timeout-minutes`, a queued `depot-runner` leaves the whole run hanging. A `timeout-minutes` on each job makes that failure mode cleaner.

- **Stale header comment** `:4`–`:8`: "1. Add this file to .github/workflows/…" / "2. Ensure DD_API_KEY secret is set" — step 1 is done, and the rest duplicates the PR body. Trimming it keeps the file from carrying setup instructions that read as TODOs.

- **shellcheck noise** (SC2086 unquoted `$GITHUB_OUTPUT`, SC2046 on the unquoted `$(date)`/`$(bc)` substitutions). Low priority, but the SC2046 hits are on exactly the fragile inline-substitution-into-JSON lines from the MAJOR above; quoting/refactoring resolves both.

## ✅ Looks good

- The per-step body faithfully mirrors `lint-test-build.yml` — same `NODE_OPTIONS=--openssl-legacy-provider` guard, same `ACCESS_TOKEN` npm config, same `checkout@v4`/`setup-node@v4`. Consistent with repo convention.
- `pull_request` scoped to `paths: [the workflow file]` is the right call to avoid burning CI on every future PR post-merge.
- Intentionally no yarn cache, so it measures cold installs equally on both runners — correct for an apples-to-apples benchmark.
- Uses `pull_request` (not `pull_request_target`), so no elevated-token exposure.

allantaylor8907:
Thanks @prsimp — pushed cfdd0cfd addressing the blocker, the major, and most of the minors.

**🔴 Fixed — `secrets` in `if:`.** `DD_API_KEY` now lives in job-level `env`; the gate is `if: env.DD_API_KEY != ''`. Dropped the `vars.DD_API_KEY` half since the docs describe it as a secret. `actionlint` is clean apart from the expected unknown-label warning on `depot-ubuntu-24.04`.

**🟠 Fixed — `bc` leading-zero bug.** Cost is now computed with `awk "BEGIN{printf \"%.6f\", ...}"` (always emits `0.030000`), and the entire Datadog payload is built with `jq -n` using `--argjson`/`--arg`, which sidesteps both the JSON-validity issue and the SC2046 word-splitting concerns in one move. The compare-job's summary table is fixed alongside.

**🟡 MINORs — addressed:**
- **Dedup.** Extracted the shared checkout-less lint/test/build into `.github/actions/runner-benchmark/action.yml`. Each runner job is now ~25 lines (start timer → checkout → composite → timer/cost → DD). I went composite-action over `strategy.matrix.include` because per-matrix-combo job outputs aren't first-class in GHA, and the artifact dance to fan results into `compare` would have re-introduced more lines than it saved. Open to switching to matrix+artifacts if you prefer.
- **`timeout-minutes`.** 30 on each runner job, 5 on `compare`. A queued Depot job now fails cleanly instead of hanging the run.
- **Stale header.** Trimmed; the file just references PE-3158 and points at the composite.
- **shellcheck noise.** All `$GITHUB_OUTPUT`/`$GITHUB_STEP_SUMMARY` quoted, multi-line writes grouped with `{ ... } >>`, and the fragile inline `bc`/`date` substitutions are gone via `jq`.

**🟡 MINOR — cost-rate question (not changed but verified):** the $0.006/min figure for GitHub Linux 2-core x64 is current per https://docs.github.com/en/billing/reference/actions-runner-pricing (verified 2026-06-04, dated comment + source link added inline at the env vars). Depot $0.004/min also verified against https://depot.dev/pricing. So the workflow numbers stay; the `~39% drop` line in the Linear ticket was forward-looking and confused things — the $0.006 already reflects the post-drop rate.

**Still gated on Depot org setup** (DNS-level — nothing in code can resolve it): the `depot-runner` job will sit `Queued → timeout` until the Depot GitHub app is installed on the `Kajabi` org. The `github-runner` + `compare` columns should now produce data on this PR regardless.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

prsimp (approved):
**Verdict: Approve.** Re-reviewed `cfdd0cfd` — the two issues that kept the workflow from running are fixed, verified locally.

**🔴 BLOCKER — `secrets` context in `if:` — resolved.** `actionlint` 1.7.9 now compiles the file clean (only the expected `depot-ubuntu-24.04` unknown-label warning remains). `DD_API_KEY` moved to job-level `env` in all three jobs and the gate is `if: env.DD_API_KEY != ''`; the `vars.DD_API_KEY` half is gone. The earlier 0-second / 0-job parse failure no longer occurs.

**🟠 MAJOR — `estimated_cost_usd` invalid JSON — resolved.** Verified the cost path emits `0.030000` (leading zero) and the `jq -n --argjson cost` payload parses as valid JSON with `value` as a real number. Building the entire payload with `jq -n` also clears the SC2046 word-splitting on the old inline substitutions.

**Cost rates — my earlier MINOR was incorrect, withdrawn.** I cited `$0.008/min` for GitHub 2-core Linux. The current rate is `$0.006/min` (GitHub actions-runner-pricing) and Depot ubuntu-24.04 2 vCPU is `$0.004/min` (depot.dev/pricing), both verified 2026-06-04. The figures and the inline source comments are correct as written.

Remaining MINORs — `timeout-minutes`, composite-action dedup of the lint/test/build steps, trimmed header comment, and shellcheck quoting — are all addressed.

